PR #2, opened 2016-11-03
CMake build and Documentation

bkeys:
This is a complete CMake build for IrrlichtBAW that includes all the previous features of the Code::Blocks project and more. This is good for the future of Irrlicht as it will finally have a modern build system; this will also help with efforts to make this cross platform. The added features of the CMake build are
- Library can now be compiled as static or shared
- Examples can be compiled optionally

I also took the liberty of running Doxygen over the include/ directory if we ever need documentation for the public interfaces.
We also wrote a short README to let people know what IrrlichtBAW is and mentions all the contributors.


devshgraphicsprogramming:
Please split this into some merge-able chunks, for example your very first commit fucks with my LICENSE file... and 946 files changed and 22000 lines removed while 6000+ added is a bit too much of a change to view especially when most of these changed lines simply have different line endings in some places

PR #9, opened 2017-10-31
Fix code format and add some minor changes

manhnt9:
Note for the commit "Move os.h from source/Irrlicht to include/":
I think include/ folder should have the code that user needs. When I add the include folder to my project, it shows that os.h is missing so I move the file there. I don't include os.h directly but some file in irrlicht.h needs it.


devshgraphicsprogramming:
This change is too big

PR #14, opened 2018-01-28
BAW mesh writer

Crisspl:


devshgraphicsprogramming:
Overall great, but why pack SMaterial?

I can understand SMaterialLayer as the interface is already finalized.

devshgraphicsprogramming:
Another question, why do we create separate types for:
irr::core::Blob::E_BLOB_TYPE::EBT_ATTR_BUFFER, 
irr::core::Blob::E_BLOB_TYPE::EBT_IDX_BUFFER
????

a data buffer is a data buffer, and it should remain so without any type distinctions

Crisspl:
(Also STextureSamplingParams got packed). To be able to just copy object data without worrying about alignment on various compilers (mesh exported by compilation on compiler A could be not importable on compiler B compilation). Am I missing something or it''s just worthless? 

Separate blob types for index and attribute buffer - I just thought it would be useful to know what kind of buffer is it, but don't really know why yet. As for now I don't have any good reason. Remove this (leave only EBT_ATTR_BUFFER)?

devshgraphicsprogramming:
Ok we can try with packing (but I thought that padding and packing is specified par the C++ standard), but I need to do a code review if it won't fuck something up in BaW :)

As for EBT_ATTR_BUFFER, I'd have it as EBT_RAW_DATA_BUFFER

Crisspl:
>but I thought that padding and packing is specified par the C++ standard
  
From what I know, padding is not standardized

devshgraphicsprogramming:
sorry I confused OpenGL GLSL packing (std430, etc.)

in that case we need some static_asserts to check the offset alignment of any SSE3 based members and force packing+16byte alignment on SMaterial

devshgraphicsprogramming:
On another note, maybe we should change  
+		enum E_BLOB_COMPRESSION_TYPE
+		{
+			EBCT_NONE = 0,
+			EBCT_LZ4,
+			EBCT_LZMA,
+			EBCT_COUNT
+		};
To
+		enum E_BLOB_CODING_TYPE
+		{
+			EBCT_RAW = 0,
+			EBCT_AES128_GCM,
+			EBCT_LZ4,
+			EBCT_LZ4_AES128_GCM,
+			EBCT_LZMA,
+			EBCT_LZMA_AES128_GCM,
+			EBCT_COUNT
+		};

For added possibility of password protection in the future.

Then obviously the bitfield "uint64_t compressionType : 2;" would have to be expanded to 3

devshgraphicsprogramming:
Another note... its not good to have private members holding data of the file to be written:
core::array<core::BlobHeader> m_headers;
core::array<uint32_t> m_offsets;

This will hamper multi-threading our asset loading, as having two or more concurrent writeMesh(...) calls will use the same private members.

Instead I'd make a CBAWMeshWriter::SSaveContext class (with no default ctors) to hold these.

P.S. Suggestion, make uint32_t genHeaders(ICPUMesh * _mesh); a template function, and add /*genAllHeaders(const std::vector<IReferenceCounted*> assetsToSave);*/ , as afterall we will use this to save more than one mesh at once and more assets than just meshes.

P.P.S. On top of that I would make the "std::set<const core::ICPUBuffer*> countedBuffers;" context wide!

devshgraphicsprogramming:
Final note...

You shouldn't single out "std::set<const core::ICPUBuffer*> countedBuffers;" , multiple IMeshBuffer s can use the same MeshDataFormatDescriptors and you could even add the same ICPUMeshBuffer to an ICPUMesh more than once!

So catch multiple references to the same object whenever you can!

So maybe change it to a "std::set<const IReferenceCounted*>"

devshgraphicsprogramming:
Final Note... for a IMeshBuffer to display its necessary for it to have a FormatDescriptor set.

So in genHeaders you can just do:
    if (!meshBuffer || !desc)
        continue;

Also one IMeshDataFormatDesc can be reused by multiple meshes (trust me they not only can but are in BaW) and if you're really mental you can attach a IMeshBuffer twice or more times to an IMesh.

Hence instead of an "std::set<const core::ICPUBuffer*>", I'd use an "std::set<const IReferenceCounted*>" and just keep track of EVERYTHING that way. 

Afterall a pointer is a pointer, and you will get no collisions!

Crisspl:
Ok, quick fix, thanks! About multithreading thing and others I will talk to you later

source/Irrlicht/CBAWMeshWriter.cpp
```diff
@@ -0,0 +1,269 @@
...
+			_file->write(m_headers.const_pointer(), m_headers.size() * sizeof(core::BlobHeader));
+
+			_file->seek(prevPos);
+
+			return true;
+		}
+
+		uint32_t CBAWMeshWriter::genHeaders(ICPUMesh * _mesh)
+		{
+			m_headers.clear();
```
devshgraphicsprogramming: what about a blob for the _mesh itself!?

source/Irrlicht/CBawFile.h
```diff
@@ -45,6 +60,7 @@ namespace irr { namespace core
 			EBT_COUNT
 		};
 
+		//! Meant to be used as pointer to blob's data
 		uint8_t data[1];
```
devshgraphicsprogramming: Maybe boot this and replace with a 'void* getData() {return this;}'

source/Irrlicht/CBAWMeshWriter.cpp
```diff
@@ -0,0 +1,269 @@
...
+#ifdef _DEBUG
+			setDebugName("CBAWMeshWriter");
+#endif
+		}
+
+		template<>
+		void CBAWMeshWriter::exportAsBlob<ICPUMeshBuffer>(ICPUMeshBuffer* _obj, uint32_t _headerIdx, io::IWriteFile* _file)
+		{
+			uint8_t* const data = (uint8_t*)std::malloc(m_headers[_headerIdx].blobSizeDecompr);
+			uint8_t* nearestFree = data;
```
devshgraphicsprogramming: All this stuff is nice, but not very maintanable.. maybe make a "struct MeshBufferBlobV1 : public Blob" which will construct itself from an ICPUMeshBuffer*

And also use the stack instead of heap! 
Making our exporter faster!

source/Irrlicht/CBAWMeshWriter.cpp
```diff
@@ -0,0 +1,269 @@
...
+			calcAndPushNextOffset(!_headerIdx ? 0 : m_headers[_headerIdx-1].blobSizeDecompr);
+
+			_file->write(path.c_str(), len);
+		}
+		template<>
+		void CBAWMeshWriter::exportAsBlob<IMeshDataFormatDesc<core::ICPUBuffer> >(IMeshDataFormatDesc<core::ICPUBuffer>* _obj, uint32_t _headerIdx, io::IWriteFile* _file)
+		{
+			uint8_t* const data = (uint8_t*)std::malloc(m_headers[_headerIdx].blobSizeDecompr);
+
+			E_COMPONENTS_PER_ATTRIBUTE* cpa = (E_COMPONENTS_PER_ATTRIBUTE*)data;
```
devshgraphicsprogramming: All this stuff is nice, but not very maintanable.. maybe make a "struct MeshDataFormatDescBlobV1 : public Blob" which will construct itself from an IMeshDataFormatDesc<ICPUBuffer>*

And also use the stack instead of heap! 
Making our exporter faster!

source/Irrlicht/CBAWMeshWriter.cpp
```diff
@@ -0,0 +1,269 @@
...
+			E_COMPONENT_TYPE* attType = (E_COMPONENT_TYPE*)((uint8_t*)cpa + EVAI_COUNT * sizeof(E_COMPONENTS_PER_ATTRIBUTE));
+			size_t* attStride = (size_t*)((uint8_t*)attType + EVAI_COUNT * sizeof(E_COMPONENT_TYPE));
+			size_t* attOffset = (size_t*)((uint8_t*)attStride + EVAI_COUNT * sizeof(size_t));
+			uint32_t* attDiv = (uint32_t*)((uint8_t*)attOffset + EVAI_COUNT * sizeof(size_t));
+			uint64_t* attBufs = (uint64_t*)((uint8_t*)attDiv + EVAI_COUNT * sizeof(uint32_t));
+
+			uint64_t* idxBuf = (uint64_t*)((uint8_t*)attBufs + EVAI_COUNT * sizeof(uint64_t));
+
+			for (int i = 0; i < EVAI_COUNT; ++i)
```
devshgraphicsprogramming: Loop unrolling will help and happen automagically by the compiler... but!

The memory is being walked very erratically, which the compiler won't fix, its much better for performance to have many loops for each array to be saved.

source/Irrlicht/CBAWMeshWriter.cpp
```diff
@@ -0,0 +1,269 @@
...
+				if (!meshBuffer || !desc)
+					continue;
+
+				if (countedObjects.find(meshBuffer) == countedObjects.end())
+				{
+					core::BlobHeader bh;
+					bh.handle = reinterpret_cast<uint64_t>(meshBuffer);
+					bh.compressionType = core::Blob::EBCT_RAW;
+					bh.blobType = core::Blob::EBT_MESH_BUFFER;
+					bh.blobSize = bh.blobSizeDecompr = // size of significant members of ICPUMeshBuffer
```
devshgraphicsprogramming: This is complicated to maintain, maybe we should set blobSizeDecompr later?

Or at least as sizeof(MeshBufferBlobV1) ?

source/Irrlicht/CBAWMeshWriter.cpp
```diff
@@ -0,0 +1,269 @@
...
+					}
+				}
+
+				if (countedObjects.find(desc) == countedObjects.end())
+				{
+					core::BlobHeader bh;
+					bh.handle = reinterpret_cast<uint64_t>(desc);
+					bh.compressionType = core::Blob::EBCT_RAW;
+					bh.blobType = core::Blob::EBT_DATA_FORMAT_DESC;
+					bh.blobSize = bh.blobSizeDecompr = // size of members of ICPUMeshDataFormatDesc
```
devshgraphicsprogramming: This is complicated to maintain, maybe we should set blobSizeDecompr later?

Or at least as sizeof(IMeshDataFormatDescBlobV1) ?

devshgraphicsprogramming:
I've reviewed a bit more of your code, hopefully we will get this is very soon...

source/Irrlicht/CBAWMeshWriter.cpp
Crisspl: Considered as not needed since it's just collection of ptrs to mesh buffers. Plus bounding box, but this can be recalculated after adding all loaded mesh buffers.

source/Irrlicht/CBawFile.h
Crisspl: Yea, better

source/Irrlicht/CBAWMeshWriter.cpp
devshgraphicsprogramming: Sure, but think what happens once we put more than one IMesh in a file.

P.S. The bounding box can be quite expensive to recalculate for a skinned mesh.

source/Irrlicht/CBAWMeshWriter.cpp
Crisspl: Nice idea,  thanks

source/Irrlicht/CBAWMeshWriter.cpp
```diff
@@ -28,40 +28,34 @@ namespace irr {
...
-			nearestFree += sizeof(size_t);
-			*((uint32_t *)nearestFree) = _obj->getBaseInstance();
-			nearestFree += sizeof(uint32_t);
-			*((E_PRIMITIVE_TYPE *)nearestFree) = _obj->getPrimitiveType();
-			nearestFree += sizeof(E_PRIMITIVE_TYPE);
-			*((E_VERTEX_ATTRIBUTE_ID *)nearestFree) = _obj->getPositionAttributeIx();
-			nearestFree += sizeof(E_VERTEX_ATTRIBUTE_ID);
-
+			const uint32_t cnt = _obj->getMeshBufferCount();
+			const uint32_t size = sizeof(uint32_t) + sizeof(core::aabbox3df) + cnt*sizeof(uint64_t);
```
devshgraphicsprogramming: Use placement new to create a MeshBlob over a larger data memory allocation

source/Irrlicht/CBAWMeshWriter.cpp
```diff
@@ -28,40 +28,34 @@ namespace irr {
...
 			_file->write(data, m_headers[_headerIdx].blobSizeDecompr);
-			calcAndPushNextOffset(!_headerIdx ? 0 : m_headers[_headerIdx-1].blobSizeDecompr);
+			calcAndPushNextOffset(!_headerIdx ? 0 : m_headers[_headerIdx - 1].blobSizeDecompr);
 
 			std::free(data);
 		}
 		template<>
+		void CBAWMeshWriter::exportAsBlob<ICPUMeshBuffer>(ICPUMeshBuffer* _obj, uint32_t _headerIdx, io::IWriteFile* _file)
+		{
+			const core::MeshBufferBlobV1 data(*_obj);
```
devshgraphicsprogramming: ADT violation

source/Irrlicht/CBawFile.cpp
```diff
@@ -0,0 +1,46 @@
...
+// This file is part of the "Irrlicht Engine" and "Build A World".
+// For conditions of distribution and use, see copyright notice in irrlicht.h
+// and on http://irrlicht.sourceforge.net/forum/viewtopic.php?f=2&t=49672
+
+#include "CBawFile.h"
+
+namespace irr { namespace core
+{
+
+	MeshBufferBlobV1::MeshBufferBlobV1(const scene::ICPUMeshBuffer& _mb)
```
devshgraphicsprogramming: change to pointer

PR #21, opened 2018-01-31
BAW mesh loader

Crisspl:


devshgraphicsprogramming:
please merge your branch, I needed to change some things for C++03 TR1 compliance

devshgraphicsprogramming:
I think this merge didnt take my std::memcpy for memcpy and set<>::emplace to set<>::insert fixes

For ref: https://github.com/buildaworldnet/IrrlichtBAW/commit/b3363be991506280768d45fa1fc41ceb764a26e5

Crisspl:
Well, it took, but for some reason doesn't show in diff..
And why actually C header (`"stdint.h"`) and why quatation marks instead of `<>` ?

devshgraphicsprogramming:
Becuase cstdint is from C++11, GCC throws a fit about it.

All that stdint provides is typedefs for (u)intN_t

include/CFinalBoneHierarchy.h
```diff
@@ -100,6 +101,35 @@ namespace scene
                 createAnimationKeys(inLevelFixedJoints);
             }
 
+			void fillExportBlob(void* _dataPtr, uint32_t* _size)
```
devshgraphicsprogramming: Maybe have a "Blob* SerializeToBlob(void** initialDataPtr, uint32_t& blobSize=0);" function for future use in all classes?

With a requirement that initialDataPtr is either NULL or an address of a pointer to memory which can be reallocated using realloc.

And then at the end of the function return *initialDataPtr;

Ofc blobSize would be the starting size of the initialDataPtr, and after it gets written the true size.

include/IMeshBuffer.h
```diff
@@ -21,7 +21,7 @@ namespace irr
 {
 namespace scene
 {
-
+	//! Available vertex attribute ids
```
devshgraphicsprogramming, line 24: could add that As of 2018 most OpenGL implementations support 16 attributes (some CAD GPUs more)

include/IMeshBuffer.h
```diff
@@ -43,6 +43,7 @@ namespace scene
         EVAI_COUNT
     };
 
+	//! Available components' sizes enumeration
     enum E_COMPONENTS_PER_ATTRIBUTE
     {
         ECPA_REVERSED_OR_BGRA=0,
```
devshgraphicsprogramming: special ID for reverse XYZW order

include/IMeshBuffer.h
```diff
@@ -43,6 +43,7 @@ namespace scene
         EVAI_COUNT
     };
 
+	//! Available components' sizes enumeration
```
devshgraphicsprogramming, line 47: Best thing about this enum is that it can be casted to an int to give you the actual component count :+1:

include/IMeshBuffer.h
```diff
@@ -134,11 +136,20 @@ GL_INVALID_OPERATION is generated by glVertexAttribPointer if size is GL_BGRA an
         {0xdeadbeefu,8,16,24,32}
     };
 
+	//! Checks whether given combination of attribute's type and component count is valid
+	/**@param type Type of an atrribute.
+	@param components Component count of an atrribute.
+	@returns true if such combination is valid or false otherwise.
+	*/
```
devshgraphicsprogramming: In OpenGL some formats are special like RGB30_A2 or RGB9_E5 which force the number of components and give no flexibility in the number of channels

include/IMeshBuffer.h
```diff
@@ -134,11 +136,20 @@ GL_INVALID_OPERATION is generated by glVertexAttribPointer if size is GL_BGRA an
...
     inline bool validCombination(const E_COMPONENT_TYPE& type, const E_COMPONENTS_PER_ATTRIBUTE& components)
     {
         return vertexAttrSize[type][components]<0xdeadbeefu;
     }
 
+	//! Checks whether given attribute type is a normalized one.
```
devshgraphicsprogramming, line 152: In OpenGL parlance Normalized-Integer means whether an integer value gets rescaled to [0,1] or [-1,1] float value respectively for unsigned and signed integer, by the time it reaches the shader input as a vertex attribute or texel from a texture.

include/CFinalBoneHierarchy.h
```diff
@@ -9,6 +9,7 @@
 #include "IGPUMappedBuffer.h"
 #include "quaternion.h"
 #include "irrString.h"
+#include "CBawFile.h"
```
devshgraphicsprogramming, line 12: If we're including it here, then move CBawFile.h from "source/Irrlicht" to "include"

include/IMeshBuffer.h
```diff
@@ -157,6 +168,7 @@ GL_INVALID_OPERATION is generated by glVertexAttribPointer if size is GL_BGRA an
             size_t attrOffset[EVAI_COUNT];
             uint32_t attrDivisor[EVAI_COUNT];
 	public:
+		//! Default constructor.
```
devshgraphicsprogramming, line 174: we should move this into the protected section, as this class is an ADT

Definitely dont want to be able to create these objects out of thin-air

include/IMeshBuffer.h
```diff
@@ -573,8 +585,9 @@ GL_INVALID_OPERATION is generated by glVertexAttribPointer if size is GL_BGRA an
 	public:
 	    ICPUMeshBuffer(core::LeakDebugger* dbgr=NULL) : IMeshBuffer<core::ICPUBuffer>(NULL,dbgr), posAttrId(EVAI_ATTR0) {}
 
-
+		//! Returns id of position attribute.
```
devshgraphicsprogramming, line 591: As mentioned in out previous conversations:

We only keep track of a position attribute, as every vertex needs to have at least a position to be displayed on the screen. Certain vertices may not have colors, normals, texture coords, etc. but a position is always present.

I wanted to give flexibility and not just assume attribute 0 to be the old gl_Vertex blindly.

On another note we need to know at least what is the vertex attribute for position to calculate bounding boxes reliably.

However if you want to do some weird GPGPU based off drawing vertices, you can safely ignore this if you don't use bounding box culling.

include/IMeshBuffer.h
```diff
@@ -602,6 +615,8 @@ GL_INVALID_OPERATION is generated by glVertexAttribPointer if size is GL_BGRA an
 		    return ((uint8_t*)meshLayout->getIndexBuffer()->getPointer())+indexBufOffset;
         }
 
+		//! Get access to Indices.
+		/** \return Pointer to indices array. */
```
devshgraphicsprogramming: for consistency call it the index array

include/CFinalBoneHierarchy.h
```diff
@@ -100,6 +101,35 @@ namespace scene
                 createAnimationKeys(inLevelFixedJoints);
             }
 
+			void fillExportBlob(void* _dataPtr, uint32_t* _size)
+			{
+				new (_dataPtr) core::FinalBoneHierarchyBlobV1(boneCount, NumLevelsInHierarchy, keyframeCount);
+				uint8_t* ptr = (uint8_t*)_dataPtr + sizeof(core::FinalBoneHierarchyBlobV1);
```
devshgraphicsprogramming: On another note, maybe instead of doing the copying here... do it in the constructor of the FinalBoneHierarchyBlobV1 class? or a static createOnMemory(void* dataPtr) function?
At least for consistency :)

We can get all the private members of CFinalBoneHierarchy, i.e.
keyframeCount == CFinalBoneHierarchy::getKeyframeCount()
keyframes == CFinalBoneHierarchy::getKeys()
interpolatedAnimations == CFinalBoneHierarchy::getInterpolatedAnimationData(0)
nonInterpolatedAnimations == CFinalBoneHierarchy::getNonInterpolatedAnimationData(0)
etc.

suddani:
Wunderbar

Am 01.02.2018 12:46 schrieb "devshgraphicsprogramming" <
notifications@github.com>:

include/IMeshBuffer.h
```diff
@@ -612,6 +627,11 @@ GL_INVALID_OPERATION is generated by glVertexAttribPointer if size is GL_BGRA an
 		    return ((const uint8_t*)meshLayout->getIndexBuffer()->getPointer())+indexBufOffset;
         }
 
+		//! Accesses given index of mapped position attribute buffer.
+		/** @param ix Index number which is to be returned.
```
devshgraphicsprogramming: Vertex with index ix which is to be returned

include/IMeshBuffer.h
```diff
@@ -612,6 +627,11 @@ GL_INVALID_OPERATION is generated by glVertexAttribPointer if size is GL_BGRA an
...
+		@returns `ix`th index of mapped attribute buffer or (0, 0, 0, 1) vector if an error occured (e.g. no such index).
```
devshgraphicsprogramming: the 'ix'th vertex

no such vertex

Not index

include/IMeshBuffer.h
```diff
@@ -625,6 +645,11 @@ GL_INVALID_OPERATION is generated by glVertexAttribPointer if size is GL_BGRA an
 		    return outPos;
         }
 
+		//! Accesses data of buffer of attribute of given id
+		/** @param attrId Attribute id.
+		@returns Pointer to corresponding buffer's data incremented by `baseVertex`
```
devshgraphicsprogramming: and by 'bufferOffset' also

Basically it will get the start of the array at the same point as OpenGL will get upon a glDraw*

When accessing the array take account of the attribute stride as consecutive elements in the array will not be attribute #attrId of consecutive vertices

include/IMeshBuffer.h
```diff
@@ -649,6 +674,14 @@ GL_INVALID_OPERATION is generated by glVertexAttribPointer if size is GL_BGRA an
             return ((uint8_t*)mappedAttrBuf->getPointer())+ix;
         }
 
+		//! Accesses given index of given attribute. Index number is incremented by `baseVertex`.
```
devshgraphicsprogramming: vertex attribute not index attribute

source/Irrlicht/CBawFile.h
```diff
@@ -66,11 +78,26 @@ namespace irr { namespace core
 	};
 
 #include "irrpack.h"
-	//! Utility struct. Used only while loading (CBAWMeshLoader). Cast blob pointer to MeshBlob* to make life easier.
-	struct MeshBlob : Blob
+	//! Utility struct. Cast blob pointer to MeshBlob* to make life easier.
+	struct MeshBlobV1 : Blob
```
devshgraphicsprogramming: see my note about variable size blobs

source/Irrlicht/CBawFile.h
```diff
@@ -66,11 +78,26 @@ namespace irr { namespace core
...
+		//! WARNING: Constructor saves only bounding box and mesh buffer count (not mesh buffer pointers)
+		explicit MeshBlobV1(const aabbox3df & _box, uint32_t _cnt);
+
+		aabbox3df box;
 		uint32_t meshBufCnt;
+		uint64_t meshBufPtrs[1];
+	} PACK_STRUCT;
+
+	//! Utility struct. Cast blob pointer to MeshBlob* to make life easier.
+	struct SkinnedMeshBlobV1 : Blob
```
devshgraphicsprogramming: see my note about variable size blobs

source/Irrlicht/CBawFile.h
```diff
@@ -107,6 +144,15 @@ namespace irr { namespace core
 		uint64_t attrBufPtrs[scene::EVAI_COUNT];
 		uint64_t idxBufPtr;
 	} PACK_STRUCT;
+
+	struct FinalBoneHierarchyBlobV1 : Blob
```
devshgraphicsprogramming: see my note about variable size blobs

source/Irrlicht/CBawFile.h
```diff
@@ -107,6 +144,15 @@ namespace irr { namespace core
...
+	{
+		FinalBoneHierarchyBlobV1(size_t _bCnt, size_t _numLvls, size_t _kfCnt);
```
devshgraphicsprogramming: Maybe have a constructor/static create function directly from the object being saved

+see my note about variable size blobs

source/Irrlicht/CBawFile.h
```diff
@@ -66,11 +78,26 @@ namespace irr { namespace core
 	};
 
 #include "irrpack.h"
-	//! Utility struct. Used only while loading (CBAWMeshLoader). Cast blob pointer to MeshBlob* to make life easier.
-	struct MeshBlob : Blob
+	//! Utility struct. Cast blob pointer to MeshBlob* to make life easier.
+	struct MeshBlobV1 : Blob
 	{
+		//! WARNING: Constructor saves only bounding box and mesh buffer count (not mesh buffer pointers)
+		explicit MeshBlobV1(const aabbox3df & _box, uint32_t _cnt);
```
devshgraphicsprogramming: Maybe have a constructor/static create function directly from the object being saved

+see my note about variable size blobs

devshgraphicsprogramming:
About  variable size blobs such as MeshBlobV1, FinalBoneHierarchyBlobV1, etc.

Lets inherit them from a template struct
`template<class T>
struct VariableSizeBlob : public Blob
{
    static size_t caclBlobSizeForObj(T* _obj) {return 0;}
    static void*  allocMemForBlob(T* _obj) {return malloc(caclBlobSizeForObj<T>());}
};`
as i.e. class MeshBlobV1 : public VariableSizeBlob<ICPUMesh>

Then specialize the calcBlobSizeForObj function for all objects?

And finally change the structs inheriting from VariableSizeBlob to classes and make the constructors private and only provide a static placement new wrapper ..

source/Irrlicht/CBawFile.h
```diff
@@ -66,11 +78,26 @@ namespace irr { namespace core
...
+		aabbox3df box;
 		uint32_t meshBufCnt;
+		uint64_t meshBufPtrs[1];
+	} PACK_STRUCT;
+
+	//! Utility struct. Cast blob pointer to MeshBlob* to make life easier.
+	struct SkinnedMeshBlobV1 : Blob
+	{
+		//! WARNING: Constructor saves only bone hierarchy, bounding box and mesh buffer count (not mesh buffer pointers)
+		explicit SkinnedMeshBlobV1(scene::CFinalBoneHierarchy* _fbh, const aabbox3df & _box, uint32_t _cnt);
```
devshgraphicsprogramming: Maybe have a constructor/static create function directly from the object being saved

+see my note about variable size blobs

source/Irrlicht/CBAWMeshWriter.cpp
```diff
@@ -159,34 +211,51 @@ namespace irr {
 		{
 			_ctx.headers.clear();
 
+			bool isMeshAnimated = true;
+			ICPUSkinnedMesh* skinnedMesh = 0;
+
 			if (_mesh)
 			{
-				core::BlobHeader bh;
+				skinnedMesh = dynamic_cast<ICPUSkinnedMesh*>(_mesh);
```
devshgraphicsprogramming: can be a static_cast or reinterpret_cast, just check the mesh type before casting

source/Irrlicht/CBAWMeshWriter.cpp
```diff
@@ -75,6 +102,18 @@ namespace irr {
 			_file->write(path.c_str(), len);
 		}
 		template<>
+		void CBAWMeshWriter::exportAsBlob<scene::CFinalBoneHierarchy>(scene::CFinalBoneHierarchy* _obj, uint32_t _headerIdx, io::IWriteFile* _file, SContext& _ctx)
+		{
+			uint8_t data[1u<<14]; // 16kB
```
devshgraphicsprogramming: You will need to use the heap for this bad boy

devshgraphicsprogramming:
Also I'm seeing a lot of 
`_ctx.headers[_headerIdx].blobSize = _ctx.headers[_headerIdx].blobSizeDecompr = size;
core::XXHash_256(&data, _ctx.headers[_headerIdx].blobSizeDecompr, _ctx.headers[_headerIdx].blobHash);
_file->write(data, _ctx.headers[_headerIdx].blobSizeDecompr);`
In every exportAsBlob

maybe we should make a general function for this on the Blob or in the CBAWMeshWriter, so we don't repeat ourselves?

include/IMeshBuffer.h
```diff
@@ -649,6 +674,14 @@ GL_INVALID_OPERATION is generated by glVertexAttribPointer if size is GL_BGRA an
...
+		/** If component count of given attribute is less than 4, only first ones of output vector's members will be written.
```
devshgraphicsprogramming: Big note for this function, NOT ALL FORMAT CONVERSIONS TO RGBA32F/XYZW32F ARE IMPLEMENTED!

include/IMeshBuffer.h
```diff
@@ -649,6 +674,14 @@ GL_INVALID_OPERATION is generated by glVertexAttribPointer if size is GL_BGRA an
...
+		@param[out] output Vector object to which index's value will be returned.
```
devshgraphicsprogramming: vectorSIMDf object

include/IMeshBuffer.h
```diff
@@ -649,6 +674,14 @@ GL_INVALID_OPERATION is generated by glVertexAttribPointer if size is GL_BGRA an
...
+		@param[in] attrId Atrribute id.
+		@param[in] ix Index which is to be accessed. Will be incremented by `baseVertex`.
```
devshgraphicsprogramming: vertex

include/IMeshBuffer.h
```diff
@@ -649,6 +674,14 @@ GL_INVALID_OPERATION is generated by glVertexAttribPointer if size is GL_BGRA an
...
+		@returns true if successful or false if an error occured (e.g. no such index or given atrribute's type not implemented).
```
devshgraphicsprogramming: ix out of range, no attribute specified/bound or given attribute's format conversion to vectorSIMDf unsupported

include/IMeshBuffer.h
```diff
@@ -863,6 +896,13 @@ GL_INVALID_OPERATION is generated by glVertexAttribPointer if size is GL_BGRA an
             return true;
         }
 
+		//! Sets value of given index of given attribute.
+		/** @param input Value which is to be set.
+		@param attrId Atrribute id.
+		@param ix Index which is to be set. Will be incremented by `baseVertex`.
```
devshgraphicsprogramming: vertex not index, the index is "of the vertex"

include/IMeshBuffer.h
```diff
@@ -863,6 +896,13 @@ GL_INVALID_OPERATION is generated by glVertexAttribPointer if size is GL_BGRA an
             return true;
         }
 
+		//! Sets value of given index of given attribute.
```
devshgraphicsprogramming: Big note for this function, NOT ALL FORMAT CONVERSIONS FROM RGBA32F/XYZW32F (vectorSIMDf) ARE IMPLEMENTED!

include/IMeshBuffer.h
```diff
@@ -157,6 +168,7 @@ GL_INVALID_OPERATION is generated by glVertexAttribPointer if size is GL_BGRA an
             size_t attrOffset[EVAI_COUNT];
             uint32_t attrDivisor[EVAI_COUNT];
 	public:
```
devshgraphicsprogramming, line 173: Should add a comment with a link to https://www.khronos.org/opengl/wiki/Vertex_Specification
telling people to read that to understand the key terms like attribute IDs, indices, attribute formats etc.

source/Irrlicht/CBawFile.h
Crisspl: About variable size blob: I'm allocating memory on stack for it (see for example CBAWMeshWriter.cpp:54), it's your idea actually. So now `VariableSizeBlob` would allocate mem on heap for us... At most it could make sense for `CFinalBoneHierarchy` and corresponding blob struct right now

devshgraphicsprogramming: Well if we know the blob to be less than 16kb then lets placement new it on the stack, but for things which can be much bigger like the CFinalBoneHierarchy (as it contains all the animations), we definitely need the heap.

So that VariableSizeBlob::caclBlobSizeForObj static function could help us differentiate between the two cases.

We dont have to call "allocMemForBlob"

It was more of an idea to have a static create function which requires you to say how much memory you reserved or allocated and assert if the constructor would overrun it, and that needs some sort of mechanism to allow us to create the object via a placement new but having the constructor declared as private.

source/Irrlicht/CBawFile.h
Crisspl: >Maybe have a constructor/static create function directly from the object being saved
  
We already talked about this and decided to load blob buffer with variable-size data outside ctor since it has to rely on the fact that the memory actually is allocated (and check it from inside ctor) and it's kinda dangerous

source/Irrlicht/CBawFile.h
Crisspl: >Maybe have a constructor/static create function directly from the object being saved
  
We already talked about this and decided to load blob buffer with variable-size data outside ctor since it has to rely on the fact that the memory actually is allocated (and check it from inside ctor) and it's kinda dangerous

source/Irrlicht/CBawFile.h
Crisspl: >Maybe have a constructor/static create function directly from the object being saved
  
We already talked about this and decided to load blob buffer with variable-size data outside ctor since it has to rely on the fact that the memory actually is allocated (and check it from inside ctor) and it's kinda dangerous

source/Irrlicht/CBawFile.h
devshgraphicsprogramming: True, thats why I'm pondering making the constructor private and instead making a static "createOverMemory" function which will call the constructor through placement new in private and check the memory size provided...

It would have to be a static function and a private constructor as constructors are not allowed to not create the object or to return anything else than the object.

include/CBawFile.cpp
```diff
@@ -0,0 +1,100 @@
+// Copyright (C) 2018 Krzysztof "Criss" Szenk
```
devshgraphicsprogramming: Sorry I didn't explain this before, but we never put .cpp files outside source/Irrlicht

include/CBawFile.cpp
```diff
@@ -0,0 +1,100 @@
...
+
+
+namespace irr { namespace core
+{
+
+		size_t VariableSizeBlob<scene::ICPUMesh>::calcBlobSizeForObj(scene::ICPUMesh* _obj)
+		{
+			return sizeof(MeshBlobV1) + (_obj->getMeshBufferCount()-1) * sizeof(uint64_t);
+		}
+		void* VariableSizeBlob<scene::ICPUMesh>::allocMemForBlob(scene::ICPUMesh* _obj) { return malloc(calcBlobSizeForObj(_obj)); }
```
devshgraphicsprogramming: does it not compile if you leave it unspecialized?

include/CBawFile.cpp
```diff
@@ -0,0 +1,100 @@
...
+
+		size_t VariableSizeBlob<scene::ICPUSkinnedMesh>::calcBlobSizeForObj(scene::ICPUSkinnedMesh* _obj)
+		{
+			return sizeof(SkinnedMeshBlobV1) + (_obj->getMeshBufferCount() - 1) * sizeof(uint64_t);
+		}
+		void* VariableSizeBlob<scene::ICPUSkinnedMesh>::allocMemForBlob(scene::ICPUSkinnedMesh* _obj) { return malloc(calcBlobSizeForObj(_obj)); }
```
devshgraphicsprogramming: does it not compile if you leave it unspecialized?

include/CBawFile.cpp
```diff
@@ -0,0 +1,100 @@
...
+			const uint32_t levelsCnt = _obj->getHierarchyLevels();
+			const uint32_t kfCnt = _obj->getKeyFrameCount();
+			return
+				sizeof(FinalBoneHierarchyBlobV1) +
+				sizeof(*_obj->getBoneData())*boneCnt +
+				_obj->getSizeOfAllBoneNames() +
+				kfCnt*(sizeof(*_obj->getInterpolatedAnimationData()) + sizeof(*_obj->getNonInterpolatedAnimationData()) + sizeof(*_obj->getKeys())) +
+				levelsCnt * sizeof(*_obj->getBoneTreeLevelEnd());
+		}
+		void* VariableSizeBlob<scene::CFinalBoneHierarchy>::allocMemForBlob(scene::CFinalBoneHierarchy* _obj) { return malloc(calcBlobSizeForObj(_obj)); }
```
devshgraphicsprogramming: does it not compile if you leave it unspecialized?

include/CBawFile.cpp
Crisspl: O, yea, it shoukld be obvious for me, my bad

Crisspl: i'll put it back

include/CBawFile.h
```diff
@@ -77,9 +78,31 @@ namespace core
 		void* getData() { return this; }
 	};
 
+	template<typename T>
+	struct VariableSizeBlob : Blob {};
```
devshgraphicsprogramming: "curiously recurring template" will allow us to declare static calcBlobSizeForObj and allocMemForBlob functions which can be "overriden" like virtual functions

include/CBawFile.cpp
Crisspl: I don't understand how I could leave it unspecialized

source/Irrlicht/CBawFile.h
devshgraphicsprogramming: True, thats why I'm pondering making the constructor private and instead making a static "createOverMemory" function which will call the constructor through placement new in private and check the memory size provided...

It would have to be a static function and a private constructor as constructors are not allowed to not create the object or to return anything else than the object.

include/IMeshBuffer.h
devshgraphicsprogramming: still forgot to change index's value to vertex's value

source/Irrlicht/CBAWMeshWriter.cpp
```diff
@@ -317,4 +327,10 @@ namespace irr {
 			_ctx.offsets.push_back(!_ctx.offsets.size() ? 0 : _ctx.offsets.getLast() + _blobSize);
 		}
 
+		void CBAWMeshWriter::finalizeHeader(core::BlobHeaderV1 & _header, const void* _data, uint32_t _size)
```
devshgraphicsprogramming: maybe make it a function of the BlobHeaderV* ?

include/CBawFile.h
Crisspl: yea maye.. But what's the point of doing so?

include/CBawFile.cpp
devshgraphicsprogramming: "curiously recurring template" will allow us to declare static calcBlobSizeForObj and allocMemForBlob functions which can be "overriden" like virtual functions

source/Irrlicht/CBAWMeshWriter.cpp
Crisspl: nice :+1:

include/CBawFile.h
devshgraphicsprogramming: not having to remember to specialize the allocMemForBlob function every time :)

include/CBawFile.h
```diff
@@ -7,27 +7,40 @@
...
-	struct BlobHeader
+	struct BlobHeaderV1
 	{
 		uint32_t blobSize;
 		uint32_t blobSizeDecompr;
 
-		uint64_t compressionType : 3;
-		uint64_t blobType : 7;
+		uint8_t compressionType;
+		uint8_t blobType;
```
devshgraphicsprogramming: I'm going to be a dick and change this again, could we have:
`uint8_t compressionType;
uint8_t dummy[3];
uint32_t blobType; //go to the wiki page for BlobType Registry`

https://github.com/buildaworldnet/IrrlichtBAW/wiki/BlobType-Registry

Order matters, I want to keep everything 64bit and type-aligned

include/CBawFile.h
Crisspl: The function is not specialized. It's not even a template, only struct is a template. See how it's used [in mesh writer](https://github.com/buildaworldnet/IrrlichtBAW/pull/21/files#diff-69beb7cce857086a5dc3a282bae8004aR65) (CBAWMeshWriter.cpp:65)

devshgraphicsprogramming: Sorry I meant declared and defined for every struct.

Crisspl:
I'm going to have much less time for a few days now. Could you just run writer/loader on some more complicated meshes to see if it's working? I'll also test it myself of course :D but some feedback from you would be nice since you know the exact needs

source/Irrlicht/CBAWMeshFileLoader.cpp
```diff
@@ -0,0 +1,206 @@
...
+ICPUSkinnedMesh* CBAWMeshFileLoader::make<ICPUSkinnedMesh>(SPair& _data, SContext& _ctx) const
+{
+	typename core::CorrespondingBlobTypeFor<ICPUSkinnedMesh>::type* blob = core::toBlobPtr<ICPUSkinnedMesh>(_data.blob);
+
+	CCPUSkinnedMesh* mesh = new scene::CCPUSkinnedMesh();
+	mesh->setBoneReferenceHierarchy(make<CFinalBoneHierarchy>(_ctx.blobs[blob->boneHierarchyPtr], _ctx));
+	mesh->setBoundingBox(blob->box);
+	for (uint32_t i = 0; i < blob->meshBufCnt; ++i)
+		mesh->addMeshBuffer(make<SCPUSkinMeshBuffer>(_ctx.blobs[blob->meshBufPtrs[i]], _ctx));
+	// shall i call mesh->finalize()?
```
Crisspl: What about this?

source/Irrlicht/CBAWMeshFileLoader.cpp
```diff
@@ -0,0 +1,206 @@
...
+		texture = m_sceneMgr->getVideoDriver()->getTexture(texname);
+	}
+	else
+	{
+		const io::path path = _ctx.filePath + texname;
+		newTexture = m_sceneMgr->getVideoDriver()->findTexture(path) == 0;
+		// try to read from the path relative to where the .baw is loaded from
+		texture = m_sceneMgr->getVideoDriver()->getTexture(path);
+	}
+	//! @todo @bug Do somemthing with `newTexture`? In obj loader something happens in case where newTexture is true
```
Crisspl: What about this? Don't really know why anything would be needed, but .obj loader does something...

source/Irrlicht/CBAWMeshWriter.cpp
Crisspl: Nope, `dynamic_cast` is the only one really safe here. `reinterpret_cast` just gave me crash in analogic situation (base class can have a bit different offset and only `dynamic_cast` cares about that).

devshgraphicsprogramming: ICPUSkinnedMesh is a direct inheritor of ICPUMesh, so reinterpret_cast should be safe.
There's no multiple inheritance etc. like for IVirtualTextures

if you were casting from IReferenceCounted, then you'd have a problem.

source/Irrlicht/CBAWMeshFileLoader.cpp
devshgraphicsprogramming: I'll test next week, and see

devshgraphicsprogramming:
Still don't like the fact that allocMemForBlob is not a member of VariableSizeBlob

We can make it work with a CRTP.
[https://en.wikipedia.org/wiki/Curiously_recurring_template_pattern#Static_polymorphism](url)

This ties is with the other thing I don't like which is that ALL blobs are not constructed and filled in their constructors, VariableSizeBlobs would just use static some static func, i.e. "createFromObj" which would use placement new internally, with private constructors.

source/Irrlicht/CBAWMeshFileLoader.cpp
devshgraphicsprogramming: I'll test next week, and see

include/CBawFile.h
```diff
@@ -173,14 +173,57 @@ namespace core
 
 	struct FinalBoneHierarchyBlobV1 : VariableSizeBlob<scene::CFinalBoneHierarchy>
 	{
+		friend class scene::CFinalBoneHierarchy;
```
devshgraphicsprogramming: I don't like friends.
Because I don't have any...

include/CFinalBoneHierarchy.h
```diff
@@ -101,6 +101,42 @@ namespace scene
                 createAnimationKeys(inLevelFixedJoints);
             }
 
+			//! Constructor from blob.
+			/**It's the only class in which such constructor exists since a lot of private members are copied with just `memcpy` call. Blob's data should be filled by fillExportBlob().
+			Meant to be used by loaders only.
+			@see @ref CBAWMeshFileLoader CBAWMeshWriter
+			*/
+			explicit CFinalBoneHierarchy(typename core::CorrespondingBlobTypeFor<CFinalBoneHierarchy>::type* _blob) :
```
devshgraphicsprogramming: I don't like this.

The blobs should create a CFinalBoneHierarchy, not the CFBH from Blob.

The reason is that imagine having 4 versions of the blob and also updating the CFBH class.

If you want the loader to be backward compatible with all Blob versions 1 through 4, then you will have 4 constructors polluting this file which should only be about getting interpolated keyframes for an armature.

Instead have a constructor:
`explicit CFinalBoneHierarchy(BoneReferenceData* boneFlatArrayBegin, BoneReferenceData* boneFlatArrayEnd, ..., etc.)`

Then when we move to C++11 we can use move semantics to move the bone name strings.

include/CFinalBoneHierarchy.h
devshgraphicsprogramming: any comment?

source/Irrlicht/CBAWMeshFileLoader.cpp
```diff
@@ -210,19 +210,20 @@ ICPUMesh* CBAWMeshFileLoader::createMesh(io::IReadFile* _file)
 	const uint32_t filesize = _file->getSize();
 	uint8_t* const fileBuffer = (uint8_t*)malloc(filesize);
 	_file->read(fileBuffer, filesize);
```
devshgraphicsprogramming: No need to read the entire file at once.

You dont check the file header, and that the file has a minimal length to support at least 1 blob it needs to be done in order not to load some bogus file and crash.

First only read the file header and check that it even matches, and determine the file version.... then dispatch to the correct loadHeadersVN function.

source/Irrlicht/CBAWMeshFileLoader.cpp
```diff
@@ -210,19 +210,20 @@ ICPUMesh* CBAWMeshFileLoader::createMesh(io::IReadFile* _file)
...
-	core::BawFileV1* const bawFile = (core::BawFileV1*)fileBuffer;
-	const uint32_t HEADERS_FILE_OFFSET = sizeof(core::BawFileV1::fileHeader) + sizeof(core::BawFileV1::numOfInternalBlobs) + sizeof(core::BawFileV1::blobOffsets[0])*bawFile->numOfInternalBlobs;
+	const core::BawFileV1* const bawFile = (core::BawFileV1*)fileBuffer;
+	const uint32_t HEADERS_FILE_OFFSET = bawFile->calcHeadersOffset();
+	const uint32_t BLOBS_FILE_OFFSET = bawFile->calcBlobsOffset();
```
devshgraphicsprogramming: again this needs to be checked if it not past the end of the file.

include/CFinalBoneHierarchy.h
Crisspl: Uh, sorry, must have missed this one. I'll talk to you about this and other issues in few days. As for now I don;t really see the need since `exportAsBlob` functions in mesh writer would do exactly same thing.

source/Irrlicht/CBAWMeshFileLoader.h
```diff
@@ -20,16 +20,18 @@ namespace irr { namespace scene
...
+	struct SBlobData
 	{
-		core::BlobHeaderV1* header;
-		void* blob;
+		const core::BlobHeaderV1* header;
+		const void* blob;
+		bool wasCreated;
 	};
 
 	struct SContext
```
devshgraphicsprogramming: Imagine the BAW file has a huge mesh inside it with auxillary data, or is an archive, it could be couple gigabytes!

We only want to read all the BlobHeaders into RAM, at once and linearly.
Then we can have an std::map<uint64_t,std::pair<BlobHeaderV1*,uint64_t> > which means std::map<handle,std::pair<BlobHeaderV1*,absPosInFile> >

The std::pair could be the SBlobData struct for all I care

source/Irrlicht/CBAWMeshFileLoader.h
```diff
@@ -20,16 +20,18 @@ namespace irr { namespace scene
...
 {
 private:
-	struct SPair
+	struct SBlobData
 	{
-		core::BlobHeaderV1* header;
-		void* blob;
+		const core::BlobHeaderV1* header;
+		const void* blob;
+		bool wasCreated;
```
devshgraphicsprogramming: If you already have a `createdObjects` member in the context, then there is no need for a wasCreated flag.
You can just use `std::map::find` to try-get the created object.

Or if you are scared of using two logN operations, then instead of `bool wasCreated` have a `void* createdObj` and ditch the `createdObjects' member - which now after I wrote this, seems a better idea.

Crisspl: Thanks to the flag I don't need to look for an object in map to determine if it was loaded or not

source/Irrlicht/CBAWMeshFileLoader.h
```diff
@@ -20,16 +20,18 @@ namespace irr { namespace scene
...
 	};
 
 	struct SContext
 	{
 		io::path filePath;
-		std::map<uint64_t, SPair> blobs;
+		std::map<uint64_t, SBlobData> blobs;
+		std::map<uint64_t, IReferenceCounted*> createdObjects;
```
devshgraphicsprogramming: Look at my SBlobData comment, you can ditch this in favor of an `IReferenceCounted* createdObj` in SBlobData.

source/Irrlicht/CBAWMeshFileLoader.h
```diff
@@ -52,7 +54,9 @@ class CBAWMeshFileLoader : public IMeshLoader
 
 private:
 	template<typename T>
-	T* make(SPair&, SContext&) const;
+	T* make(SBlobData&, SContext&) const;
```
devshgraphicsprogramming: maybe call it `tryMake` or `get`

source/Irrlicht/CBAWMeshFileLoader.h
```diff
@@ -52,7 +54,9 @@ class CBAWMeshFileLoader : public IMeshLoader
...
+
+	void registerObject(SBlobData& _data, SContext& _ctx, IReferenceCounted* _obj) const;
```
devshgraphicsprogramming: If you ditch `std::map<uint64_t, IReferenceCounted*> createdObjects` then you can most probably ditch this function too (or not, but it will just be almost a macro for a 1 line op).

include/CFinalBoneHierarchy.h
devshgraphicsprogramming: Blobs are really nice for serialization to a binary struct, which could be used with other data sinks such as other file formats, Networks, MPI, etc.

source/Irrlicht/CBAWMeshFileLoader.h
devshgraphicsprogramming: yes, and if the createdObj pointer is not NULL then it was definitely loaded.

You could also set it to 0xdeadbeefDEADBEEFull for things that failed to load.

devshgraphicsprogramming: The whole point of my suggestions is that there is no 2nd map

source/Irrlicht/CBAWMeshFileLoader.cpp
```diff
@@ -210,19 +210,20 @@ ICPUMesh* CBAWMeshFileLoader::createMesh(io::IReadFile* _file)
...
 
 	SContext ctx;
 	ctx.filePath = _file->getFileName();
 	if (ctx.filePath[ctx.filePath.size()-1] != '/')
 		ctx.filePath += "/";
 	std::map<uint64_t, SBlobData>::iterator meshBlobData;
 
-	core::BlobHeaderV1* const headers = (core::BlobHeaderV1*)(fileBuffer + HEADERS_FILE_OFFSET);
+	const core::BlobHeaderV1* const headers = (core::BlobHeaderV1*)(fileBuffer + HEADERS_FILE_OFFSET);
```
devshgraphicsprogramming: Just load all headers at the start, then find your mesh to load (we may support pulling named meshes or textures out of big archives).

Crisspl:
Did this one thing now. About everything else I'll talk to you in a few days because of mentioned lack of time..

devshgraphicsprogramming:
No problem, just remember to merge, I had to rip out C++11 std::free and malloc stuff from CBAWMeshWriter.cpp again

source/Irrlicht/CBAWMeshFileLoader.cpp
```diff
@@ -244,6 +247,24 @@ ICPUMesh* CBAWMeshFileLoader::createMesh(io::IReadFile* _file)
...
+
+	const char * const headerStrPattern = "IrrlichtBaW BinaryFile";
+	char headerStr[sizeof(core::BawFileV1::fileHeader)];
+	_file->read(headerStr, sizeof(headerStr));
+
+	if (strcmp(headerStr, headerStrPattern) != 0)
+		return false;
+	_ctx.fileVersion = ((uint64_t*)headerStr)[3];
+
+	_file->seek(0);
```
devshgraphicsprogramming: Don't seek to 0 after reading header

source/Irrlicht/CBAWMeshFileLoader.cpp
```diff
@@ -207,14 +207,17 @@ ICPUSkinnedMesh* CBAWMeshFileLoader::make<ICPUSkinnedMesh>(SBlobData& _data, SCo
 
 ICPUMesh* CBAWMeshFileLoader::createMesh(io::IReadFile* _file)
 {
+	SContext ctx;
+	if (!verifyFile(_file, ctx))
+		return 0;
+
 	const uint32_t filesize = _file->getSize();
```
devshgraphicsprogramming: subtract already read amount.

source/Irrlicht/CBAWMeshFileLoader.cpp
```diff
@@ -0,0 +1,274 @@
...
+core::ICPUBuffer* CBAWMeshFileLoader::make<core::ICPUBuffer>(SBlobData& _data, SContext& _ctx) const
+{
+	if (_data.wasCreated)
+		return dynamic_cast<core::ICPUBuffer*>(_ctx.createdObjects[_data.header->handle]);
+
+	typename core::CorrespondingBlobTypeFor<core::ICPUBuffer>::type* blob = core::toBlobPtr<core::ICPUBuffer>(_data.blob);
+
+	void* blobData = malloc(_data.header->blobSizeDecompr);
+	memcpy(blobData, blob->getData(), _data.header->blobSizeDecompr);
+	core::ICPUBuffer* buf = new core::ICPUBuffer(_data.header->blobSizeDecompr, blobData);
```
devshgraphicsprogramming: create a new core::ICPUBuffer(size) without a pointer and it will malloc the storage for you.

then do ctx.file->read() directly into the buf->getData()

devshgraphicsprogramming:
I think a big problem now is a lot of DRY (Don't Repeat Yourself) violation, basically every specialization of `CBAWMeshFileLoader::make`

Does 
`
if (_data.wasCreated)
    return dynamic_cast<core::ICPUBuffer*>(_ctx.createdObjects[_data.header->handle]);
`
and
`registerObject(_data, _ctx, buf);` (which I want to disappear anyway)

The function `CBAWMeshFileLoader::make` should be wrapped in a generic `loadBlob` function which should do stuff like checking if the object was already loaded/created, checksum validation (in the future decompression) and, at the end, changing the object's status to loaded.

source/Irrlicht/CBAWMeshFileLoader.cpp
```diff
@@ -0,0 +1,274 @@
...
+	setDebugName("CBAWMeshFileLoader");
+#endif
+	if (m_fileSystem)
+		m_fileSystem->grab();
+}
+
+template<>
+core::ICPUBuffer* CBAWMeshFileLoader::make<core::ICPUBuffer>(SBlobData& _data, SContext& _ctx) const
+{
+	if (_data.wasCreated)
```
devshgraphicsprogramming: this should really disappear from every `CBAWMeshFileLoader::make` and be moved into a loadBlob function or something else appropriately named

devshgraphicsprogramming:
I think I forgot to mention rule 12,13 and 14 of HPC:
12) Recursion is evil, and so are deep call-stacks, ergo rule 13
13) Don't inline your functions, but keep them inlineable
14) Eliminate deep call-stacks, by usting your own stack!

I have an idea for keeping the loading from being functions calling functions, but for now its not essential.

In a nutshell if would boil down to:
1) `Obj* SomeObjectBlob::createObj<functionToGetLoadedBlobs>(Stack<handle>& _additionalBlobsToLoad)`
2) if  `SomeObjectBlob::needsFinalization(Obj* _obj)` then put on `std::pair<Blob*,void*> _objectBlobsToFinalize`
3) load everything that is on `_additionalBlobsToLoad` with repeating steps 1-2

4) For all in `_objectBlobsToFinalize` do `SomeObjectBlob::finalizeObj<functionToGetLoadedBlobs>(Obj* _obj)`

source/Irrlicht/CBAWMeshFileLoader.h
```diff
@@ -0,0 +1,71 @@
...
+	/** @returns Pointer to the created mesh. Returns 0 if loading failed.
+	If you no longer need the mesh, you should call IAnimatedMesh::drop().
+	See IReferenceCounted::drop() for more information.*/
+	virtual ICPUMesh* createMesh(io::IReadFile* file);
+
+private:
+	bool verifyFile(io::IReadFile* _file, SContext& _ctx) const;
+
+	template<typename T>
+	T* make(SBlobData&, SContext&) const;
```
devshgraphicsprogramming: we need a more general loadBlob function to do the alreadyCreated/Loaded check, checksum check, calling make<> and tagging as loaded after load.

Due to the fact that we will have multiple versions of Blobs for IMesh etc. as we evolve, as well as the ability to serialize in/out of blobs that come from other sources such as Network etc.

I think the make function should be a function of the Blob itself.

devshgraphicsprogramming: The other argument for making this a function of the Blob, is that if someone wants to extend the loading system with their own blob types, then they won't need to touch CBAWMeshFileLoader at all or any other loading and saving systems

devshgraphicsprogramming: And another argument, killer argument... is that you do 'make<TypeOfObjectToLoad>' which will fail the moment we support loading two Blob versions for the same object

source/Irrlicht/CBAWMeshWriter.cpp
devshgraphicsprogramming: 'fixed'

source/Irrlicht/CBAWMeshFileLoader.h
devshgraphicsprogramming: 'fixed'

include/CBawFile.h
devshgraphicsprogramming: 'fixed'

source/Irrlicht/CBawFile.h
devshgraphicsprogramming: see my latest commit

include/CFinalBoneHierarchy.h
```diff
@@ -101,69 +101,70 @@ namespace scene
...
 
-			//! Function filling blob's storage with the object's data. Meant to be used by loaders only.
+			//! Function filling blob's storage with the object's data. Meant to be used by exporters only.
 			/**
 			@param _dataPtr Pointer to pre-allocated memory (heap or stack - doesn't matter) of size sufficient to hold all object's data. This size can be calculated with core::FinalBoneHierarchyBlobV1::calcBlobSizeForObj().
-			@param[out] _size Through this parameter size (in bytes) of written data is returned.
 			@see @ref CBAWMeshFileLoader CBAWMeshWriter
 			*/
-			void fillExportBlob(void* _dataPtr, uint32_t* _size)
+			size_t serializeToBlob(void* _dataPtr) const
```
devshgraphicsprogramming: can we make this a template function, so that the blob is filled inside a Blob member function?
`template <typename B>
B* serializeToBlob(void* _stackPtr=NULL, const size_t& _size=0)
{
    return VariableSizeBlob<B,CFinalBoneHierarchy>::createAndTryOnStack(this,_stackPtr,_size);
}`

And then make the 'VariableSizeBlob::createAndTryOnStack' function actually use placement new and the proper private constructor (which fills all data) to have a unified API

source/Irrlicht/CBAWMeshFileLoader.h
```diff
@@ -22,16 +23,33 @@ class CBAWMeshFileLoader : public IMeshLoader
...
+		const core::BlobHeaderV0* header;
+		size_t absOffset; // absolute
 		void* createdObj;
+		mutable bool validated;
+
+		bool validate(const void* _data) const { 
+			validated = true;
+			return validated ? true : header->validate(_data); 
+		}
+		bool isLoaded() const { return createdObj; }
```
devshgraphicsprogramming, line 35: if createdObj==0xdeadbeefdeadbeefu then loaded but not usable

source/Irrlicht/CBAWMeshFileLoader.h
```diff
@@ -22,16 +23,33 @@ class CBAWMeshFileLoader : public IMeshLoader
...
 	};
 
 	struct SContext
 	{
+		void freeLoadedObjects()
+		{
+			for (std::map<uint64_t, SBlobData>::iterator it = blobs.begin(); it != blobs.end(); ++it)
+				if (it->second.createdObj)
+					delete it->second.createdObj;
```
devshgraphicsprogramming, line 44: You can't just delete a void*

devshgraphicsprogramming:
Stop what you're doing, we need to talk

PR #36, opened 2018-02-11
BAW mesh loader the 2nd PR

Crisspl:
While coding I figured out we don't need `getNeededDeps` if we have `instantiateEmpty`. Also we don't need two stacks (in loader) about which we were talking about. Now there's just vector with blobs' data, map<handle, object ptr> and two loops performing actual loading.  
Will merge after CR.

Left todo:  
* There's some bug in loading/exporting skinned meshes. Working on it, but couldn't find anything wrong so far. It works (doesn't crash), but can't see anything or.. something weird happens. Please run the modified 7th example so you can see the effect.
* Bit fields in `SMaterial`


devshgraphicsprogramming:
hmm.. I think the reason why you don't see the skinned mesh is because finalize() (in the X loader) calculates the bounding boxes of the bones and the mesh.

But I am unsure of what else finalize() does, so we need to review it to see if it doesn't overwrite the bone hierarchies etc.

Crisspl:
Thanks for the CR :)
Here's part of the docs. Again I'll have very little time for a few following days, but docs will be coming in small batches. There's a few params empty of which I wasn't sure about. I'll ask about them when I gather more unsureness.. :D

Crisspl:
> since you already have B, the blob type in the template parameter `void* _blob` could be `B* _blob`  
  
But then I'd have to cast to it while calling in the switch what would be kinda problematic. It is doable of course, but worth it?

devshgraphicsprogramming:
Can you merge your pull against my master , because as you start commenting, you will make conflicts

Crisspl:
Can do.   
Also about *Serializable base class: there's a problem with mesh/skinned-mesh inheritance from VariableSizeBlobSerializable (ISkinnedMesh inherits IMesh), same with mesh buffers.

include/CBAWFile.h
```diff
@@ -179,63 +209,66 @@ namespace core
...
 		uint32_t indexValMax;
 		uint32_t maxVertexBoneInfluences;
 	} PACK_STRUCT;
 
 	//! Simple struct of essential data of ICPUMeshDataFormatDesc that has to be exported
-	struct MeshDataFormatDescBlobV0 : TypedBlob<MeshDataFormatDescBlobV0, scene::IMeshDataFormatDesc<core::ICPUBuffer> >
+	struct MeshDataFormatDescBlobV0 : TypedBlob<MeshDataFormatDescBlobV0, scene::IMeshDataFormatDesc<core::ICPUBuffer> >, FixedSizeBlob<MeshDataFormatDescBlobV0, scene::IMeshDataFormatDesc<core::ICPUBuffer> >
 	{
+	private:
+		enum { VERTEX_ATTRIB_CNT = 16 };
```
devshgraphicsprogramming: can we make it equal to scene::EVAI_COUNT ?

or do we run into circular refs?

include/IMeshBuffer.h
```diff
@@ -161,7 +162,7 @@ GL_INVALID_OPERATION is generated by glVertexAttribPointer if size is GL_BGRA an
 
 
 	template <class T>
-	class IMeshDataFormatDesc : public virtual IReferenceCounted
+	class IMeshDataFormatDesc : public virtual IReferenceCounted, public core::BlobSerializable<IMeshDataFormatDesc<core::ICPUBuffer> >
```
devshgraphicsprogramming: this makes the entire template class serializable, not just ICPUMeshDataFormatDesc (which is a typedef for IMeshDataFormatDesc<ICPUBuffer>)

will have to convert the ICPUMeshDataFormatDesc typedef into a class inheriting from IMeshDataFormatDesc<ICPUBuffer> and BlobSerializable

include/CBAWFile.h
Crisspl: > or do we run into circular refs?  
  
Yes, I had to forward declare every class used by blobs to make them able to inherit from classes from CBAWFile.h. So there's the best solution I came up with. I put assert (or rather `_IRR_DEBUG_BREAK_IF` so we will know 
when `EVAI_COUNT` change https://github.com/buildaworldnet/IrrlichtBAW/pull/36/commits/22ddd15def5d4041b3480288da97e5c6d7859784#diff-b7b624bc03dce20e1e4c837d54b34971R103

devshgraphicsprogramming:
> But then I'd have to cast to it while calling in the switch what would be kinda problematic. It is doable of course, but worth it?

Ok, lets keep it the way it is.

> Also about *Serializable base class: there's a problem with mesh/skinned-mesh inheritance from VariableSizeBlobSerializable (ISkinnedMesh inherits IMesh), same with mesh buffers.

Well since you need the object in order to serialize it, there is no need for the serialize method to be static.
Since there is no need for the serialize method to be static, then there is no need for static-polymorphism.
Since there is no need for static-polymorphism, the Serializable class doesn't need to be a template class with a Curiously Recurring... bla bla bla
Hence the Serializable class can be an Abstract Data Type with a single public method, a pure virtual function `serializeToBlob.`
(which yes, I know requires us to override the function in every class inheriting from Serializable).

But it could have a protected utlity template method which does exactly what the current `serializeToBlob` method does right now (have the utility method be a template, not the entire class).
At this point overriding the serializeToBlob function would become just an exact copy-paste.

include/IMeshBuffer.h
Crisspl: Yea, I know it's a todo, got it in mind

include/ICPUBuffer.h
```diff
@@ -37,10 +37,18 @@ class ICPUBuffer : public IBuffer
             size = sizeInBytes;
         }
 
+		//! Returns buffer's type which in this case is always equal to `E_BUFFER_TYPE::EBT_UNSPECIFIED_BUFFER`.
```
devshgraphicsprogramming: we initially wanted to support buffer typing (like in DX10) but OpenGL  4.0+ brought ARB_buffer_storage which makes the buffer type irrelevant, since you explicitly state if the buffer should be on the GPU or not.
Aside from that OpenGL never had strict buffer typing, the buffer type was more of a hint as to whether to store the buffer in RAM,DMA RAM or VRAM.
So this E_BUFFER_TYPE enum is deprecated.

include/ICPUBuffer.h
```diff
@@ -37,10 +37,18 @@ class ICPUBuffer : public IBuffer
...
         virtual E_BUFFER_TYPE getBufferType() const {return EBT_UNSPECIFIED_BUFFER;}
-        //! size in BYTES
+
+        //! Returns size in bytes.
         virtual const uint64_t& getSize() const {return size;}
 
+		//! Reallocates internal data. Invalidate any sizes, pointers etc. returned before!
+		/** @param newSize New size of memory.
+		@param forceRetentionOfData Doesn't matter.
```
devshgraphicsprogramming: Doesn't matter right now, but if we switch to a memory allocator that does not offer a free `realloc` then this will necessitate a memcpy.

Crisspl: It's one of things I wasn't sure about meaning. I'll ask you about this later anyway :)

include/IFrameBuffer.h
```diff
@@ -50,12 +53,31 @@ class IRenderBuffer;
 class IFrameBuffer : public virtual IReferenceCounted
 {
     public:
+		//! Attaches given texture to given attachment point.
+		/** @param attachmentPoint Identifies attchment point.
+		@param tex Texture being attached.
+		@param mipMapLayer
+		@param layer
```
devshgraphicsprogramming: layer of the framebuffer to attach to, for explanation see the OpenGL Wiki for the topic of "layered rendering"
values >= mean that a particular `layer` of a 2D or cubemap texture array, or 3D texture is attached to the FrameBuffer.
value <0 means the entire 3D texture or, 2D texture or cubemap array is bound making the FrameBuffer layered, and enabling you to use gl_Layer for layered rendering.

Look at example 15 to see layered rendering used on a cubemap to render point shadows in a single pass.

include/IFrameBuffer.h
```diff
@@ -50,12 +53,31 @@ class IRenderBuffer;
...
+		@returns Whether attachment operation went succesfully.
```
devshgraphicsprogramming: Only returns if the attachment has been attached (glFramebuffer* called) and the FBO is queued up for rebinding.
A return value of TRUE does not mean that the attachment color format is renderable or that the combination of attachments is valid.
Only after rebindRevalidate() is called by the driver internally or by the user manually do the attachments drawn into by the FrameBuffer change.

include/IFrameBuffer.h
```diff
@@ -50,12 +53,31 @@ class IRenderBuffer;
...
+		/** @param attachmentPoint Identifies attchment point.
+		@param rbf Render buffer being attached.
+		@returns Whether attachment operation went succesfully.
+		*/
         virtual bool attach(const E_FBO_ATTACHMENT_POINT &attachmenPoint, IRenderBuffer* rbf) = 0;
 
+		//! Rebinds frame buffer object.
+		/** @returns whether FBO is bound or was successfully rebound.
+		*/
         virtual bool rebindRevalidate() = 0;
```
devshgraphicsprogramming: This is a general function present in all our video namespace objects which reference/use resizeable resources such as textures and buffers.
ARB_buffer_storage and ARB_texture_storage remove the ability to respecify storage for OpenGL objects (essentially no GPU realloc) so to resize objects we need to destroy the old and create a new object, this means that the OpenGL handles (object IDs) change so the new object needs to be bound in place of the old (texture to an FBO, buffer to a VAO).
Whenever a bound object gets modified this way an atomic counter increases and this triggers the rebind* function to do more than just return `true`.

The driver always calls these functions internally before using the object with a rebind*() function.

However for the FrameBuffer you can call it manually after changing attachments as it will also call `glCheckFramebufferStatus` which will make the rebindRevalidate() function return false if the color formats you are trying to render to are invalid (like RGB9_E5) or if a certain combination of attachments is invalid (try specifying the DEPTH and STENCIL attachments to be different with any Nvidia GPU).

This function does nothing if no changes to bound objects have occurred.

include/IGPUMappedBuffer.h
```diff
@@ -17,10 +17,13 @@ namespace video
 class IGPUMappedBuffer : public virtual video::IGPUBuffer
 {
     public:
-        //! WARNING: RESIZE will invalidate pointer
-        //! WARNING: NEED TO FENCE BEFORE USE!!!!!!!!!!!!!
+		//! Gets internal pointer.
+        /** WARNING: RESIZE will invalidate pointer!
+        WARNING: NEED TO FENCE BEFORE USE!
+		@returns Internal pointer. */
```
devshgraphicsprogramming: Its best you use a GPU Fence to ensure any operations which you've queued up that are writing to this buffer or reading from it have completed before you start using this pointer, otherwise you will have a race condition.

include/IGPUMappedBuffer.h
```diff
@@ -17,10 +17,13 @@ namespace video
...
         virtual void* getPointer() = 0;
 
+		//! @returns Always true.
```
devshgraphicsprogramming, line 29: We only support persistently mapped buffers with ARB_buffer_storage.
Its almost the fastest across all cards, and is more in line with what Vulkan has.
Please don't ask us to support Buffer Orpahning, or Map/Unmap.

include/SCollisionEngine.h
```diff
@@ -15,13 +15,22 @@ class SCollisionEngine
...
+		//! Destructor.
         ~SCollisionEngine()
         {
 			for (size_t i=0; i<colliders.size(); i++)
 				colliders[i]->drop();
         }
 
 		//! Returns a 3d ray which would go through the 2d screen coodinates.
+		/**
+		@param[out] origin Attachment point of the output ray
```
devshgraphicsprogramming: start point

include/SCollisionEngine.h
```diff
@@ -55,6 +64,13 @@ class SCollisionEngine
         }
 
 		//! Calculates 2d screen position from a 3d position.
+		/**
+		@param pos 3d position which is to be projected on screen
+		@param driver Driver
+		@param camera Camera on which calculations will depend
+		@param iseViewPort Whether to use viewport or current render target's size
+		@returns {-100000, -100000} (minus ten thousand) if an error occured
```
devshgraphicsprogramming: if the point is behind the camera

include/coreutil.h
```diff
@@ -393,6 +478,7 @@ inline void XXHash_256(const void* input, size_t len, uint64_t* out)
     out[3] += v4;
 }
 
+//! Utility class easing the process of finding memory leaks. Usable only in debug build. Thread-safe.
```
devshgraphicsprogramming: Doesn't have a Windows implementation yet.

Crisspl: It's mostly not documented. I'll have to ask what `getBackTrace` actually do because i couldn't figure out.

include/CBAWFile.h
devshgraphicsprogramming: good solution

devshgraphicsprogramming:
> While coding I figured out we don't need getNeededDeps if we have instantiateEmpty. Also we don't need two stacks (in loader) about which we were talking about.

Yes, and now every blob gets loaded once and an empty object is made for every blob in the file, only to delete the objects which do no get referenced by any other one.
A stack of dependencies allows us to read 10 actually needed blobs out of 1000.

Also now the code loads the blob from the file twice (once for instantiateEmptyWithoutDeps and secondly for finalize), I believe?

Crisspl:
Ok, I didn't even consider posiblity of blob not referenced by any other. Why would there be such? Are you talking about using this as an archive file you mentioned some time?  I'll get it done later !
> Also now the code loads the blob from the file twice (once for instantiateEmptyWithoutDeps and secondly for finalize), I believe?
  
Yes, unfortunately there's the cost of simplicity :/

Crisspl:
There it is, but I can't go below two loads of blob unless we keep all the needed blobs in memory (that would also mean heap only I think) through the whole process of loading... which is not our goal either.

devshgraphicsprogramming:
Hmm.. obviously reduction of disk IO > reducing memory fragmentation.

So the heap it is.

Certain things can always be loaded and finalized at once, like ICPUBuffers and textures, so heap not necessary (any object that has 0 dependencies).

Other things, yes, i guess the heap it is then.

Obviously lets keep the beautiful createOnStack function, but for heap allocation lets pass the stackPointer to be NULL.


One last thing before I approve the pull request, can we rename `source/Irrlicht/blobsLoading.cpp` to something more conformant to our standards (camelcase, etc.)?

source/Irrlicht/CBAWMeshFileLoader.h
```diff
@@ -32,12 +32,12 @@ class CBAWMeshFileLoader : public IMeshLoader
 		SBlobData(const core::BlobHeaderV0* _hd=NULL, size_t _offset=0xdeadbeefdeadbeef) : header(_hd), absOffset(_offset), heapBlob(NULL), validated(false) {}
 		~SBlobData() { free(heapBlob); }
 		bool validate() const { 
-			validated = true;
-			return validated ? true : (heapBlob && header->validate(heapBlob)); 
+			validated = false;
+			return validated ? true : (validated = (heapBlob && header->validate(heapBlob)));
```
devshgraphicsprogramming: careful about short-circuit evaluation

source/Irrlicht/CBAWMeshWriter.cpp
```diff
@@ -186,7 +186,7 @@ namespace irr {
 
 			if (_mesh)
 			{
-				skinnedMesh = _mesh->getMeshType()!=EMT_ANIMATED_SKINNED ? NULL:reinterpret_cast<ICPUSkinnedMesh*>(_mesh); //ICPUSkinnedMesh is a direct non-virtual inheritor
+				skinnedMesh = _mesh->getMeshType()!=EMT_ANIMATED_SKINNED ? NULL:dynamic_cast<ICPUSkinnedMesh*>(_mesh); //ICPUSkinnedMesh is a direct non-virtual inheritor
```
devshgraphicsprogramming: do we need dynamic_cast instead of static_cast ?

(reinterpret is always wrong)

devshgraphicsprogramming:
ok... I shall push my latest stuff and then we merge.

PR #39, opened 2018-02-24
BAW compression + encryption(coming)

Crisspl:
Here's first approach into compression. Works fine etc., but times of loadings compressed model are much worse then the raw one (3-4 times longer) (tested ofc on release :D with this big 18MB tallguy model). However I managed to shorten this time by ~10-15% by messing with lzma properties (lzma gives more flexibility, but unfortunately is much less documented - actually no docs at all). Also a thing to do is (if possible) get to know some kind of upper bound of lzma compression size (similarly as in lz4 compression) so that we can decide whether we can compress to [preallocated] memory on stack.


devshgraphicsprogramming:
Ok, so a few remarks:
A) Interestingly enough, supposedly, lzma has roughly the same decompression speed as lz4
B) I suggest hashing of post-encryption and post-compression data, has is there only to check blob was not corrupted by the disk i/o errors

source/Irrlicht/CBAWMeshFileLoader.cpp
```diff
@@ -208,10 +211,91 @@ void* CBAWMeshFileLoader::tryReadBlobOnStack(const SBlobData & _data, SContext &
 		dst = _stackPtr;
 	else
 		dst = malloc(_data.header->blobSizeDecompr);
+
+	const bool compressed = (_data.header->compressionType != core::Blob::EBCT_RAW);
```
devshgraphicsprogramming: i dont think this is a valid way to check for compression (could have encryption instead of compression too)

source/Irrlicht/CBAWMeshFileLoader.cpp
```diff
@@ -208,10 +211,91 @@ void* CBAWMeshFileLoader::tryReadBlobOnStack(const SBlobData & _data, SContext &
...
+
+	uint8_t stackCompressed[1u<<14];
```
devshgraphicsprogramming: Stack use here is redundant and possibly problematic.
1) We can't grab more stack safely (Windows has only 1MB by default)
2) Anything under 16k should not really be compressed
3) Statistically the larger decompressed stuff will be compressed

So when compressed, either:
A) Don't try and make dst=_stackPtr and always malloc dst straight away and possibly use _stackPtr as working area for dstCompressed data
B) always malloc dstCompressed

source/Irrlicht/CBAWMeshWriter.h
```diff
@@ -21,11 +23,42 @@ namespace scene
...
+		};
+
 	private:
 		struct SContext
 		{
 			core::array<core::BlobHeaderV0> headers;
 			core::array<uint32_t> offsets;
+			const WriteProperties* props;
+		};
+		struct LzmaMemMngmnt
```
devshgraphicsprogramming: Don't have this in two places.
I know its in separate classes, but still D.R.Y. unless you have a reason for actual different implementations.

source/Irrlicht/CBAWMeshFileLoader.cpp
```diff
@@ -208,10 +211,91 @@ void* CBAWMeshFileLoader::tryReadBlobOnStack(const SBlobData & _data, SContext &
...
+
+void CBAWMeshFileLoader::SContext::regularDrop()
+{
+	for (std::map<uint64_t, SBlobData>::iterator it = blobs.begin(); it != blobs.end(); ++it)
+	{
+		const uint32_t type = it->second.header->blobType;
+		const uint64_t handle = it->second.header->handle;
+
+		using namespace core;
+		uint32_t typesNeededToBeDropped[]
```
devshgraphicsprogramming: what's this about?

dont we releaseObj everything?

source/Irrlicht/CBAWMeshWriter.cpp
```diff
@@ -12,288 +12,399 @@
...
 
-		bool CBAWMeshWriter::writeMesh(io::IWriteFile* _file, ICPUMesh* _mesh, int32_t _flags)
-		{
-			if (!_mesh || !_file)
-				return false;
+	bool CBAWMeshWriter::writeMesh(io::IWriteFile* _file, ICPUMesh* _mesh, int32_t _flags)
+	{
+		WriteProperties wp;
+		if (_flags & EMWF_WRITE_COMPRESSED)
+			wp.encryptBlobBitField = ECT_EVERYTHING;
```
devshgraphicsprogramming: make the default ECT_RAW_BUFFERS|ECT_ANIMATION_DATA|ECT_TEXTURES

source/Irrlicht/CBAWMeshWriter.h
```diff
@@ -21,11 +23,42 @@ namespace scene
 
 	class CBAWMeshWriter : public IMeshWriter
 	{
+	public:
+		enum E_COMPRESSION_TARGETS
+		{
+			ECT_NOTHING = 0x00,
+			ECT_RAW_BUFFERS = 0x01,
+			ECT_TEXTURES = 0x02,
```
devshgraphicsprogramming: what you use now for textures should really be ECT_TEXTURE_PATHS
Although ECT_TEXTURES should already be reserved

devshgraphicsprogramming:
If the loading is slow, then maybe we should up the thresholds for compression `blobLz4EncrThresh(4096), blobLzmaEncrThresh(32768)` to something higher?

I also think that compression level parameter and algo should be exposed in writeprops.

Also can you check if the hash function in lz4/xxHash is the same one as in coreutil.h?

Finally I don't think you should worry about getting an upper bound on compression size, as we already threshold to compress stuff larger than 4k and 32k and the max stack we can take without crashing people's programs is 16k.
So it's safe to assume to always allocate on the heap and not worry about doing things on the stack for compressed blobs.

source/Irrlicht/CBAWMeshFileLoader.cpp
Crisspl: It was done quite a long ago due to mem-leaks patching. Irrlicht's reference counted objects has to be dropped after setting them as other object 'ownership'. E.g.:
```
ICPUMeshBuffer* mesh = SCPUMesh();
{
ICPUMeshDataFormatDesc* desc = new ICPUMeshDataFormatDesc();
mesh->setMeshDataAndFormat(desc);
}
```
Here before `desc` goes out of scope it has to dropped so that when meshbuffer is being destroyed, `desc`'s refcount will be ==1 and meshbuffer's destructor will actually destroy it.  
But it must be done after *all* such sets (since one object can be 'owned' more then once but drop must be only one) and above applies only for reference counted objects and not even all of them... So `regularDrop` is called after all the loading. What's more.. we want to support any kind of blobs so we have to assume `releaseObj` will not do just drop but actual `delete`. That's what I was talking about while whining about this reference-counting irrlicht's thing and not some kind of `unique_ptr`.

source/Irrlicht/CBAWMeshWriter.h
Crisspl: Without this there'll be even more repeating (explicitly passing treshold values to writing function). Besides.. I don't really see repeating in this.

source/Irrlicht/CBAWMeshFileLoader.cpp
Crisspl: I know, changes will come along with encryption

Crisspl: O, they already came actually

Crisspl:
>Also can you check if the hash function in lz4/xxHash is the same one as in coreutil.h?  
  
No, it's something different, not even one-call. A few functions forming some "hashing pipeline".

source/Irrlicht/CBAWMeshFileLoader.cpp
devshgraphicsprogramming: but why do you call releaseObj only on some blobs?

source/Irrlicht/CBAWMeshWriter.h
devshgraphicsprogramming: you have one LzmaMemMngmnt for write and one LzmaMemMngmnt for load ?

Should't this be in TypedBlob or CBAWFile?

devshgraphicsprogramming:
from what I gathered by reading the source of xxHash is that it can be done in bulk or streamed mode.

Bulk consists of 1 function -- `XXH64 ()`

source/Irrlicht/CBAWMeshWriter.h
Crisspl: Oh.. I thought you were talking about props pointer :D
Sure, I just forgot about this. I'll move it to CBawFile.h

source/Irrlicht/CBAWMeshFileLoader.cpp
Crisspl: Because not all of our ref-counted objects has to be dropped (i.e. mesh through the whole loading time has ref-count==1 because no other object grabs it) and also, in case of non-reference-counted objects releaseObj would call just `delete` and we don't want it. There's just a finite set of objects that must be dropped and infinite set of possible blobs so...

devshgraphicsprogramming: createMesh is designed to only ever return one mesh... so thats the only thing not getting deleted.

loaded textures upon getTexture are added to the IVideoDriver texture cache and can neither be dropped or deleted, removeTexture must be called upon them... and I'm totally fine with cache pollution from unreferenced textures.

Since we have infinite set of possible blobs, every single one should have releaseObj called except for the top blobs being returned (like the root blob that we started loading).

Anything thats not being returned implicitly or explicitly by load routines should get destoroyed, because thats the very definition of a memory leak.

Crisspl:
Here comes encryption with a few notes as for now:  
* No MAC checking yet  
* I didn't really have a neat idea about how to do "no password" thing, so here's kinda a little monster. Please look closer at this, maybe it is ok...  
* I'll also have to correct that now passwords **must** be 16 characters long. I guess passwords should be able to be any length <=16?
  
About MACs: I had an idea of making another block in our format (before block of offsets for example; it will be 0-size block when no blob is encrypted) and also another few bytes for MAC-count. I'll have to spend a little more time to tell how complicated it will be, but for now I think it'd work fine.

devshgraphicsprogramming:
We should not care about MAC as we don't care if someone is trying to tamper with our data, like whats the worst that can happen? a dick will show up as a model?

Like we dont care about rejecting other content, we only care about that content being protected.

If we really gave a crap about real security then we'd use aes256.

Also I'm fine with requiring 16 char passwords, but if you're not then aesGladman, internally does some well-defined method of generating a passphrase (16byte) from hashing a password, it even has its own international spec number.

devshgraphicsprogramming (commented):
Does the original passworded zip file irrlicht archive loader work with new AES?

devshgraphicsprogramming:
Also now the convert2BAW tool will now have to take a password argument :)

source/Irrlicht/CBAWFile.cpp
```diff
@@ -271,4 +272,23 @@ size_t FinalBoneHierarchyBlobV0::calcNonInterpolatedAnimsByteSize() const
...
 
+bool runAes128gcm(const void* _input, size_t _inSize, void* _output, size_t _outSize, const unsigned char* _key, const unsigned char* _iv, bool _encrypt)
+{
+	const EVP_CIPHER* cipherType = EVP_aes_128_gcm();
+	EVP_CIPHER_CTX* ctx = EVP_CIPHER_CTX_new();
+
+	if (!ctx)
+		return false;
+
+	EVP_CipherInit_ex(ctx, cipherType, NULL, _key, _iv, _encrypt);
```
devshgraphicsprogramming, line 283: i wonder if there is a way to copy a freshly intialized context so we don't have to init new ones all the time?
does EVP have a clone/copy function?

source/Irrlicht/CBAWFile.cpp
```diff
@@ -271,4 +272,23 @@ size_t FinalBoneHierarchyBlobV0::calcNonInterpolatedAnimsByteSize() const
...
+	EVP_CIPHER_CTX_set_padding(ctx, 0);
+
+	int outSize = 0;
+	EVP_CipherUpdate(ctx, (unsigned char*)_output, &outSize, (unsigned char*)_input, int(_inSize));
```
devshgraphicsprogramming, line 287: shouldn't the function return outSize or at least write it to _outSize ?

source/Irrlicht/CBAWMeshFileLoader.cpp
```diff
@@ -33,7 +33,7 @@ CBAWMeshFileLoader::CBAWMeshFileLoader(scene::ISceneManager* _sm, io::IFileSyste
 
 ICPUMesh* CBAWMeshFileLoader::createMesh(io::IReadFile* _file)
 {
-	unsigned char pwd[16] = "\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0";
+	unsigned char pwd[16] = "hejkahejkahejka";
```
devshgraphicsprogramming: why are you creating an encrypted file by default?

source/Irrlicht/CBAWMeshFileLoader.h
```diff
@@ -51,16 +51,22 @@ class CBAWMeshFileLoader : public IMeshLoader
 			for (std::map<uint64_t, void*>::iterator it = createdObjs.begin(); it != createdObjs.end(); ++it)
 				loadingMgr.releaseObj(blobs[it->first].header->blobType, it->second);
 		}
-		void regularDrop();
+		void releaseAllButMesh()
+		{
+			for (std::map<uint64_t, void*>::iterator it = createdObjs.begin(); it != createdObjs.end(); ++it)
+			{
+				const uint32_t t = blobs[it->first].header->blobType;
+				if (t != core::Blob::EBT_MESH && t != core::Blob::EBT_SKINNED_MESH)
```
devshgraphicsprogramming: no.... just check against the specific blob header's data you are returning

(make the function take the createObjs iterator of the blob that you chose as the mesh to load)

source/Irrlicht/CBAWMeshWriter.cpp
```diff
@@ -340,18 +331,33 @@ namespace irr {namespace scene {
...
-			unsigned char dummy[10];
-			const unsigned char* pwd = (_ctx.pwdVer[0] == 0xff && _ctx.pwdVer[1] == 0xff) ? /* "no password" */ (const unsigned char*)"hejkahejkahejkaa" : _ctx.props->encryptionPassPhrase;
-			fcrypt_init(1, pwd, 16, _ctx.props->initializationVector, dummy, &cryptCtx);
-			fcrypt_encrypt(reinterpret_cast<unsigned char*>(data), compressedSize, &cryptCtx);
-			comprType |= core::Blob::EBCT_AES128_GCM;
-			fcrypt_end(dummy, &cryptCtx);
+			const size_t encrSize = core::BlobHeaderV0::calcEncSize(compressedSize);
+			void* in = malloc(encrSize);
+			memset(((uint8_t*)in) + (compressedSize-16), 0, 16);
+			memcpy(in, data, compressedSize);
```
devshgraphicsprogramming: if you used EVP padding properly inside core::runAes128gcm, you wouldn't need a redundant memcpy
(hell even if you can't make that work, round up the `data` malloc to 16 bytes and you'll never need padding)

devshgraphicsprogramming:
Added some comments in the latest push

include/CBAWFile.h
```diff
@@ -65,6 +95,10 @@ namespace core
 		void finalize(const void* _data, size_t _sizeDecompr, size_t _sizeCompr, uint8_t _comprType);
 		//! Calculates hash from `_data` and compares to current one (`blobHash` member).
 		bool validate(const void* _data) const;
+		//! Calculates size of blob along with required padding
+		static uint32_t calcEncSize(uint32_t _size) { return (_size/16 + 1)*16; }
```
devshgraphicsprogramming: if you want to round up to a multiple of PoT

(size+PoT-1)&(0-PoT)

tools/convert2BAW/convert2BAW/main.cpp
```diff
@@ -41,6 +43,30 @@ int main(int _optCnt, char** _options)
 				gatherWhat = EGT_INPUTS;
 			else if (core::equalsIgnoreCase("o", _options[i]+1))
 				gatherWhat = EGT_OUTPUTS;
+			else if (i+1 != _optCnt && core::equalsIgnoreCase("iv", _options[i]+1))
+			{
+				++i;
+				gatherWhat = EGT_UNDEFINED;
+				if (core::length(_options[i]) != 16)
```
devshgraphicsprogramming: the IV cannot be treated as a string.. must be 16 hex values

tools/convert2BAW/convert2BAW/main.cpp
```diff
@@ -41,6 +43,30 @@ int main(int _optCnt, char** _options)
...
+					continue;
+				}
+				memcpy(properties.initializationVector, _options[i], 16);
+				continue;
+			}
+			else if (i+1 != _optCnt && core::equalsIgnoreCase("pwd", _options[i]+1))
+			{
+				++i;
+				gatherWhat = EGT_UNDEFINED;
+				if (core::length(_options[i]) != 16)
```
devshgraphicsprogramming: the cipherkey cannot be treated as a string.. must be 16 hex values

devshgraphicsprogramming:
Also can we have an option to specify whether to save texture paths relative to the input or output files in the Convert2BAW tool? (by implication also have this option in WriteProps, i.e. if non 0 length relative path exists then save relative to that path instead of the output file)

PR #49, opened 2018-03-12
C++11 impl of things from FW_Mutex.h

Crisspl:


source/Irrlicht/FW_Mutex.h
```diff
@@ -19,7 +31,9 @@
 
 inline void FW_SleepMs(const uint64_t &milliseconds)
 {
-#ifdef WIN32
+#if defined(FW_MUTEX_H_CXX11_IMPL)
+	std::this_thread::sleep_for(std::chrono::duration<uint64_t, std::milli>(milliseconds));
```
devshgraphicsprogramming: also need a yield() equivalent for when milliseconds==0

source/Irrlicht/FW_Mutex.h
```diff
@@ -40,7 +54,9 @@ inline void FW_SleepMs(const uint64_t &milliseconds)
 //nanoseconds :D
 inline void FW_SleepNano(const uint64_t &nanoseconds)
 {
-#ifdef WIN32
+#if defined(FW_MUTEX_H_CXX11_IMPL)
+	std::this_thread::sleep_for(std::chrono::duration<uint64_t, std::nano>(nanoseconds));
```
devshgraphicsprogramming: also need a yield() equivalent for when nanoseconds==0

devshgraphicsprogramming:
Since we're porting to C++11 then it would be useful to implement FW_GetTimestampNs() with std::chrono

source/Irrlicht/FW_Mutex.h
```diff
@@ -198,7 +224,53 @@ not using add because we atomically wait for value to be 0 before swapping
 **/
 #define FW_AtomicCounterMagicBlockVal 0x1000000
 
-#if _MSC_VER && !__INTEL_COMPILER
+#if defined(FW_MUTEX_H_CXX11_IMPL)
+
+#define FW_AtomicCounter std::atomic_int
```
devshgraphicsprogramming: std::atomic is neither copyable nor movable. -- volatile int used to work for this...

but we never copy or move these

source/Irrlicht/FW_Mutex.h
```diff
@@ -198,7 +224,53 @@ not using add because we atomically wait for value to be 0 before swapping
...
+inline void FW_AtomicCounterIncr(FW_AtomicCounter &lock)
+{
+	if (lock.fetch_add(1, std::memory_order_release)+1 > FW_AtomicCounterMagicBlockVal)
```
devshgraphicsprogramming: more than two threads can use this atomic lock so please use std::memory_order_acq_rel

source/Irrlicht/FW_Mutex.h
```diff
@@ -198,7 +224,53 @@ not using add because we atomically wait for value to be 0 before swapping
...
+		while (lock >= FW_AtomicCounterMagicBlockVal) ;
+}
+inline void FW_AtomicCounterDecr(FW_AtomicCounter &lock)
+{
+	lock.fetch_sub(1, std::memory_order_acquire);
```
devshgraphicsprogramming: more than two threads can use this atomic lock so please use std::memory_order_acq_rel

source/Irrlicht/FW_Mutex.h
```diff
@@ -198,7 +224,53 @@ not using add because we atomically wait for value to be 0 before swapping
...
+}
+inline void FW_AtomicCounterBlock(FW_AtomicCounter &lock)
+{
+	int zero = 0;
+	while (!lock.compare_exchange_weak(zero, FW_AtomicCounterMagicBlockVal, std::memory_order_acquire))
```
devshgraphicsprogramming: more than two threads can use this atomic lock so please use std::memory_order_acq_rel

source/Irrlicht/FW_Mutex.h
```diff
@@ -198,7 +224,53 @@ not using add because we atomically wait for value to be 0 before swapping
...
+		zero = 0;
+}
+inline void FW_AtomicCounterDecrBlock(FW_AtomicCounter &lock)
+{
+	//to make sure we get lock first, and dont keep waiting forever because every read op has priority
+	//but we check if someone else has a read or write block, in that case we're not the only ones having the lock
+	//so lock value is more than the intended FW_AtomicCounterMagicBlockVal
+	if (lock.fetch_add(FW_AtomicCounterMagicBlockVal - 1, std::memory_order_acquire) > 1)
```
devshgraphicsprogramming: this is a lock promotion so its a release and get rolled into one... so please use std::memory_order_acq_rel

source/Irrlicht/FW_Mutex.h
```diff
@@ -198,7 +224,53 @@ not using add because we atomically wait for value to be 0 before swapping
...
+	{
+		//someone has a read or write block before we tried to swap access types
+		//so we release our lock completely and wait
+		lock.fetch_sub(FW_AtomicCounterMagicBlockVal, std::memory_order_acquire);
```
devshgraphicsprogramming: more than two threads can use this atomic lock so please use std::memory_order_acq_rel

devshgraphicsprogramming: but theoretically because it would  be sandwiched between two std::memory_order_acq_rel, we could let it be a relaxed

source/Irrlicht/FW_Mutex.h
```diff
@@ -198,7 +224,53 @@ not using add because we atomically wait for value to be 0 before swapping
...
+		//we now own no locks and wait for a free gap
+		FW_AtomicCounterBlock(lock);
+	}
+}
+inline void FW_AtomicCounterUnBlock(FW_AtomicCounter &lock)
+{
+	lock.fetch_sub(FW_AtomicCounterMagicBlockVal, std::memory_order_release);
```
devshgraphicsprogramming: more than two threads can use this atomic lock so please use std::memory_order_acq_rel

source/Irrlicht/FW_Mutex.h
```diff
@@ -198,7 +224,53 @@ not using add because we atomically wait for value to be 0 before swapping
...
+}
+inline void FW_AtomicCounterUnBlockIncr(FW_AtomicCounter &lock)
+{
+	//we had the lock value at >=FW_AtomicCounterMagicBlockVal
+	//no other write thread could have gotten out of FW_AtomicCounter***Block()
+	//because we got there before and value of lock was never 0 between this thread
+	//grabbing the write lock and now - trying to release it
+	//so we don't need to wait for write lock to be released to grab the read lock
+	lock.fetch_sub(FW_AtomicCounterMagicBlockVal-1, std::memory_order_release);
```
devshgraphicsprogramming: more than two threads can use this atomic lock so please use std::memory_order_acq_rel

source/Irrlicht/FW_Mutex.h
```diff
@@ -198,7 +224,53 @@ not using add because we atomically wait for value to be 0 before swapping
...
+}
+
+#elif _MSC_VER && !__INTEL_COMPILER
 
 #define FW_FastLock(lock) volatile long lock = 0
```
devshgraphicsprogramming, line 299: these babies also need to be ported to C++11

devshgraphicsprogramming:
You've taken out SwitchToThread() / pthread_yield() from my busywait loops, need yield back.

Although you could add a for (uint32_t i=0; (!predicate)&&i<kBusySpinsBeforeYield; i++) loop of kBusySpinsBeforeYield (preferrably 1000) iterations without a yield before the while with a yield

source/Irrlicht/CNullDriver.cpp
```diff
@@ -26,7 +26,10 @@ FW_AtomicCounter CNullDriver::ReallocationCounter = 0;
 
 FW_AtomicCounter CNullDriver::incrementAndFetchReallocCounter()
 {
-#if _MSC_VER && !__INTEL_COMPILER
+// omg this has to be rewritten
+#if defined(FW_MUTEX_H_CXX11_IMPL)
+	return ReallocationCounter += 1;
```
devshgraphicsprogramming, line 29: this ADD+STORE NEEDS to be atomic, the return type can be different/normal but the ADD needs to be ATOMIC

source/Irrlicht/FW_Mutex.h
devshgraphicsprogramming: everything below

devshgraphicsprogramming:
If we can merge the latest `master` into this pull request branch, fix the outstanding issues, then we could merge this pull request.

Then we could switch back and forth between legacy and c++11 implementation.

source/Irrlicht/CNullDriver.cpp
Crisspl: It is atomic, look at https://en.cppreference.com/w/cpp/atomic/atomic/operator_arith2

source/Irrlicht/FW_Mutex.h
Crisspl: they are, but in their `#if` branch

devshgraphicsprogramming: did we ever solve the acquire release semantics on the spinlocks?

source/Irrlicht/CNullDriver.cpp
devshgraphicsprogramming: oh overlooked the change in define for FW_AtomicCounter

but the return type of that function still needs to change to a plain int32_t

source/Irrlicht/FW_Mutex.h
Crisspl: I think so. Anyway everything should be in out comments here

source/Irrlicht/CNullDriver.cpp
Crisspl: Yea, i just noticed

PR #63, opened 2018-04-09
matrix3x4SIMD

Crisspl:
Made PR for easier review


devshgraphicsprogramming:
Great Stuff!!!

But we still need `transformPlane`, `transformBoxEx`, `makeInverse`, `buildRotateFromTo` and `buildAxisAlignedBillboard` as well as the `concatenatePrecisely*` family of functions using doubles in the intermediate representation of matrices

The `transform*` functions can be made members of plane<> and aabbox3d<> as it makes more sense for the matrix to not implement transforms of 50 different geometric primitives (and be aware of them, compile-wise).

Finally since we removed the standard rotation, we should have test which checks parity between matrix4x3::setRotationRadians() and matrix3x4SIMD::setRotation(quaternion::fromEuler())

examples_tests/06.MeshLoaders/main.cpp
```diff
@@ -94,6 +95,19 @@ class SimpleCallBack : public video::IShaderConstantSetCallBack
 
 int main()
 {
+	core::quaternion q(1, 2, 3, 4);
```
devshgraphicsprogramming: at least normalize it before doing stuff with it!!!!
(all library functions will assume normalized quaternion input in order to be faster)

include/matrix3x4SIMD.h
```diff
@@ -392,12 +500,68 @@ struct matrix3x4SIMD
...
 		return res;
 	}
+
+	inline __m128d halfRowAsDouble(size_t _n, bool _0) const
+	{
+		double tmp[2]{
+			_0 ? rows[_n].x : rows[_n].z,
+			_0 ? rows[_n].y : rows[_n].w
+		};
+		return _mm_loadu_pd(tmp);
```
devshgraphicsprogramming: instead of nonSSE float to double loadu there are instructions to promote the lower half of the register to double such as _mm_cvtps_pd

P.S. most probably need 1 more instruction for the higher part of the register to move it down for bool _0=true

include/matrix3x4SIMD.h
```diff
@@ -46,6 +46,35 @@ struct matrix3x4SIMD
...
+		__m128d r21 = _a.halfRowAsDouble(2u, 1);
+
+		matrix3x4SIMD out;
+
+		double tmp[4];
+
+		_mm_storeu_pd(tmp, matrix3x4SIMD::doJob_d(r00, r01, _b, 0));
+		_mm_storeu_pd(tmp+2, matrix3x4SIMD::doJob_d(r00, r01, _b, 1));
+		for (size_t i = 0u; i < 4u; ++i)
+			out.rows[0].pointer[i] = tmp[i];
```
devshgraphicsprogramming: there's cvtpd_ps for parallel double to float conversion

add a shuffle or movelh and you can put two registers made from double conversion into 1 register

devshgraphicsprogramming:
good, but still need to get rid of these redundant TRANSPOSEs and hadd_ps that could be replaced with add_ps (by removing a TRANSPOSE)

Also why aesGladman is now compiled in from C instead of C++ on the windows project ?

Crisspl:
>still need to get rid of these redundant TRANSPOSEs and hadd_ps

Don't undersatnd most of your comments about this. Tomorrow..

>Also why aesGladman is now compiled in from C instead of C++ on the windows project ?

Some unpushed changes apprently.. That's how it has to be (changed about time when we were doing openssl stuff) - i.e. nothing changed now, just project file was unpushed

include/matrix3x4SIMD.h
```diff
@@ -48,29 +56,25 @@ struct matrix3x4SIMD
...
-		for (size_t i = 0u; i < 4u; ++i)
-			out.rows[1].pointer[i] = tmp[i];
-		_mm_storeu_pd(tmp, matrix3x4SIMD::doJob_d(r20, r21, _b, 0));
-		_mm_storeu_pd(tmp+2, matrix3x4SIMD::doJob_d(r20, r21, _b, 1));
-		for (size_t i = 0u; i < 4u; ++i)
-			out.rows[2].pointer[i] = tmp[i];
+		const __m128 mask0011 = BUILD_MASKF(0, 0, 1, 1);
+
+		__m128 second = _mm_cvtpd_ps(matrix3x4SIMD::doJob_d(r00, r01, _b, false));
+		out.rows[0] = vectorSIMDf(_mm_cvtpd_ps(matrix3x4SIMD::doJob_d(r00, r01, _b, true))) | _mm_and_ps(_mm_movelh_ps(second, second), mask0011);
```
devshgraphicsprogramming: A single shuffle_ps could do the work of OR+AND+movehl

https://software.intel.com/sites/landingpage/IntrinsicsGuide/#!=undefined&techs=SSE,SSE2,SSE3&expand=5174,3301,3301,4791&text=shuff

include/matrix3x4SIMD.h
```diff
@@ -141,14 +155,10 @@ struct matrix3x4SIMD
...
 	}
-	inline matrix3x4SIMD& setInverseTranslation(const core::vectorSIMDf& _translation)
-	{
-		return setTranslation(-_translation);
-	}
 	inline vectorSIMDf getTranslation() const
 	{
 		__m128 xmm0 = _mm_unpackhi_ps(rows[0].getAsRegister(), rows[1].getAsRegister()); // (0z,1z,0w,1w)
-		__m128 xmm1 = _mm_unpackhi_ps(rows[2].getAsRegister(), rows[3].getAsRegister()); // (2z,3z,2w,3w)
+		__m128 xmm1 = _mm_unpackhi_ps(rows[2].getAsRegister(), _mm_setr_ps(0.f, 0.f, 0.f, 1.f)); // (2z,3z,2w,3w)
```
devshgraphicsprogramming: this is beautiful

include/matrix3x4SIMD.h
```diff
@@ -306,55 +317,58 @@ struct matrix3x4SIMD
 		return *this;
 	}
 
-	//! @bug @todo not fine
 	inline bool getInverse(matrix3x4SIMD& _out) const
```
devshgraphicsprogramming: why so many core::transpose4 ?

include/matrix3x4SIMD.h
```diff
@@ -306,55 +317,58 @@ struct matrix3x4SIMD
...
-		_out.rows[2] = te;
+		if (core::iszero(d.x, FLT_MIN))
+			return false;
 
-		__m128 rdet = _mm_rcp_ps(determinant4.getAsRegister());
-		_out.rows[0] *= rdet;
-		_out.rows[1] *= rdet;
-		_out.rows[2] *= rdet;
+		reinterpret_cast<vectorSIMDf*>(_out)[0] = c1crossc2 / d;
+		reinterpret_cast<vectorSIMDf*>(_out+3)[0] = (c2.crossProduct(c0)) / d;
```
devshgraphicsprogramming: this will crash, vectorSIMDf has a memory alignmend requirement of 16 bytes, plus you can't guarantee _out is 16 byte aligned in the first place

include/matrix3x4SIMD.h
```diff
@@ -503,25 +517,21 @@ struct matrix3x4SIMD
...
-		__m128d r1 = _mtx.halfRowAsDouble(1u, _0);
-		__m128d r2 = _mtx.halfRowAsDouble(2u, _0);
+		__m128d r0 = _mtx.halfRowAsDouble(0u, _xyHalf);
+		__m128d r1 = _mtx.halfRowAsDouble(1u, _xyHalf);
+		__m128d r2 = _mtx.halfRowAsDouble(2u, _xyHalf);
 
 		const __m128d mask01 = _mm_castsi128_pd(_mm_setr_epi32(0, 0, 0xffffffff, 0xffffffff));
 
 		__m128d res;
 		res = _mm_mul_pd(_mm_shuffle_pd(_a0, _a0, 0), r0);
```
devshgraphicsprogramming: could you explain the shuffles to me?

Crisspl: Similarly to float32 version. `_a0` and `_a1` are two halves of a row. So:  
* `_mm_shuffle_pd(_a0, _a0, 0)` will give us first element of first half broadcasted (i.e. x brodcasted)  
* `_mm_shuffle_pd(_a0, _a0, 3)` will give us second element of first half broadcasted (y broadcasted). `3` comes from the fact that 3==0b11  
  
Same for second half. The trailing `if` is because in first half it'd have no sense since first half of 3rd row is {0, 0} and so we would have to do something like _mm_add_pd(res, _mm_and_pd(_a1, **mask00**));

include/quaternion.h
```diff
@@ -23,12 +23,15 @@ namespace irr
...
 {
 
+struct matrix3x4SIMD;
 
 //! Quaternion class for representing rotations.
 /** It provides cheap combinations and avoids gimbal locks.
 Also useful for interpolations. */
 class quaternion : private vectorSIMDf
 {
+	friend struct matrix3x4SIMD;
```
devshgraphicsprogramming: don't add this, make quaternion oblivious of matrix3x4SIMD

you can reinterpret_cast quaternion to vectorSIMDf very safely

include/matrix3x4SIMD.h
devshgraphicsprogramming: I was just confused as to the 0b11 shuffle mask for a shuffle_pd

https://software.intel.com/sites/landingpage/IntrinsicsGuide/#!=undefined&techs=SSE,SSE2,SSE3&expand=5174,3301,3301,4780,4791&text=shuffle

Crisspl: tested it. Works

include/matrix3x4SIMD.h
```diff
@@ -213,12 +213,20 @@ struct matrix3x4SIMD
 
 	inline void transformVect(float* _out, const float* _in) const
 	{
-		vectorSIMDf c0 = rows[0], c1 = rows[1], c2 = rows[2], c3(0.f, 0.f, 0.f, 1.f);
-		core::transpose4(c0, c1, c2, c3);
+		vectorSIMDf vec(_in);
+		vectorSIMDf r0 = rows[0] * vec, 
+			r1 = rows[1] * vec, 
+			r2 = rows[2] * vec, 
+			r3 = vec & BUILD_MASKF(0, 0, 0, 1);
```
devshgraphicsprogramming: with 3 hadds this will go in the res[3] element which will never be memcopied out... as this is the non 4x1 function maybe for this specific instance (mul with 3x1 vector) you should use the default zero vector for r3

include/matrix3x4SIMD.h
```diff
@@ -389,25 +408,28 @@ struct matrix3x4SIMD
...
+		//ttmp = _mm_hadd_ps(ttmp, ttmp);
+		//col3.y = vectorSIMDf(ttmp).x;
+
+		//tmp = (-vectorSIMDf(r2) & mask1110) * _center;
+		//ttmp = tmp.getAsRegister();
+		//ttmp = _mm_hadd_ps(ttmp, ttmp);
+		//ttmp = _mm_hadd_ps(ttmp, ttmp);
+		//col3.z = vectorSIMDf(ttmp).x;
+		
+		//result slightly (but significantly) differ from above.. why?
```
devshgraphicsprogramming: there's no fast way to insert a custom translation into a row matrix, so we must do it simply in non-SIMD fashion (you could do two transposes but thats simply slower).

there's no need for a transpose
so  multiply all rows by center to obtain `r_i = ri*_center`
then use 3 hadds to reduce the 3 rows (and a random register) into a single 4x1 vector `col3`
then get `finalCol3 = _center-_translation-col3`
you can then save the 3 out of 4 components of finalCol3 to elements 3,7,11 of the matrix with regular C memory ops

include/matrix3x4SIMD.h
```diff
@@ -389,25 +408,28 @@ struct matrix3x4SIMD
...
+		col3 = (-vectorSIMDf(c0) - c1 - c2) * _center;
+
+		col3 += (_center - _translation) & mask1110;
```
devshgraphicsprogramming: there's no need for the mask as the final component doesn't get saved as there is no row[3]

include/matrix3x4SIMD.h
```diff
@@ -389,25 +408,28 @@ struct matrix3x4SIMD
...
+		//col3.y = vectorSIMDf(ttmp).x;
+
+		//tmp = (-vectorSIMDf(r2) & mask1110) * _center;
+		//ttmp = tmp.getAsRegister();
+		//ttmp = _mm_hadd_ps(ttmp, ttmp);
+		//ttmp = _mm_hadd_ps(ttmp, ttmp);
+		//col3.z = vectorSIMDf(ttmp).x;
+		
+		//result slightly (but significantly) differ from above.. why?
+		col3 = (-vectorSIMDf(c0) - c1 - c2) * _center;
```
devshgraphicsprogramming: this is wrong, it should have been `-c0*_center.x-c1*_center.y-c2*_center.z`

PR #64, opened 2018-04-12
Fix line endings

numberZero:
`git diff master fix-eol --ignore-space-change` shows there are no code changes.


devshgraphicsprogramming:
rather than changing all our files and requesting a massive pull...

could we get a .sh script that fixes up the line endings (into UNIX not windows) and replaces tabs by 4 spaces?

P.S. Need to fix my IDE to go into space mode, been putting it off for waaay too long.

numberZero:
You’re definitely right) here is it:
```
find -name '*.h' -o -name '*.cpp' | xargs sed -i 's/\r$//g; s/\t/    /g'
```

Crisspl:
>Need to fix my IDE to go into space mode, been putting it off for waaay too long.
  
Uh i'll do it too then. Code looks really crappy sometimes on github because of tabs vs spaces duel.

devshgraphicsprogramming:
4 spaces are good, tabs are a bit bad... not as a normal shit-throwing fest of TABS vs. Spaces but simply of the Meld, TortoiseGit, CodeBlocks and Visual Studio mix that we use renders spaces to be roughly the same indent size.

@numberZero I know what the find does, but can you explain the role of the sed arguments?

numberZero:
@devshgraphicsprogramming 

`-i` means “edit files in-place” (by default `sed` writes everything to stdout, even when processing multiple files)

The next argument is the program, it has 2 “search-and-replace” statements in this case:
```
s/\r$//g
```
replaces `\r` that is just before line end with nothing (lines end in `\r\n` on Windows but with `\n` on UNIX). (It actually doesn’t need the trailing `g`)
 
```
s/\t/    /g
```
replaces tab with 4 spaces; `g` is a flag to make it replace each tab on the line (by default, `s` does one replace per line).

The following arguments `sed` expect are file names, `xargs` supplies them.

*P.S. Tabs are great!*

devshgraphicsprogramming:
Thanks for the `sed` explanation, will make a .sh script in the top directory for this.
A tested .bat file for windows would be welcome

P.S. I know `xargs` XD

PR #81, opened 2018-04-30
CMake

numberZero:
This may not be the best CMake file possible, but it works, at least on my system.

### DIFF
--- a/source/Irrlicht/CMakeLists.txt
+++ b/source/Irrlicht/CMakeLists.txt
@@ -334,10 +334,13 @@ include_directories(
 	zlib
 )
 
-add_definitions(
-	$<$<CONFIG:Debug>:_DEBUG>
-	$<$<BOOL:FAST_MATH>:IRRLICHT_FAST_MATH>
-)
+if($<CONFIG:Debug>)
+	add_definitions(_DEBUG)
+endif()
+
+if($<BOOL:FAST_MATH>)
+	add_definitions(IRRLICHT_FAST_MATH)
+endif()
 
 add_library(Irrlicht STATIC
 	${IRRLICHT_SRCS_COMMON}
--- a/source/Irrlicht/CMakeLists.txt
+++ b/source/Irrlicht/CMakeLists.txt
@@ -383,3 +383,17 @@ target_compile_definitions(IrrlichtServer PRIVATE
 	NO_IRR_COMPILE_WITH_BURNINGSVIDEO_
 	NO_IRR_COMPILE_WITH_OPENGL_
 )
+
+if(NOT CONFIGURED_ONCE)
+	# Add some optional flags to be used by default
+	if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
+		set(CMAKE_C_FLAGS_DEBUG "${CMAKE_C_FLAGS_DEBUG} -ggdb -g3"
+			CACHE STRING "Flags used by the C compiler during debug build." FORCE)
+		set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -ggdb -g3"
+			CACHE STRING "Flags used by the C++ compiler during debug build." FORCE)
+	endif()
+
+	# This is to add flags once only
+	set(CONFIGURED_ONCE TRUE
+		CACHE INTERNAL  "A flag showing that CMake has configured at least once.")
+endif()
--- a/source/Irrlicht/CMakeLists.txt
+++ b/source/Irrlicht/CMakeLists.txt
@@ -7,35 +7,84 @@ if(NOT CMAKE_BUILD_TYPE)
 	set(CMAKE_BUILD_TYPE Release CACHE STRING "Build type: Debug or Release" FORCE)
 endif()
 
-option(FAST_MATH "Enable fast math" ON)
+option(FAST_MATH "Enable fast low-precision math" ON)
 
 find_package(OpenSSL REQUIRED)
 
-set(Irrlicht_SRCS
+set(IRRLICHT_SRCS_COMMON
+# Mesh loaders
 	C3DSMeshFileLoader.cpp
+	CB3DMeshFileLoader.cpp
+	CBAWMeshFileLoader.cpp
+	CCSMLoader.cpp
+	CDMFLoader.cpp
+	CLMTSMeshFileLoader.cpp
+	CLWOMeshFileLoader.cpp
+	CMS3DMeshFileLoader.cpp
+	CMY3DMeshFileLoader.cpp
+	COBJMeshFileLoader.cpp
+	COCTLoader.cpp
+	COgreMeshFileLoader.cpp
+	CPLYMeshFileLoader.cpp
+	CSMFMeshFileLoader.cpp
+	CSTLMeshFileLoader.cpp
+	CXMeshFileLoader.cpp
+
+# Mesh writers
+	CBAWMeshWriter.cpp
+	COBJMeshWriter.cpp
+	CPLYMeshWriter.cpp
+	CSTLMeshWriter.cpp
+
+# Other mesh-related stuff
 	CAnimatedMeshHalfLife.cpp
 	CAnimatedMeshSceneNode.cpp
-	CB3DMeshFileLoader.cpp
 	CBAWFile.cpp
-	CBAWMeshFileLoader.cpp
-	CBAWMeshWriter.cpp
-	CBillboardSceneNode.cpp
 	CBlobsLoadingManager.cpp
-	CBurningShader_Raster_Reference.cpp
+	CMeshSceneNode.cpp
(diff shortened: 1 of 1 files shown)


devshgraphicsprogramming:
Great, but the windows target also needs some of the GCC flags or equivalent, like fno-exceptions etc.

Also there should be 4 targets:
OpenGL {Release,Debug} outputs (lib)Irrlicht(.a)
Server {Release,Debug} outputs (lib)IrrlichtServer(.a)
    same as current makefile when invoked with "make serverlib BAW_SERVER=1 NEW__MESHES_DONE=1"

c.f. with the codeblocks project file (XML) for compile options, defines and linker options under linux


Also what do we invoke CMake with and how do we compile?
What files will CMake generate/overwrite?

devshgraphicsprogramming:
Also forgot to mention that your CMake file should have a configurable path to the openssl includes and libraries (crypto & ssl) 
In our case we presume to be in ../../../openssl relative to ./source/Irrlicht, but it would be nice to either let it be some other path or system default.

numberZero:
> the windows target also needs some of the GCC flags

MinGW gets the flags already, as CMake recognizes it as GCC. If you mean MSVC, well, I need to check the VC project files then as I don’t remember their flags at all.

> Also there should be 4 targets:
OpenGL {Release,Debug} outputs (lib)Irrlicht(.a)
Server {Release,Debug} outputs (lib)IrrlichtServer(.a)

Will add the Server target. Debug/release is what CMake cares of.

> Also what do we invoke CMake with and how do we compile?

Make an empty directory anywhere, enter it, call `cmake <path-to-the-source>` (that is, to the directory with the main `CMakeLists.txt` file; that should probably be the repository root and not where I put it currently), then `make`. It will build the default configuration. You can supply additional arguments to CMake like `-DCMAKE_BUILD_TYPE=Debug` or `-DFAST_MATH=OFF` to use different configuration. You can also call `cmake-gui` instead of `cmake`, if you have it; it is handy.

> What files will CMake generate/overwrite?

It works only in the folder you created for it, not changing anything outside it.

> Also forgot to mention that your CMake file should have a configurable path to the openssl includes and libraries (crypto & ssl)
In our case we presume to be in ../../../openssl relative to ./source/Irrlicht, but it would be nice to either let it be some other path or system default.

Oh, forgot that, sorry. Will add, that’s just a few lines in CMake.

numberZero:
By the way, where is `MacOSX/CIrrDeviceMacOSX.mm` mentioned in the CodeBlocks project?

devshgraphicsprogramming:
There is no MacOSX target at all.

Search in COpenGLDriver.h for a nicely named structure which is comment about Apple :)

devshgraphicsprogramming:
>  (that is, to the directory with the main CMakeLists.txt file; that should probably be the repository root and not where I put it currently)

I think its good where it is, the root should have a global CMake that would make examples too (and have test targets)

source/Irrlicht/CMakeLists.txt
```diff
@@ -9,6 +9,8 @@ endif()
 
 option(FAST_MATH "Enable fast math" ON)
 
+find_package(OpenSSL REQUIRED)
```
devshgraphicsprogramming: Will this cmake let me specify custom openssl path? (so I dont compile and link against my system libs)

Like this?
https://cmake.org/pipermail/cmake/2011-June/045116.html

numberZero: Yes, but note that `OPENSSL_LIBRARIES` is not a valid configuration option; use `OPENSSL_CRYPTO_LIBRARY` (`-DOPENSSL_CRYPTO_LIBRARY=/path/to/crypto.so`) and `OPENSSL_SSL_LIBRARY` instead. `cmake-gui` shows the available configuration options (in terminal, use `cmake -LAH <path-to-build-directory>` instead)

devshgraphicsprogramming: what about where it will get include files for openssl from?

numberZero: @devshgraphicsprogramming from `OPENSSL_INCLUDE_DIR`. On my system, that’s just `/usr/include` by default, so that you can write e.g. `openssl/aes.h` to reference `/usr/include/openssl/aes.h`. That’s the way libraries are used with CMake.

numberZero:
@devshgraphicsprogramming `Makefile` misses the following files:
```
CGPUTransientBuffer.cpp
COpenCLHandler.cpp
COpenGL2DTextureArray.cpp
COpenGL2DTexture.cpp
COpenGL3DTexture.cpp
COpenGLCubemapTexture.cpp
COpenGLMultisampleTextureArray.cpp
COpenGLMultisampleTexture.cpp
COpenGLOcclusionQuery.cpp
COpenGLPersistentlyMappedBuffer.cpp
COpenGLQuery.cpp
COpenGLTransformFeedback.cpp
COpenGLVAO.cpp
coreutil.cpp
CTRTextureGouraudAlpha.cpp
CTRTextureGouraudAlphaNoZ.cpp
```
but it *does* include other similar files. Is that intentional?

devshgraphicsprogramming:
its most probably bad that it misses
`coreutil.cpp
CGPUTransientBuffer.cpp`

but the Makefile only supports a headless server target so no OpenGL or graphics output device of any kind thats why.

What surplus files does it include?

devshgraphicsprogramming:
Anyhow, far off is this thing from being ready?

devshgraphicsprogramming:
You're missing the O3 flag, as well as linker flags (some compiler flags need to be repeated as linker flags) such as mfp-math=sse, fuse=gold, etc.

For gdb to work reliably not only does Debug build require -g but also -ggdb3

Again c.f. with codeblocks project (you can see compile and linker flags separately)

And you're missing the essential define NEW__MESHES_DONE, again .c.f. with codeblocks project

numberZero:
> You're missing the O3 flag, as well as linker flags (some compiler flags need to be repeated as linker flags) such as mfp-math=sse, fuse=gold, etc.

Can’t find any documentation on `mfp-math=sse` as linker flag, although it seems to be used when link-time optimization is in effect (it is disabled by default). On `fuse=gold`, is that mandatory or recommended?

> For gdb to work reliably not only does Debug build require -g but also -ggdb3

That should probably be set as default and not as required value then.

> And you're missing the essential define NEW__MESHES_DONE, again .c.f. with codeblocks project

Can’t find *any* mention of `NEW__MESHES_DONE`, where is it used? I see `NEW_MESHES` only.

numberZero:
> What surplus files does it include?

The makefile includes, among others,

* COpenGLDriver.o
* COpenGLTexture.o
* COpenGLSLMaterialRenderer.o
* COpenGLExtensionHandler.o
* COpenGLFrameBuffer.o
* COpenGLRenderBuffer.o

what looks very strange for a headless target.

devshgraphicsprogramming:
`fuse=gold` is mandatory, its the "new" multi-threaded linker in the GCC toolchain.

> That should probably be set as default and not as required value then.

You'd be surprised, unless CMake states that explicitly somewhere I'd bank on ggdb3 not being enabled (I wonder if -g is set correctly too).

> Can’t find any mention of NEW__MESHES_DONE, where is it used? I see NEW_MESHES only.

Granted, but you're still missing NEW_MESHES as a define.


On the topic of extra OpenGL files, lets leave them in for now, `NO_IRR_COMPILE_WITH_OPENGL_` takes care of not compiling in OpenGL.
(I'll open a separate issue to clean that up.


Let me know if the CMake is tested and works.

devshgraphicsprogramming:
Please add the `NEW_MESHES` define to the CMake

If you confirm that the CMake not only builds, but that the result links with an example, then I can merge.

I can merge pull and you can add information to the Wiki
https://github.com/buildaworldnet/IrrlichtBAW/wiki/Building-The-Library

numberZero:
> For gdb to work reliably not only does Debug build require -g but also -ggdb3

Added it as default for GCC debug builds. Clang doesn’t support that flag AFAIK, and there are other debuggers (LLDB) out there that may better work with other configuration.

> fuse=gold is mandatory

Just realized you ask to add linking flags while linking is not done at all; static libraries (on Linux at least) are just archives consisting of object files.

On `NEW_MESHES`, sorry, forgot to mention it is defined *unconditionally* in `IrrCompileConfig.h`.

> If you confirm that the CMake not only builds, but that the result links with an example, then I can merge.

It does. Of course linking requires some flags like `-lcrypto` as dependencies are not propagated by static libraries (CMake is able to do that job, if only it is used to build the program as well; that’s not the case in this PR), but with these it works fine.

P.S. It appears my old GPU is waaay too old to use IrrlichtBAW. I plan to replace it very soon, but until then, I can only test it with altered GL version check (as Mesa exposes GL 3.3 only). Well, HelloWorld works, at least)

devshgraphicsprogramming:
Go to town on the wiki page ;)

Issue #123, opened 2018-07-25
Abstract all resource caches with CObjectCache and CConcurrentObjectCache


devshgraphicsprogramming:
Texture, Multisample Texture, Framebuffers, Meshes and SceneNodes are all cached by the engine either by some sort of a name string or just to make sure we released all Graphics API resources.

While I believe there should be no need to cache MS Textures and Framebuffers, developers could find it useful to use a dummy cache just so that they don't need to write de-initialization by hand.

`CObjectCache` would be a template class in two partial specialization variants, one taking a object type template parameter (which much inherit from IReferenceCounted) and  a container template parameter (so one can use `std::vector`, `std::unordered_map` etc.), while the other does all of the above plus takes a template parameter for the key type (std::string etc.).

`CConcurrentObjectCache` would be just the same as `CObjectCache` with the exception of using slim read write locks for safe insertion, retrieval and deletion in the cache.

devshgraphicsprogramming:
Some resources require more involved delete mechanism rather than drop, such as ITexture .

devshgraphicsprogramming:
A functional in the constructor of the cache could be fun.

devshgraphicsprogramming:
Grab/Greet only after we know addition to cache is possible, i.e. after
https://github.com/buildaworldnet/IrrlichtBAW/blob/master/include/CObjectCache.h#L104
Add a greet function, do not assume type is IReferenceCounted


Note to self (I'll do these myself):
1) allow insertion of nullptr values for keys
2) Change `T* getByKey(const K& _key)` to `bool getByKey(T* &_outVal,const K& _key)` beause of nullptr allowed
3) ~Rework `contains` to use `getByKey`~
4) Add a `bool swapValue(const K& _key, T* _swapFor)`
5) Add a `bool getByKeyOrReserve(T* &_outVal, const K& _key)` which returns true and puts the found value in _outVal or false and returns nullptr in _outVal indicating that a nullptr was inserted for that key (this will require either write lock or lock upgrade)

These changes will allow us to avoid duplication of work scenario, like this one:
```
//Thread 1 starts loading "castle.obj"
//Thread 2 starts loading "castle.obj"
Thread 1 queries cache and find nothing
Thread 2 queries cache and find nothing
Thread 1 loads model and puts it in cache, cache grabs
Thread 2 loads model and tries to put it in cache, failing
Thread 1 drops model, cache retains last reference
[...]

PR #143, opened 2018-09-05
MultiObjectCache

Crisspl:


devshgraphicsprogramming:
It would be nice to unify the methods for all cache types.

the single-object caches can just return 1-sized ranges

Crisspl:
That's kind of stuff you should tell me before. It changes approach like a lot, but i'll see tomorrow how we can do that

devshgraphicsprogramming:
Well at least these classes are not being used anywhere, yet.

> That's kind of stuff you should tell me before.

Agreed. However it would be beneficial to commit the intended function declarations first (or interfaces), then we could discuss the approach before the trouble of writing definitions and implementations.

include/CObjectCache.h
```diff
@@ -9,58 +9,110 @@
...
+    struct is_multi_container : std::false_type {};
+    template<>
+    struct is_multi_container<std::multimap> : std::true_type {};
+    template<>
+    struct is_multi_container<std::unordered_multimap> : std::true_type {};
+
+    template<template<typename...> class T>
+    struct is_assoc_container : std::false_type {};
+    template<>
+    struct is_assoc_container<std::map> : std::true_type {};
```
devshgraphicsprogramming: whats the reason behind not adding multimap and unordered_multimap to the is_assoc_container true set?

include/CObjectCache.h
```diff
@@ -9,58 +9,110 @@
...
+    template<>
+    struct is_assoc_container<std::unordered_map> : std::true_type {};
+
+    struct Dummy {};
```
devshgraphicsprogramming: is this doing anything?

include/CObjectCache.h
```diff
@@ -9,58 +9,110 @@
...
 #include <map>
 #include <unordered_map>
-#include "IReferenceCounted.h"
 
 namespace irr { namespace core
 {
 
 namespace impl
 {
+    struct CMultiCache_tag {};
```
devshgraphicsprogramming: remember about the FORCE_EBO macro (otherwise empty base optimization wont happen for msvc)

include/CObjectCache.h
```diff
@@ -9,58 +9,110 @@
...
+    template<template<typename...> class T>
+    struct is_assoc_container : std::false_type {};
+    template<>
+    struct is_assoc_container<std::map> : std::true_type {};
+    template<>
+    struct is_assoc_container<std::unordered_map> : std::true_type {};
+
+    struct Dummy {};
+
+#define DEF_CACHE_CTORS(_Name, _Base) \
```
devshgraphicsprogramming: do you know about the C++11 constructor inheritance?
```c++
class Derived : public Base
{
public:
   using Base::Base; //inherits all of Base's constructors
}
```

include/CObjectCache.h
```diff
@@ -70,191 +122,442 @@ namespace impl
...
-    using ValueType = std::pair<K, T*>;
+            return true;
+        }
+    };
+    template<template<typename...> class ContainerT_T, typename ContainerT>
+    struct CPreInsertionVerifier<ContainerT_T, ContainerT, false, true>
+    {
+        static bool verify(const std::pair<typename ContainerT::iterator, bool>& _insertionRes)
+        {
+            return _insertionRes.second;
```
devshgraphicsprogramming, line 159: I don't understand this, why is a null `second` invalid?

Reference:
https://github.com/buildaworldnet/IrrlichtBAW/issues/123#issuecomment-409851961
https://github.com/buildaworldnet/IrrlichtBAW/commit/ed1029bd1995e07686dbd57c212d3a1ae8281fac

devshgraphicsprogramming:
I allowed the insertion and caching of null pointers a month ago

https://github.com/buildaworldnet/IrrlichtBAW/commit/ed1029bd1995e07686dbd57c212d3a1ae8281fac#diff-d57fe204ada75d53d408f90f1606d69bR103

Also referenced it in the issue #123 

The non-nullness of the returned pointer is no longer an indicator of success.

Greet and Dispose are always called, even on nullptr... they are supposed to check for it and handle that.

include/CObjectCache.h
Crisspl: Because it doesn't matter `CPreInsertionVerifier<>::verify()` for multicaches always returns `true` disregarding container type. But i'll add them just for logic sense.

include/CObjectCache.h
Crisspl: Seems like nope, i don't remember what it was so.. it's gone

include/CObjectCache.h
Crisspl: Memory layout doesn't matter here. Or do you care? I can add if you do

include/CObjectCache.h
Crisspl: Forgot about this, i'll change

include/CObjectCache.h
Crisspl: `second` is `bool` here. It's the pair returned by non-multi assoc container from `std`

Crisspl:
> I allowed the insertion and caching of null pointers a month ago

I know, everything i'm aware of is up to date with your changes

include/CObjectCache.h
devshgraphicsprogramming: As a matter of principle we should always enable empty base optimization on empty structs and classes

devshgraphicsprogramming:
ok I'll review code once again

include/CObjectCache.h
```diff
@@ -9,58 +9,110 @@
...
         ContainerT m_container;
 
-        using ValueType = typename ContainerT_T<K..., T>::value_type::second_type; // container's value_type is always instantiation of std::pair
-        using NoPtrValueType = typename std::remove_pointer<ValueType>::type;
-        using PtrToConstVal_t = const NoPtrValueType*; // ValueType is always pointer type
+        // typedefs for implementation only
+        using PairType_impl = typename ContainerT::value_type;
+        //! Always pointer type
+        using ValueType_impl = typename PairType_impl::second_type; // container's value_type is always instantiation of std::pair
+        static_assert(std::is_pointer<ValueType_impl>::value, "ValueType_impl must be pointer type!");
```
devshgraphicsprogramming, line 84: What are the limitations to using non-pointer types? (as well as pointer types)

include/CObjectCache.h
```diff
@@ -9,58 +9,110 @@
...
+        using KeyType_impl = typename PairType_impl::first_type;
+        using NoPtrValueType_impl = typename std::remove_pointer<ValueType_impl>::type;
```
devshgraphicsprogramming, line 86: you have no use for these 2 "no pointer" types.

include/CObjectCache.h
```diff
@@ -9,58 +9,110 @@
...
         {
             if (m_greetingFunc)
                 m_greetingFunc(_object);
         }
 
     public:
-        inline explicit CObjectCacheBase(const std::function<void(ValueType)>& _greeting, const std::function<void(ValueType)>& _disposal) : m_greetingFunc(_greeting), m_disposalFunc(_disposal) {}
-        inline explicit CObjectCacheBase(std::function<void(ValueType)>&& _greeting, std::function<void(ValueType)>&& _disposal) : m_greetingFunc(std::move(_greeting)), m_disposalFunc(std::move(_disposal)) {}
+        inline explicit CObjectCacheBase(const GreetFuncType& _greeting, const DisposalFuncType& _disposal) : m_greetingFunc(_greeting), m_disposalFunc(_disposal) {}
+        inline explicit CObjectCacheBase(GreetFuncType&& _greeting, DisposalFuncType&& _disposal) : m_greetingFunc(std::move(_greeting)), m_disposalFunc(std::move(_disposal)) {}
```
devshgraphicsprogramming, line 122: a move constructor would require you to zero out or invalidate the input reference.

Also why not copy and move constructors for the cache base itself?

include/CObjectCache.h
```diff
@@ -9,58 +9,110 @@
...
 
-        inline bool contains(PtrToConstVal_t _object) const
+        inline bool contains(ValueType_PtrToConst_impl _object) const
```
devshgraphicsprogramming, line 124: could be a `const typename value_type`

include/CObjectCache.h
devshgraphicsprogramming: ok I see, it was just a bit hard to read.

include/CObjectCache.h
```diff
@@ -70,191 +122,433 @@ namespace impl
...
-class CObjectCache<K, T, ContainerT_T, true> : public impl::CObjectCacheBase<ContainerT_T, std::pair<K, T*>>
-{
-    using ValueType = std::pair<K, T*>;
+            return true;
+        }
+    };
+    template<template<typename...> class ContainerT_T, typename ContainerT>
+    struct CPreInsertionVerifier<ContainerT_T, ContainerT, false, true>
+    {
+        static bool verify(const std::pair<typename ContainerT::iterator, bool>& _insertionRes)
```
devshgraphicsprogramming, line 157: single cache `insert` only returns a `bool` how is this overload/specialization going to work?

Have you tested this with the different container types? (even instantiation?)

include/CObjectCache.h
```diff
@@ -70,191 +122,433 @@ namespace impl
...
+        {
+            return _insertionRes.second;
+        }
+    };
 
-public:
-    inline explicit CObjectCache(const std::function<void(T*)>& _greeting, const std::function<void(T*)>& _disposal) : impl::CObjectCacheBase<ContainerT_T, std::pair<K, T*>>(_greeting, _disposal) {}
-    inline explicit CObjectCache(std::function<void(T*)>&& _greeting = nullptr, std::function<void(T*)>&& _disposal = nullptr) : impl::CObjectCacheBase<ContainerT_T, std::pair<K, T*>>(std::move(_greeting), std::move(_disposal)) {}
+    //! Use in non-static member functions
+#define MY_TYPE typename std::remove_reference<decltype(*this)>::type
```
devshgraphicsprogramming: this silly macro impacts readability

devshgraphicsprogramming:
The operation of `CPreInsertionVerifier` is really hard to decipher without comments.

The macros don't help either.

include/CObjectCache.h
```diff
@@ -70,191 +122,433 @@ namespace impl
...
+    return true;
+#define INSERT_IMPL_ASSOC \
+    auto res = this->m_container.insert({ _key, _val });\
+    const bool verif = impl::CPreInsertionVerifier<ContainerT_T, typename Base::ContainerT, std::is_base_of<impl::CMultiCache_tag, MY_TYPE>::value>::verify(res);\
+    if (verif)\
+        greet(_val);\
+    return verif;
+
+    template<
+        bool isMultiContainer,
```
devshgraphicsprogramming: this template paramter name is misleading, I think you meant `isNotVectorBased`

include/CObjectCache.h
```diff
@@ -70,191 +122,433 @@ namespace impl
...
-    //! Returns false if item with the _key did not exist in the cache and a nullptr was inserted for it
-    inline bool getByKeyOrReserve(T** _outval, const K& _key)
+    public:
+        inline typename Base::RangeType findRange(const typename Base::KeyType_impl& _key)
+        {
+            return findRange_internal<typename Base::RangeType>(_key);
+        }
+        inline typename Base::ConstRangeType findRange(const typename Base::KeyType_impl& _key) const
+        {
+            return findRange_internal<typename Base::ConstRangeType>(_key);
```
devshgraphicsprogramming: how does your compiler like the fact that you are calling `findRange_internal`, a function without a `const` qualifier?

Because GCC sure-as-hell won't like it.

devshgraphicsprogramming: although if you made the internal function a `static` and take the container as a parameter then you'd be safe

include/CObjectCache.h
```diff
@@ -70,191 +122,433 @@ namespace impl
...
+        }
+    };
+
+    template<
+        bool IsMultiContainer,
```
devshgraphicsprogramming: again misleading parameter name, either use `isNonVectorBased` or `isAssociativeContainer`

include/CObjectCache.h
```diff
@@ -70,191 +122,433 @@ namespace impl
...
 
-		*_outval = it->second;
-		return true;
-    }
+        bool getKeyRangeOrReserve(typename Base::RangeType* _outrange, const typename Base::KeyType_impl& _key)
+        {
+            *_outrange = findRange(_key);
+            if (!Base::isNonZeroRange(*_outrange))
+            {
+                this->m_container.insert(_outrange->second, {_key, nullptr});
```
devshgraphicsprogramming: i think we need to modify `_outrange->second` so that the final `_outrange` is a valid range and can be used normally

include/CObjectCache.h
```diff
@@ -70,191 +122,433 @@ namespace impl
...
+                    found = it;
+                    break;
+                }
+            }
+            return found;
+        }
+    };
+
+    template<
+        bool isVectorContainer,
```
devshgraphicsprogramming, line 340: now that's a well named template parameter

include/CObjectCache.h
```diff
@@ -70,191 +122,433 @@ namespace impl
...
-            m_container.erase(it);
+            INSERT_IMPL_ASSOC
         }
-    }
 
-private:
-    inline typename ContainerT::const_iterator find(const K& _key) const
+        inline typename Base::RangeType findRange(const typename Base::KeyType_impl& _key)
+        {
+            auto it = this->m_container.lower_bound(_key);
```
devshgraphicsprogramming: why `lower_bound` and not `find` or whatever built-in function it has ?

include/CObjectCache.h
```diff
@@ -70,191 +122,433 @@ namespace impl
...
+        inline RngType findRange_internal(const typename Base::KeyType_impl& _key)
+        {
+            auto cmpf = [](const typename Base::PairType_impl& _a, const typename Base::PairType_impl& _b) -> bool {return _a.first < _b.first; };
+            typename Base::PairType_impl lookingFor{_key, nullptr};
+
+            RngType range;
+            range.first = std::lower_bound(std::begin(Base::m_container), std::end(Base::m_container), lookingFor, cmpf);
+            if (range.first == std::end(Base::m_container) || _key < range.first->first)
+            {
+                range.second = range.first;
```
devshgraphicsprogramming: if you retuned std::end() for first and second then any function under a lock (or single threaded) would not have to use `isNonZeroRange` to determine if valid range was found

include/CObjectCache.h
```diff
@@ -70,191 +122,433 @@ namespace impl
...
+            return true;
         }
 
-		*_outval = it->second;
-		return true;
-    }
+        bool getKeyRangeOrReserve(typename Base::RangeType* _outrange, const typename Base::KeyType_impl& _key)
+        {
+            *_outrange = findRange(_key);
+            if (!Base::isNonZeroRange(*_outrange))
```
devshgraphicsprogramming, line 300: see my general comment about checking ranges

include/CObjectCache.h
```diff
@@ -70,191 +122,433 @@ namespace impl
...
+        using Base = CUniqObjectCacheBase<isVectorContainer, ContainerT_T, T, K...>;
+
+    public:
+        using Base::Base;
+
+        inline void removeObject(const typename Base::ValueType_impl _obj, const typename Base::KeyType_impl& _key)
+        {
+            typename Base::RangeType range = findRange(_key);
+            auto it = range.first;
+            if (Base::isNonZeroRange(range) && it->second == _obj)
```
devshgraphicsprogramming: see my general comment about `isNonZeroRange`

include/CObjectCache.h
```diff
@@ -70,191 +122,433 @@ namespace impl
...
+
+        //! Returns true if had to insert
+        bool swapObjectValue(const typename Base::KeyType_impl& _key, const typename Base::ValueType_PtrToConst_impl _obj, typename Base::ValueType_impl _val)
+        {
+            greet(_val); // grab before drop
+
+            typename Base::RangeType range = findRange(_key);
+            auto it = range.first;
+
+            if (Base::isNonZeroRange(range) && it->second == _obj)
```
devshgraphicsprogramming, line 445: see my general comment about `isNonZeroRange`

devshgraphicsprogramming:
In general using `isNonZeroRange` to check a range's validity is not really that efficient.

As everything except for vector and possibly map/multimap could suffer big performance hits from the `std::distance` call as the "complexity" part of the spec says
http://www.cplusplus.com/reference/iterator/distance/

the complexity for non-random access iterators will be linear in the distance size.

A far better method would be to simply check if `range.first!=range.second`

include/CObjectCache.h
Crisspl: I have, a line below. Readability purposes

include/CObjectCache.h
Crisspl: >a move constructor would require you to zero out or invalidate the input reference.

That's why I use `std::move`. So `std::function` takes care of that

>Also why not copy and move constructors for the cache base itself?

Sorry, I don't understand what you're asking for

include/CObjectCache.h
Crisspl: No, it'd be `T* const` and I want `const T*`

include/CObjectCache.h
Crisspl: See all specializations of `CPreInserionVerifier`. It's not the one that is used in case you're talking about

include/CObjectCache.h
Crisspl: Ok, I'll remove but imo it's exactly the opposite

include/CObjectCache.h
Crisspl: I meant "is multimal or unordered_multimap". isNotVectorBased would be misleading  here because it doesn't tell that only "multi" containers are allowed here (and vector)

include/CObjectCache.h
Crisspl: That's a bug. Compiler didn't compile the const variant of function and so didn't give an error while testing. I'll fix as you proposed with static

include/CObjectCache.h
Crisspl: See my answer to the comment about the same thing

include/CObjectCache.h
Crisspl: I don't understand. Also what means "normally"?

include/CObjectCache.h
Crisspl: will check tomorrow. Now I'm writing from a phone

include/CObjectCache.h
Crisspl: But it wouldn't give any information where to put new element/relation between contents of cache and elent being inserted. `findRange` is used also in functions where insertions happens

Crisspl:
Yea that's wrong. I'll change it to `!=` as you suggest

devshgraphicsprogramming:
I think all uses of `NoPtrValueType_impl` could be replaced with `const
value_type &` or `const_reference_type` then the cache could take
non-ptr types too.

devshgraphicsprogramming:
ok I see that std::move shoudl take care of the first problem

By "copy and move constructors for the cache base itself?" I mean
```
inline explicit CObjectCacheBase(const CObjectCacheBase& other)
inline explicit CObjectCacheBase(CObjectCacheBase&& other)
```

devshgraphicsprogramming:
I guess if you change `isNonZeroRange` to use the != operator instead of
`std::distance` this should be fine to stay.

devshgraphicsprogramming:
Right now you have a situation that if you find a key, then the range
returned will be a valid range of all key values for that key.

However if the key is not present, you insert a nullptr key, and you
still return the old found range.

There are 2 problems with this:
1. Insertion (container modification) can invalidate existing iterators
2. Now the real valid range is the iterator returned by
`m_container.insert` and the `std::next` of that iterator

problem number 1 is the biggest issue as it will cause a segfault if
someone wants to use the returned range to know where their reserved
nullptr is.

devshgraphicsprogramming:
OK, at least write a big comment on that line.

devshgraphicsprogramming:
it is the specialization used for single element associative cache, but
I forgot it returns a std::pair<iterator,bool>

include/CObjectCache.h
devshgraphicsprogramming: Granted, we could do the following as a replacement to `ValueType_PtrToConst_impl`
```
constexpr static isTypeAPointer = std::is_pointer<ValueType_impl>::value;
typedef typename std::conditional<isTypeAPointer,const typename std::remove_pointer<ValueType_impl>::type*,const ValueType_impl&>::type ConstObjectHandle_impl
```

This would leverage SFINAE to handle non-pointer cached object types

P.S. Right now non-ptr objects will fail because they won't be able to instantiate the `NoPtrValueType_impl` and `ValueType_PtrToConst_impl` aliases.

Crisspl:
>By "copy and move constructors for the cache base itself?" I mean
```
inline explicit CObjectCacheBase(const CObjectCacheBase& other)
inline explicit CObjectCacheBase(CObjectCacheBase&& other)
```

They're generated by the compiler for every class ever as long as the're not explicitely deleted in the class or one of its memebers.

Crisspl:
Also please answer in the conversation because it's hard to read when all the talking is scattered all over the page

include/CObjectCache.h
Crisspl: Non-pointers will fail because that's how cache is designed and it's also made sure by `static_assert`. Also I'm not sure if you understand how it is. Cache's template parameter takes any type, it doesn't have to be pointer type! But underlaying container **always** contains pointer type (and so `ValueType_impl` typedef is always pointer type) which is quite obvious since it's a cache and not a regular container and i believe that's why you initially asked for cache as container for pointers.
Example code:
```
core::CConcurrentObjectCache<core::ICPUBuffer, std::string, std::unordered_map> cache;
core::ICPUBuffer* b = new core::ICPUBuffer(0xff);
cache.insert("b", b);
cache.removeObject(b, "b");
```

Crisspl:
>However if the key is not present, you insert a nullptr key, and you
still return the old found range.

Oh yeah, it's a bug, thanks

devshgraphicsprogramming:
> Also please answer in the conversation because it's hard to read when all the talking is scattered all over the page

Reply by email seems not to put my replies in the right thread.

Do you want me to copy the replies in the right places?

devshgraphicsprogramming:
> They're generated by the compiler for every class ever as long as the're not explicitely deleted in the class or one of its memebers.

Ok, but can we get a static_assert that the end classes are copyable and movable ?

Crisspl:
Sure,
Btw, concurrent variants won't be moveable nor copyable (also in the sense of assignment) with c++11 impl. of locks (`std::atomic`).

devshgraphicsprogramming:
I take thats a given about concurrents :)

include/CObjectCache.h
devshgraphicsprogramming: and?

include/CObjectCache.h
devshgraphicsprogramming: ok I guess thats fine with using != for isNonZeroRange

include/CObjectCache.h
devshgraphicsprogramming: damn...

I thought I would instantiate the cache as `core::CConcurrentObjectCache<core::ICPUBuffer*, std::string, std::unordered_map>`


> i believe that's why you initially asked for cache as container for pointers.

Its not required for the cache to work on all types, but I thought if it could be achieved through changing a single line of code, then it should be done.

Crisspl: Well, it already works on all types. That's why we have customizable greet and dispose functions

Crisspl: Also thanks to nature of containing just pointers we don't even care about copyable, moveable etc.

Crisspl:
So i'll explicetely delete copy/move ctors/assignment ops in concurrents so they're uniform regardless of locks implementation, ok?

include/CObjectCache.h
Crisspl: Well, seems like "just because". What's the differenece anyway

include/CObjectCache.h
Crisspl: Also that's why there's two typedefs: `CachedType` (public) and `ValueType_impl` (private)

Crisspl:
>Ok, but can we get a static_assert that the end classes are copyable and movable ?


Hm turns out it's not really possible, because such static_assert must be inside template class definition (so it  triggers at template instantiation) but at least MSVC doesn't allow putting "undefined class" (because it's *inside* the definition) into `is_copy_constructibe<>` etc. Btw it's implemented with some compiler intrinsic here.
Also it can't live outside the definition because it'd have been concretized for some template args. Best bet imo would be to put some instantiations of cache template type into those type traits but it's extremely ugly

include/CObjectCache.h
devshgraphicsprogramming: fair enough I see its too much work to support all types.

devshgraphicsprogramming:
> Best bet imo would be to put some instantiations of cache template type into those type traits but it's extremely ugly

Ok lets not bother with that.

devshgraphicsprogramming:
I closed and merged but some areas I marked with `// can be get some general specialization for any allocator?` will need an update for the future because we need to support all std::vector (map, etc.) regardless of allocator used.

Crisspl:
> will need an update for the future because we need to support all std::vector (map, etc.) regardless of allocator used.

Yea, i've seen your plans about this somewhere. I'll do this in next commit

devshgraphicsprogramming:
You'll need a big merge for that ;)

Crisspl:
You've done it :D and i already pulled to my local

PR #164, opened 2018-11-08
Fix compile errors for OpenCL build

manhnt9:


devshgraphicsprogramming:
This is not a way to fix the issue... I think `core::` is missing from the names of the classes the OpenCL objects inherit from

As well as possibly the `irr/core/BaseClasses.h` include

manhnt9:
I see many other classes have been commented out that same public keyword so I did the similar. Do you mind explaining what AllocationOverrideDefault is and why it's needed for this class?

devshgraphicsprogramming:
AllocationOverrideDefault overrides the default C++ operator new and delete to make sure that when the class is `new`ed the memory its been allocated resides on a address boundary aligned to the default irrlicht alignment (now 16, but soon 32 for AVX).

All classes in irrlicht that do not inherit from a class (like `IReferenceCounted`) that is derived from `AllocationOverrideDefault` should declare inheritance from `AllocationOverrideDefault`, unless:
1. It is a packed class that needs to be packed in arrays more tighty than the default irrlicht alignment would allow (i.e. vectorSIMDf)
2. It requires alignment bigger than the defaut irrlicht alignment (but it should still inherit from `AlignedBase` )
3. Requires a specific alignment that may never change (but should still inherit from `AlignedBase`), an example here are classes for interop with GLSL like matrices, vectors so that structs containing arrays of them have the same layout as the UBO/SSBO etc. in the shader
4. Is supposed to be byte-compatible with some other struct or data declared outside of irrlicht (i.e. a meta-type that you use as an accessor to plain void* memory)

devshgraphicsprogramming:
> I see many other classes have been commented out that same public keyword so I did the similar.

These were the classes I could not decide on how to handle their alignment (which category they belong to)

manhnt9:
Pushed the new fix.

devshgraphicsprogramming:
thanks a lot :D

Issue #162, opened 2018-10-30
Provide full GPU API, Device and Window Format enumeration [kill IVideoModeList]


devshgraphicsprogramming:
Before `createDevice` we should have a  `queryGraphicsAPIsAndDevices(void*)` and `queryCL(const IVideoDeviceCapReport*)` functions which would return a list of `IVideoAPI`s which mainly contain a list of constant immutable `IVideoDeviceCapReport` (renamed `IVideoCapabilityReporter` , which would no longer be the base of `IDriver`) 

This would allow us to know if the engine was compiled with OpenGL+Vulkan support, and if Vulkan was detected on the system (#79 ).

Each API could support different resolutions, window configs etc. so its important that the `SWindowVideoMode`s (improved and renamed IVideoModeList) are per-API-per-device, especially with multi-GPU.

`SWindowVideoMode` should consist of:
- `SFramebufferConfig` which provides the resolution, `E_FORMAT`s for color and depth/stencil and sample counts
- Refresh Rate
- Double Buffering 

`queryCL` function would return a vector of `IOpenCLCapReport` for any platform and device combination present on the system which can interop with the argument provided API+Device.

`IVideoDeviceCapReport` would be a constant immutable object with only constant methods, and it would provide us with:
- Vendor ID+Name
- Numeric API version (same version macro as VK, 10 bit major, 11bit minor and patch)
- The Windowed Framebuffer Configurations that the device supports in the form of vector of `SWindowVideoMode`s (empty vector implies offscreen rendering only)
- Supported Queue Families and their caps (compute, graphics, transfer, etc.)
- Graphics Features (EDF_ enum + more) as determined from VK/GL extensions and VK Features
- GLSL Extensions as determined from VK/GL extensions
- Everything that current `IVideoCapabilityReporter` provides
- Required alignments for offsets and bindings
- Max UBO, SSBO, Texture Type sizes etc.
- Other Device/API limits (Compute work group sizes, MSAA samples, max FBO attachments, max descriptor bindings, etc.)
- Supported E_FORMATs for rendering, sampled images, r/w images, vertex source data, buffer data, etc.
- Memory Heaps and their properties (has to be determined through sampling on OpenGL)
- Static Constexpr for certain capabilities which we fix to the minimum  (i.e. push constans size)

[...]

PR #165, opened 2018-11-12
Add OpenCL context

manhnt9:
Add OpenCL context creation to function `getCLDeviceFromGLContext` because:
- Reuse cl_context_properties and cl_device_id

Thus it's not only `getCLDeviceFromGLContext` anymore => function name changed to `initOpenCL`.


source/Irrlicht/COpenCLHandler.cpp
```diff
@@ -14,6 +14,7 @@ clGetGLContextInfoKHR_fn COpenCLHandler::pClGetGLContextInfoKHR = NULL;
 cl_platform_id COpenCLHandler::platforms[IRR_MAX_OCL_PLATFORMS];
 cl_uint COpenCLHandler::actualPlatformCount = 0;
 COpenCLHandler::SOpenCLPlatformInfo COpenCLHandler::platformInformation[IRR_MAX_OCL_PLATFORMS];
+cl_context COpenCLHandler::openCLContext = 0;
```
devshgraphicsprogramming, line 17: this shoud not be a static/singleton multiple contexts should be possible

source/Irrlicht/COpenCLHandler.h
```diff
@@ -211,8 +211,10 @@ class COpenCLHandler
 
         static const SOpenCLPlatformInfo& getPlatformInfo(const size_t& ix) {return platformInformation[ix];}
 
+        static cl_context getOpenCLContext() {return openCLContext;}
+
 #ifdef _IRR_COMPILE_WITH_OPENGL_
-        static bool getCLDeviceFromGLContext(cl_device_id& outDevice,
```
devshgraphicsprogramming, line 215: we need both functions, for example when the OpenCL device has already been created externally and we need to see if its the right one for sharing with the OpenGL GPU.

source/Irrlicht/COpenGLDriver.cpp
```diff
@@ -445,7 +445,7 @@ bool COpenGLDriver::initDriver(CIrrDeviceWin32* device)
 	}
 
 #ifdef _IRR_COMPILE_WITH_OPENCL_
-	ocl::COpenCLHandler::getCLDeviceFromGLContext(clDevice,hrc,HDc);
+	ocl::COpenCLHandler::initOpenCL(clDevice,hrc,HDc);
```
devshgraphicsprogramming, line 448: do not init an OpenCL device context by default for everyone, only if asked for in device creation parameters.

While I'm not asking for a full #162 fix and implementation, that API calls for a query OpenCL devices and a list of devices for which to create contexts in device creation paramters...

Basically a simple `createOpenCLInteropContext` boolean in device creation parameters will do for now.

source/Irrlicht/COpenCLHandler.cpp
manhnt9: whole COpenCLHandler is current used as static, so everything is static
I will store a COpenCLHandler object in COpenGLDriver, is that OK?

devshgraphicsprogramming:
@manhnt9 any progress on this?

manhnt9:
@devshgraphicsprogramming not yet, I'm developing Saga3D most of the time.

devshgraphicsprogramming:
doing it by myself.

PR #177, opened 2018-11-16
Fixes for MSVC

Crisspl:
* inline vars is cpp17 feature
* variable size array
* put protected inheritance instead of private in HeterogenousMemoryAddressAllocatorAdaptor

### DIFF
--- a/include/irr/core/alloc/address_allocator_traits.h
+++ b/include/irr/core/alloc/address_allocator_traits.h
@@ -156,9 +156,9 @@ namespace core
 
             template<class U> struct has_func_get_real_addr<U,void_t<func_get_real_addr<U> > >  : std::is_same<func_get_real_addr<U>,size_type> {};
 
-            static inline constexpr bool         supportsArbitraryOrderFrees = resolve_supportsArbitraryOrderFrees<AddressAlloc>::value;
-            static inline constexpr uint32_t maxMultiOps                           = resolve_maxMultiOps<AddressAlloc>::value;
-            static inline constexpr bool         supportsNullBuffer                 = resolve_supportsNullBuffer<AddressAlloc>::value;
+            static constexpr bool         supportsArbitraryOrderFrees = resolve_supportsArbitraryOrderFrees<AddressAlloc>::value;
+            static constexpr uint32_t maxMultiOps                           = resolve_maxMultiOps<AddressAlloc>::value;
+            static constexpr bool         supportsNullBuffer                 = resolve_supportsNullBuffer<AddressAlloc>::value;
 
             static inline void          printDebugInfo()
             {
--- a/include/irr/core/alloc/address_allocator_traits.h
+++ b/include/irr/core/alloc/address_allocator_traits.h
@@ -156,9 +156,9 @@ namespace core
 
             template<class U> struct has_func_get_real_addr<U,void_t<func_get_real_addr<U> > >  : std::is_same<func_get_real_addr<U>,size_type> {};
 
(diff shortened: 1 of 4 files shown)


include/irr/core/alloc/address_allocator_traits.h
```diff
@@ -156,9 +156,9 @@ namespace core
 
             template<class U> struct has_func_get_real_addr<U,void_t<func_get_real_addr<U> > >  : std::is_same<func_get_real_addr<U>,size_type> {};
 
-            static inline constexpr bool         supportsArbitraryOrderFrees = resolve_supportsArbitraryOrderFrees<AddressAlloc>::value;
-            static inline constexpr uint32_t maxMultiOps                           = resolve_maxMultiOps<AddressAlloc>::value;
-            static inline constexpr bool         supportsNullBuffer                 = resolve_supportsNullBuffer<AddressAlloc>::value;
+            static constexpr bool         supportsArbitraryOrderFrees = resolve_supportsArbitraryOrderFrees<AddressAlloc>::value;
```
devshgraphicsprogramming: are you sure MSVC is complaining here?

Because I will get linking errors in GCC if the `constexpr` is not `static inline` and I sure as hell compile with `-std=c++11`

source/Irrlicht/CMeshSceneNodeInstanced.cpp
```diff
@@ -222,8 +222,10 @@ uint32_t CMeshSceneNodeInstanced::addInstance(const core::matrix4x3& relativeTra
 
 bool CMeshSceneNodeInstanced::addInstances(uint32_t* instanceIDs, const size_t& instanceCount, const core::matrix4x3* relativeTransforms, const void* extraData)
 {
-    uint32_t dummyBytes[instanceCount];
-    uint32_t aligns[instanceCount];
+    core::vector<uint32_t> dummyBytes_(instanceCount);
```
devshgraphicsprogramming, line 226: good fix, however try to scope the vectors so they go out of scope after the `multi_alloc_addr`

Crisspl: wanted to make it same scope as those arrays. Anyway `dummyBytes` is used once more after `multi_alloc_addr` (line 246)

devshgraphicsprogramming: ok as you wish.

include/irr/core/alloc/address_allocator_traits.h
```diff
@@ -156,9 +156,9 @@ namespace core
 
             template<class U> struct has_func_get_real_addr<U,void_t<func_get_real_addr<U> > >  : std::is_same<func_get_real_addr<U>,size_type> {};
 
-            static constexpr bool         supportsArbitraryOrderFrees = resolve_supportsArbitraryOrderFrees<AddressAlloc>::value;
-            static constexpr uint32_t maxMultiOps                           = resolve_maxMultiOps<AddressAlloc>::value;
-            static constexpr bool         supportsNullBuffer                 = resolve_supportsNullBuffer<AddressAlloc>::value;
+            static _IRR_STATIC_INLINE_CONSTEXPR constexpr bool         supportsArbitraryOrderFrees = resolve_supportsArbitraryOrderFrees<AddressAlloc>::value;
```
devshgraphicsprogramming: _IRR_STATIC_INLINE_CONSTEXPR should replace the whole `static` and `contexpr`

devshgraphicsprogramming: I see you defined the macro correctly, so just take the extra qualifiers away

devshgraphicsprogramming:
Can you commit the CMakeLists fixups too?

PR #181, opened 2018-11-17
High performance Draw3DLine function

manhnt9:
It speeds up my ray debug rendering so much so I can't help firing a pull request.

### DIFF
--- a/include/CFinalBoneHierarchy.h
+++ b/include/CFinalBoneHierarchy.h
@@ -116,9 +116,9 @@ namespace scene
 					_interpAnimsBegin > _interpAnimsEnd ||
 					_nonInterpAnimsBegin > _nonInterpAnimsEnd
 				)
-				_IRR_DEBUG_BREAK_IF(_boneNamesEnd - _boneNamesBegin != boneCount)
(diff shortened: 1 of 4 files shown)


ext/DebugDraw/CDraw3DLine.cpp
```diff
@@ -71,6 +72,32 @@ void CDraw3DLine::draw(
     fence->drop();
 }
 
+void CDraw3DLine::draw(const core::vector<S3DLineVertex>& linesData)
```
devshgraphicsprogramming: should that be a vector of std::pair ?

devshgraphicsprogramming: since you will always require an even number of vertices, make that a vector of vertex-pairs

manhnt9: I will update soon

manhnt9:
Well something made this PR automatically closed ...

manhnt9:
Ah I lost my changes, maybe I reset my branch without caring about this PR. Anyways, the code is still in my engine so will make a new PR before new week starts.

PR #186, opened 2018-11-21
Update CMake: find and enable vulkan

manhnt9:
Verified with generated makefile, have definition and library (add, not overwrite).
Not sure why `target_link_libraries(Irrlicht INTERFACE Vulkan::Vulkan)` (new CMake syntax) doesn't work though, maybe because of the CMake version, so using traditional command.

### DIFF
--- a/source/Irrlicht/CMakeLists.txt
+++ b/source/Irrlicht/CMakeLists.txt
@@ -384,11 +384,14 @@ endif()
 find_package(Vulkan)
 
 if (Vulkan_FOUND)
+	message(STATUS "Building with Vulkan driver")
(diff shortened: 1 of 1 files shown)


devshgraphicsprogramming:
can it echo whether it found vulkan or not?

manhnt9:
CMake will print out if it found Vulkan and the library's location.
Any custom message can be printed out using CMake command `message`.

devshgraphicsprogramming:
Can you message "BUILDING with Vulkan":"NOT BUILDING with Vulkan"

devshgraphicsprogramming:
great work, thanks!

PR #187, opened 2018-11-22
CMake: add missing Vulkan include dir

manhnt9:


manhnt9:
Well, I think I must have replaced all my CMake knowledge with 3.12, so I made a mistake.
In modern CMake, we only need to `link` a library, `include dir` of a library is a `property` of that library, so only need to link and CMake will add `include dir` automatically.
This PR is for old CMake, everything is variable so we need to specify.

In new CMake, we have target and target can have multiple properties, that's more convenient.

devshgraphicsprogramming:
does this work with the new LunarG SDK that comes as a package from a repository ? (no more need for that pesky `VULKAN_SDK` environment variable)

manhnt9:
which package? I have libvulkan-dev installed and I don't have to set any environment variable.

devshgraphicsprogramming:
That's exactly what I wanted to hear.

PR #190, opened 2018-11-23
MSVC not inheriting default ctor (?) fix

Crisspl:


devshgraphicsprogramming:
need default ctors for vectorSIMD_ and vectorSIMDBool as well

PR #199, opened 2018-12-22
Fix vectorSIMD int and bool

devshgraphicsprogramming:


devshgraphicsprogramming:
I already have all these fixes @Crisspl, delete your branch.

PR #213, opened 2019-01-23
Asset Pipeline Merge

devshgraphicsprogramming:
Can't merge automatically yet, three macro-scale issues to resolve:
- Merge master into the branch first
- Address the re-raised issues and comments from the last few days
- Update the examples to compile and run under the new APIs


include/irr/asset/ICPUTexture.h
```diff
@@ -187,16 +175,7 @@ class ICPUTexture : public IAsset
...
-        for (uint32_t i = 1u; i < m_mipmaps.size(); ++i)
-        {
-            if (fmt != m_mipmaps[i]->getColorFormat())
-            {
-                fmt = EF_UNKNOWN;
-                break;
-            }
-        }
-        m_colorFormat = fmt;
+        m_colorFormat = (*std::find_if(m_mipmaps.begin(), m_mipmaps.end(), [](CImageData* mip) { return mip->getColorFormat() != EF_UNKNOWN; }))->getColorFormat();
```
devshgraphicsprogramming, line 183: what about the check that all the mips that don't have format `EF_UNKNOWN`, must have the same format?

devshgraphicsprogramming:
A conflict on CMakeLists.txt and CZipReader.cpp preventing mergeability

include/irr/asset/ICPUTexture.h
Crisspl: Since constructor (and those init functions as well) is protected, construction code assumes that `create()` already make sure that input `core::vector<asset::CImageData*>` is perfectly valid

devshgraphicsprogramming: ok so this sort of validation takes place somewhere else?

Crisspl: take a look at `create_impl()` and `validateMipchain()`

include/irr/video/SubAllocatedDataBuffer.h
```diff
@@ -153,7 +153,7 @@ class SubAllocatedDataBuffer : public virtual core::IReferenceCounted
             auto allocation = mAllocator.getCurrentBufferAllocation();
 
             IGPUBuffer* retval;
-            static_if<is_std_get_0_defined<decltype(allocation)>::value >([&](auto f){
+            static_if<true/*is_std_get_0_defined<decltype(allocation)>::value*/ >([&](auto f){ // <-- todo, apparently MSVC SFINAE bug, however i can't even reproduce this in simpler form
```
devshgraphicsprogramming: that will break, badly I have need for both in the current codebase

Also what MSVC version are you using? this compiles for us on windows.

Crisspl: VS 2017 (15.9.3)
i'll try changing toolset version.
Also i'll have to update cmake because it generates projects for VS 2015 and maybe there's the differenece between our setup

Crisspl: yea, that was it! With v141 toolset it builds just fine

src/irr/asset/CBAWMeshFileLoader.h
```diff
@@ -216,7 +216,7 @@ class CBAWMeshFileLoader : public asset::IAssetLoader
     // Compatibility functions:
 
     template<uint64_t IntoVersion>
-    std::enable_if_t<IntoVersion, bool> formatConversionProlog(
+    std::enable_if_t<IntoVersion/*cannot convert into version 0*/, bool> formatConversionProlog(
```
devshgraphicsprogramming, line 220: isn't `std::enable_if_t`c++17?

Crisspl: Nah, all this `_v` and `_t` from `<type_traits>` is c++14. And i'm always checking the standard version of a feature if i'm not sure, so don't worry :)

devshgraphicsprogramming:
@Crisspl Is this Merge Request ready?

I've done a number of stability/bugfix commits to master, if you can merge with them first before I press the holy green button I'd be much obliged.

PR #218, opened 2019-01-29
Remove deleted example from CMake build

manhnt9:
My machine fails to run CMake because of it.


devshgraphicsprogramming:
hmm... this reminds me that I havent pushed latest  BaW backports

devshgraphicsprogramming:
Had to fix it on my own.

https://github.com/buildaworldnet/IrrlichtBAW/commit/569a3f02c8141005eb0c834ac898c04d80e11b16

PR #219, opened 2019-02-02
Bullet Physics Extension

devshgraphicsprogramming:


devshgraphicsprogramming:
@khom-khun I'd remove `CCollisionShape` and inline pretty much every class method that isn't `virtual`

examples_tests/17.SimpleBulletIntegration/CMakeLists.txt
```diff
@@ -4,4 +4,12 @@ if(NOT RES)
 	message(FATAL_ERROR "common.cmake not found. Should be in {repo_root}/cmake directory")
 endif()
 
-irr_create_executable_project("" "")
\ No newline at end of file
+set(BULLET_COLLISION_SRC "" CACHE STRING "For example 17.SimpleBulletIntegration - Not required")
```
devshgraphicsprogramming: what happens if can't find Bullet on system? Can it build without?

(BTW there is some code in the top-level CMake that skips building with Vulkan if Vulkan SDK is not detected on the system, maybe you can steal that approach)

khomsidakhun: I'd assume it won't spew errors and you'd be able to do without but just to be safe I'll add some sanity checks if they leave it blank and not glob_recurse over a blank string

devshgraphicsprogramming:
Its a nice PR, it be a shame if it never got merged.

examples_tests/17.SimpleBulletIntegration/CMakeLists.txt
devshgraphicsprogramming: now your Bullet lib needs to come from the submodule

devshgraphicsprogramming:
Now basically add an option to cmake `IRR_BUILD_BULLET` (just like `IRR_BUILD_CEGUI`) that controls the building of your bullet example and the bullet dependency

Maybe merge the @manhnt9 sRGB branch into yours to access that nice `irr::ext` library building macro?

devshgraphicsprogramming:
So I guess we're ready to merge?

PR #222, opened 2019-02-19
CCustomAllocatorCPUBuffer

Crisspl:


include/irr/asset/ICPUBuffer.h
```diff
@@ -97,21 +97,26 @@ class ICPUBuffer : public core::IBuffer, public asset::IAsset
...
-            m_allocator.deallocate(reinterpret_cast<typename Allocator::pointer>(ICPUBuffer::data), ICPUBuffer::size/sizeof(typename Allocator::value_type));
-        ICPUBuffer::data = nullptr; // so that ICPUBuffer will not try deallocating
+            m_allocator.deallocate(reinterpret_cast<typename Allocator::pointer>(ICPUBuffer::data), ICPUBuffer::size);
+        ICPUBuffer::data = nullptr;
     }
 
 public:
-    IMemoryAdoptingCPUBuffer(size_t sizeInBytes, void* dat, Allocator&& allctr = Allocator()) : ICPUBuffer(sizeInBytes, dat), m_allocator(std::move(allctr))
-    {}
+    CCustomAllocatorCPUBuffer(size_t sizeInBytes, void* dat, Allocator&& alctr = Allocator()) : CCustomAllocatorCPUBuffer(sizeInBytes, alctr.allocate(allocSz), core::adopt_memory, std::move(alctr))
```
devshgraphicsprogramming: why is this constructor implicitly adopting memory?

Crisspl: just using the adopting constructor but as you can see it allocates its own mem. However it's missing memcpy

devshgraphicsprogramming:
@Crisspl  Briliant, but we're missing a null-allocator specialization (because null allocator will allocate nullptr and default custom allocator CPUBuffer will try to memcpy into that), that deletes the non-adopting constructor.

PR #250, opened 2019-04-01
SRGB 

qbasa12:
uncommented these lines, changed enum in jpg added some asset data types in copengltexture


src/irr/asset/CImageLoaderPNG.cpp
```diff
@@ -189,10 +193,19 @@ asset::IAsset* CImageLoaderPng::loadAsset(io::IReadFile* _file, const asset::IAs
...
+		if (png_get_gAMA(png_ptr, info_ptr, &image_gamma)) {
+
+		
 			png_set_gamma(png_ptr, screen_gamma, image_gamma);
-		else
+		}
+
+		else {
+			
+			os::Printer::log(std::to_wstring(image_gamma), ELL_DEBUG);
```
devshgraphicsprogramming, line 205: nice but add a human readabke string beforehand, because someone will just get a random number out of nowhere in the console and not know what it means

src/irr/asset/CImageLoaderPNG.cpp
```diff
@@ -189,10 +193,19 @@ asset::IAsset* CImageLoaderPng::loadAsset(io::IReadFile* _file, const asset::IAs
 	else
```
devshgraphicsprogramming, line 193: `screen_gamma` is actually 2.233333333

devshgraphicsprogramming: Or try to set to 1.0 and then use the SRGB formats instead in `new asset::CImageData` later on

devshgraphicsprogramming (changes requested):

source/Irrlicht/COpenGLDriver.cpp
```diff
@@ -1,3324 +1,3324 @@
...
-    if (correctOffsetsForXFormDraw)
-    {
-        size_t offsets[asset::EVAI_COUNT] = {0};
-        memcpy(offsets,&spec->getMappedBufferOffset(asset::EVAI_ATTR0),sizeof(offsets));
-        for (size_t i=0; i<asset::EVAI_COUNT; i++)
-        {
-            if (!spec->getMappedBuffer((asset::E_VERTEX_ATTRIBUTE_ID)i))
-                continue;
-
-            if (spec->getAttribDivi
```
devshgraphicsprogramming: I said to assume `Params.HandleSRGB=true` and remove it from the engine!

devshgraphicsprogramming (changes requested):
HandleSRGB should be gone and assumed always true

devshgraphicsprogramming:
TGA and BMP loaders are untouched.

devshgraphicsprogramming:
@Crisspl thanks a lot for the comments, you've saved me a great deal of time
@qbasa12 for the first time in the history of our company the commit log of a branch is so bad it needs to be either rewritten or squashed/rebased because it makes no sense.

COpenGLDriver.cpp is a mess in the diff viewer (https://github.com/buildaworldnet/IrrlichtBAW/pull/250/files ) because most probably the newlines got messed up.

source/Irrlicht/COpenGLDriver.cpp
qbasa12: allright

src/irr/asset/CImageLoaderPNG.cpp
qbasa12: actually always:1.0

source/Irrlicht/COpenGLTexture.cpp
```diff
@@ -75,8 +75,8 @@ void COpenGLFilterableTexture::regenerateMipMapLevels()
 		return;
 
     COpenGLExtensionHandler::extGlGenerateTextureMipmap(TextureName,this->getOpenGLTextureType());
-}
-
+}
+
 
 uint32_t COpenGLTexture::getOpenGLFormatBpp(const GLenum& colorformat)
```
devshgraphicsprogramming, line 81: @qbasa12 this one needs switch cases for SRGB as well

P.S. @Crisspl why are there EF_R8_SRGB, EF_R8G8_SRGB.... I'm 100% sure that OpenGL doesn't support that, is that even a thing in Vulkan?

source/Irrlicht/COpenGLTexture.cpp
```diff
@@ -341,6 +341,12 @@ void COpenGLTexture::getOpenGLFormatAndParametersFromColorFormat(const asset::E_
 			colorformat = GL_RG;
 			type = GL_FLOAT;
 		}
+			case asset::EF_R8G8B8A8_SRGB:
```
devshgraphicsprogramming, line 344: @qbasa12 there are more SRGB formats https://github.com/Crisspl/IrrlichtBAW/blob/master/include/irr/asset/EFormat.h, please provide cases for all.l

P.S. @Crisspl why are there EF_R8_SRGB, EF_R8G8_SRGB.... I'm 100% sure that OpenGL doesn't support that, is that even a thing in Vulkan?

include/SIrrCreationParameters.h
```diff
@@ -133,7 +133,8 @@ namespace irr
 		(which are usually in sRGB space) are correctly displayed. However vertex colors
 		and other explicitly set values have to be manually encoded in linear color space.
 		Default value: false. */
-		bool HandleSRGB;
+		//
```
devshgraphicsprogramming, line 136: excise this and the comment completely

source/Irrlicht/COpenGLDriver.cpp
```diff
@@ -295,7 +295,7 @@ bool COpenGLDriver::initDriver(CIrrDeviceWin32* device)
 			WGL_DOUBLE_BUFFER_ARB,Params.Doublebuffer ? 1 : 0,
 			WGL_STEREO_ARB,Params.Stereobuffer ? 1 : 0,
 			WGL_PIXEL_TYPE_ARB, WGL_TYPE_RGBA_ARB,
-			WGL_FRAMEBUFFER_SRGB_CAPABLE_ARB, Params.HandleSRGB ? 1:0,
+			WGL_FRAMEBUFFER_SRGB_CAPABLE_ARB, Params.HandleSRGB=true,
```
devshgraphicsprogramming, line 298: replace `Params.HandleSRGB` with `true` entirely

source/Irrlicht/COpenGLTexture.cpp
Crisspl: >there are more SRGB formats

The rest is taken care of or not supported in GL

>why are there EF_R8_SRGB, EF_R8G8_SRGB.... I'm 100% sure that OpenGL doesn't support that, is that even a thing in Vulkan?

All we have is slightly renamed VkFormat. Ofc it doesn't mean they are supported in GL

devshgraphicsprogramming: I'm surprised those things are supported in vulkan XD.

However those switch statements are still missing the cases for `EF_R8G8B8_SRGB`, `EF_B8G8R8_SRGB`, the BCn SRGB formats, ETC and ASTC

Crisspl: ETC and ASTC SRGB are all there, but if I haven't overlooked a few of BCn are missing (BC1-BC3 I think) so please add also them, thanks

devshgraphicsprogramming: @qbasa12 any questions?

PR #251, opened 2019-04-04
SRGBLOADER 

qbasa12:


src/irr/asset/CImageLoaderBMP.cpp
```diff
@@ -202,42 +202,43 @@ asset::IAsset* CImageLoaderBMP::loadAsset(io::IReadFile* _file, const asset::IAs
 
 	_file->read(&header, sizeof(header));
 	
-	byte BitmapHeader[header.BitmapHeaderSize];
+	std::vector<uint8_t> BitmapHeader;
```
Crisspl, line 205: use `core::vector` instead (our typedef with out custom allocator as default, defined in include/irr/core/Types.h)

src/irr/asset/CImageLoaderBMP.cpp
```diff
@@ -202,42 +202,43 @@ asset::IAsset* CImageLoaderBMP::loadAsset(io::IReadFile* _file, const asset::IAs
...
-				os::Printer::log(std::to_wstring( Extra[i]), ELL_DEBUG);
-			#endif
+#ifdef _DEBUG
+		os::Printer::log(std::to_wstring(Extra[i]), ELL_DEBUG);
+#endif
 	};
- 	byte ICC_size =  BitmaphHeader[(header.BitmapHeaderSize-2)];	
-	byte ICC[ICC_size];
-	i = 0;
+	uint8_t ICC_size = BitmapHeader.at((header.BitmapHeaderSize - 2));
```
Crisspl, line 224: Use `operator[]` instead of `at()` - faster. at() checks if you're not going out of range and throws if so

PR #260, opened 2019-04-12
cmake, SRGB formats

Crisspl:


CMakeLists.txt
```diff
@@ -45,6 +48,16 @@ macro(irr_adjust_flags)
 			#"$<$<CONFIG:RELEASE>:-O3>" # added by default and apparently cmake is not clever enough to filter duplicates for this one
 			"$<$<CONFIG:RELEASE>:-fexpensive-optimizations>"
 		)
+		if (IRR_GCC_SANITIZE_ADDRESS)
+			add_compile_options(-fsanitize=address)
```
devshgraphicsprogramming, line 52: still need those added as linker flags as well as compiler flags

Issue #253, opened 2019-04-06
Implement a DDS and KTX loader via GLI


devshgraphicsprogramming:
KTX is Khronos' replacement to DDS

Very good format, maps 1:1 to GL and VK texture upload commands.

This guy has a semi-working KTX loader based off libktx
https://github.com/Didgy74/DEngine
and has given permission to take stuff for the KTX loader from him

devshgraphicsprogramming:
If this supports KTX2 then it would be great to integrate
http://gli.g-truc.net/0.8.2/api/index.html

devshgraphicsprogramming:
Notes for DDS
https://github.com/Microsoft/DirectXTex/tree/master/DDSTextureLoader

devshgraphicsprogramming:
Need confirmation if KTX2 is a thing yet:
https://github.com/KhronosGroup/KTX-Specification/issues/82

If its not yet final, then lets go with GLI

devshgraphicsprogramming:
KTX2 is not final yet

devshgraphicsprogramming:
@Florastamine, take this over with a separate PR

devshgraphicsprogramming:
@Florastamine refer to #255 to make sure that your implementation covers that case as well.

devshgraphicsprogramming:
@achalpandeyy closing, but you might want to read.

Issue #234, opened 2019-02-23
Implement Gamma Correct Rendering


devshgraphicsprogramming:
`HandleSRGB` from creation parameters should disappear (assume always true).
https://github.com/buildaworldnet/IrrlichtBAW/pull/252#commitcomment-32961321

This should be uncommented in COpenGLDriver.cpp
```c++
	if (Params.HandleSRGB)
		glEnable(GL_FRAMEBUFFER_SRGB);
```

And finally all texture loaders should be validated to load all formats correctly (need a unit-test/example for that).

devshgraphicsprogramming:
This is a good occasion to tackle other Texture-API-related issues

devshgraphicsprogramming:
Only a few lines of code need to be changed in CImageLoader* for different LDR formats
https://github.com/buildaworldnet/IrrlichtBAW/tree/master/src/irr/asset

But needs thorough testing with real images (load image, make it into texture, copy/blit it onto the screen and compare with Windows/Linux Photo viewer).

devshgraphicsprogramming:
Maybe https://github.com/buildaworldnet/IrrlichtBAW/blob/master/src/irr/asset/CImageLoaderRGB.cpp image loader can be removed (killed)

devshgraphicsprogramming:
Image writers also need to be adjusted to correctly save textures (respect SRGB, and for non SRGB do colorformat conversion, **for which we already have functions** ).

Crisspl:
Allow only RGB color space and report error on others (return nullptr)

devshgraphicsprogramming:
linear RGB and sRGB

devshgraphicsprogramming:
A basic TODO list:
- [x] Remove `HandleSRGB` window creation parameter (assume always true)
- [x] Remove the RGB image loader
- [x] Provide extra GLenum conversions https://github.com/buildaworldnet/IrrlichtBAW/pull/250#discussion_r271620010
- [x] Fix up the Image Loaders
- [x] Create a Unit Test with multiple images being tested 
- [x] Display Image on screen with full-screen triangle (blit not allowed)
- [x] Fix up the Image Writers
- [x] Create a Screenshot of the image on the screen and save to file
- [ ] Provide a static SColorf

List of Texture Formats to Test
- [x] JPEG color
- [x] JPEG grayscale
- [x] PNG 8bit grayscale
- [x] PNG 24bit rgb
- [ ] PNG 32bit rgba
[...]

PR #273, opened 2019-04-26
WIP: sRGB Loader Test and Fixes for multiple image formats

ghost:
General:
- [x] ~Replace legacy color converters (JPG/PNG/TGA)~
- [x] ~Handle incorrect flipping of image buffers (JPG/PNG/TGA)~
- [x] ~Handle unsupported color formats during loading/saving (JPG/PNG/TGA)~
- [x] ~More test images (preferably PD/commonly used images - Lenny, jelly beans, parrot, etc.)~
- [x] ~Convert formats for display testing as needed~ (Superseded in #305)
- [x] ~Fix Linux-related build errors~

**JPG**
- [x] ~Handle grayscale (`JCS_GRAYSCALE` would load the image as `EF_R8_SRGB`)~

**PNG**
- [x] ~Fix 8-bit grayscale (8-bit got expanded to RGB(A)) & 32-bit RGBA~

**TGA**
- [x] ~Fix 8-bit grayscale & 32-bit RGBA (+RLE)~
- [x] ~Correct handling of TGA headers parsing~


devshgraphicsprogramming:
I thought I'd change the title and let you keep working, no need for PRs of a single commit

src/irr/asset/CImageWriterJPG.cpp
```diff
@@ -105,6 +105,7 @@ static bool writeJPEGFile(io::IWriteFile* file, const asset::CImageData* image,
 	switch( image->getColorFormat () )
 	{
         case asset::EF_R8G8B8_UNORM:
+		case asset::EF_R8G8B8_SRGB:
 			format = video::CColorConverter::convert_R8G8B8toR8G8B8;
```
devshgraphicsprogramming: We want to get rid of CColorConverter, it is legacy code from old irrlicht and doesn't have good format coverage/support.

Maybe you can use the `decode` and `encode` functions that @Crisspl  has made.

JPEG only supports SRGB without conversion (natively), if you encounter `EF_R8G8B8_UNORM` or anything similar then you need a convert to `EF_R8G8B8_SRB` from the input format.

However you have my approval to error-out in such case.

src/irr/asset/CImageWriterJPG.cpp
```diff
@@ -102,6 +102,8 @@ static void jpeg_file_dest(j_compress_ptr cinfo, io::IWriteFile* file)
 static bool writeJPEGFile(io::IWriteFile* file, const asset::CImageData* image, uint32_t quality)
 {
 	void (*format)(const void*, int32_t, void*) = 0;
+	bool grayscale = false;
+	
 	switch( image->getColorFormat () )
 	{
         case asset::EF_R8G8B8_UNORM:
```
devshgraphicsprogramming: none of the below code with CColorConverter will work, that conversion code is legacy

ghost:
@devshgraphicsprogramming I've pushed changes to use the newer color converters for the JPG writer, please review if everything looks correct.

ghost:
The problem now is saved images got its color reduced drastically... which didn't happen with the original converters. Am I doing something wrong? @devshgraphicsprogramming @Crisspl 

image

Crisspl:
Make sure that used convert functions (which are based on decode and encode functions) work as they're supposed to. I've just taken a look at R8G8B8_SRGB decode/encode and seems ok to me but maybe you can make sure in practice

ghost:
BMP would need a rewrite in order to properly implement ICC detection & to fix a bunch of other issues, right now it's severly broken and would refuse to load anything not coming out from MSPaint

devshgraphicsprogramming:
> 
> 
> BMP would need a rewrite in order to properly implement ICC detection & to fix a bunch of other issues, right now it's severly broken and would refuse to load anything not coming out from MSPaint

You have my blessing to nuke BMP from the engine (writer and loader) there's nothing that BMP can do that PNG can't.

Time will be better spent on KTX and DDS in the future rather than a rewrite of BMP.

src/irr/asset/CImageWriterJPG.cpp
```diff
@@ -209,7 +209,7 @@ bool CImageWriterJPG::writeAsset(io::IWriteFile* _file, const SAssetWriteParams&
     io::IWriteFile* file = _override->getOutputFile(_file, ctx, {image, 0u});
     const asset::E_WRITER_FLAGS flags = _override->getAssetWritingFlags(ctx, image, 0u);
     const float comprLvl = _override->getAssetCompressionLevel(ctx, image, 0u);
-	return writeJPEGFile(file, image, (!!(flags & asset::EWF_COMPRESSED))*(1.f-comprLvl)*100.f); // if quality==0, then it defaults to 75
+	return writeJPEGFile(file, image, (!!(flags & asset::EWF_COMPRESSED)) * (uint32_t)((1.f-comprLvl)*100.f)); // if quality==0, then it defaults to 75
```
devshgraphicsprogramming: don't use C-style cast, use `static_cast<uint32_t>`

src/irr/asset/CImageLoaderTGA.cpp
```diff
@@ -84,6 +84,29 @@ bool CImageLoaderTGA::isALoadableFileFormat(io::IReadFile* _file) const
 	return true;
 }
 
+// convertColorFlip() does color conversion as well as taking care of properly flipping the given image.
+template <typename T, E_FORMAT srcFormat, E_FORMAT destFormat>
+static void convertColorFlip(const T *src, T *dest, core::vector3d<uint32_t> &size, int c, bool flip)
```
devshgraphicsprogramming: this is the kind of thing that would be **very** useful in top level header, maybe in `./include/irr/video/convertColor.h`

src/irr/asset/CImageLoaderTGA.cpp
```diff
@@ -84,6 +84,29 @@ bool CImageLoaderTGA::isALoadableFileFormat(io::IReadFile* _file) const
...
+{
+	T *in = (T *) src;
+	T *out = (T *) dest;
+	uint32_t stride = size.X * c;
```
devshgraphicsprogramming: your strides will be different for non-identical src and dest Formats

moreover images usually have an issue with row-pitch alignment (for cache, processor architecture reasons 8 byte alignment works best, so you cannot really have 8bit images of arbitrary width, they get a padding)

devshgraphicsprogramming: just pass src and dest row pitches, instead of `c`

src/irr/asset/CImageLoaderTGA.cpp
```diff
@@ -165,54 +189,52 @@ asset::IAsset* CImageLoaderTGA::loadAsset(io::IReadFile* _file, const asset::IAs
...
-							(int32_t*) palette, 0,
-							(header.ImageDescriptor&0x20)==0);
+					os::Printer::log("Loading 8-bit non-grayscale is NOT supported.", ELL_ERROR);
+					if (palette) delete [] palette;
+					if (data)    delete [] data;
+					
+					return nullptr;
 				}
+				
+				image = new asset::CImageData(NULL,nullOffset,imageSize,0,asset::EF_R8_UNORM);
```
devshgraphicsprogramming: make sure that TGA does not mandate by the spec for greyscale images to also be gamma-encoded

src/irr/asset/CImageLoaderPNG.cpp
```diff
@@ -152,31 +152,30 @@ asset::IAsset* CImageLoaderPng::loadAsset(io::IReadFile* _file, const asset::IAs
...
 	if (BitDepth < 8)
 	{
-		if (ColorType==PNG_COLOR_TYPE_GRAY || ColorType==PNG_COLOR_TYPE_GRAY_ALPHA)
-			png_set_expand_gray_1_2_4_to_8(png_ptr);
-		else
-			png_set_packing(png_ptr);
+		switch (ColorType) {
+			case PNG_COLOR_TYPE_GRAY:
+			case PNG_COLOR_TYPE_GRAY_ALPHA:
+				png_set_expand_gray_1_2_4_to_8(png_ptr);
```
devshgraphicsprogramming, line 165: don't expand grey or grey-alpha to rgb or rgba

devshgraphicsprogramming:
I did some reading on TGA, and found that TGA 2.0 should have a gamma field, then we could check that gamma for RGB/RGBA formats is reasonably close to 2.233..... and for greyscale+greyscale-alpha its 1.0 

and if its not then issue a warning to the log (in both release and debug builds)
http://tfc.duke.free.fr/coding/tga_specs.pdf

src/irr/asset/CImageLoaderPNG.cpp
ghost: It's basically expanding 1/2/4-bit gray to 8-bit gray (if there's any). Grayscale information is still kept intact

ghost:
Gamma information is located in the last 26 bytes of a TGA file, minus the signature ("`TRUEVISION-XFILE.`", 18 bytes)... specifically these 8 bytes:
image

Zeroes mean no gamma (it's optional), and from my personal testing with our local .tga located at the `media` folder & various online sample files, it seems like this is virtually non-existent and ignored by most readers/writers. Heck, even our current TGA writer just pad them with zeroes... same w/ a bunch of other parsers.

You can load up your TGAs & verify them yourself. Check for the ending 26 bytes for zeroes

So given that I could check if gamma value is present, it's mostly futile as nobody seems to bother with encoding extra metadata into their TGAs anywy.

devshgraphicsprogramming:
Ok, but our writer should write that gamma stuff, 1.0 for grayscale and 2.23333 for color

I still worry about greyscale TGAs, can you synthesise gradient (value 0-255) in a 256x256 8bit TGA grayscale with no gamma with our writer and open it in irfan-view + some other "legit" viewer (ideally photoshop/gimp) and paste a screenshot, one look and we'll know if TGA applications expect gamma grayscale.

ghost:
Here's the code that does it:
```
	uint32_t nullOffset[3] = {0, 0, 0}, imageSize[3] = {256, 256, 1};
	asset::CImageData *gray = new asset::CImageData(nullptr, nullOffset, imageSize, 0, asset::EF_R8_UNORM);
	
	uint8_t *data = (uint8_t *) gray->getData();
	
	for (int y = 0; y < 256; ++y) {
		for (int x = 0; x < 256; ++x)
			*(data + x) = 256 - x;
		data += gray->getPitchIncludingAlignment();
	}
	...
```

What it looks like in IrfanView:
image

What Krita has to say:
image

In case you wanna try it yourself: [Gradient.zip](https://github.com/buildaworldnet/IrrlichtBAW/files/3161332/Gradient.zip)

devshgraphicsprogramming:
ok then TGA expects grayscale to be gamma corrected as well, means that grayscale TGA is actually `EF_R8_SRGB`

Note: OpenGL doesn't even have that format, but its not the image loader's responsibility to convert between unsupported format and closest supported (as long as it has its own enum in E_FORMAT then don't touch it).

For testing you'll need a convert in the example 09 testing code.

devshgraphicsprogramming:
Run the same test for PNG and JPEG when ready

ghost:
Flipping issues have been dealt with. Only taking care of JPG/PNG grayscale & DDS loaders and we should be good to go (mass testing)

src/irr/asset/CImageLoaderTGA.cpp
```diff
@@ -77,9 +77,21 @@ bool CImageLoaderTGA::isALoadableFileFormat(io::IReadFile* _file) const
...
+	if (footer.ExtensionOffset == 0)
+		os::Printer::log("Gamma information is not present!", ELL_ERROR);
+	else
+	{
+		STGAExtensionArea extension;
+		_file->seek(footer.ExtensionOffset);
+		_file->read(&extension, sizeof(STGAExtensionArea));
+		
+		// TODO: Pass this to new asset::CImageData()
+		float gamma = extension.Gamma;
```
devshgraphicsprogramming, line 98: AFAIK if gamma is all zero then it means it also not present

ghost:
@devshgraphicsprogramming This is what I've got for JPG + PNG (JPG is saved as maximum quality, 100, same for PNG)

JPG: [GrayscaleJPG.zip](https://github.com/buildaworldnet/IrrlichtBAW/files/3167298/GrayscaleJPG.zip)
image

PNG: [GrayscalePNG.zip](https://github.com/buildaworldnet/IrrlichtBAW/files/3167300/GrayscalePNG.zip)
image

src/irr/asset/CImageWriterTGA.cpp
```diff
@@ -159,7 +159,7 @@ bool CImageWriterTGA::writeAsset(io::IWriteFile* _file, const SAssetWriteParams&
 	
 	STGAExtensionArea extension;
 	extension.ExtensionSize = sizeof(extension);
-	extension.Gamma = format == asset::EF_R8_UNORM ? 1.0f : ((100.0f / 30.0f) - 1.0f);
+	extension.Gamma = (format == asset::EF_R8_UNORM) ? 1.0f : ((100.0f / 30.0f) - 1.0f);
```
devshgraphicsprogramming: can you explain the second gamma choice? why does it work out to 2.33333 ?

IEC 61966-2-1 works out more closely to 2.23333

ghost: Hmmmmm yes, I meant 2.23333. Typo, my fault.

devshgraphicsprogramming: also, format==EF_R8_UNORM is not a check for gamma

devshgraphicsprogramming: we have this beauty for this
https://github.com/buildaworldnet/IrrlichtBAW/blob/315172ea3d1b9b51da4c1ef30412c02b99f1066a/include/irr/asset/EFormat.h#L1085

devshgraphicsprogramming: there should be a runtime counterpart to this, if not create one

devshgraphicsprogramming:
> 
> 
> @devshgraphicsprogramming This is what I've got for JPG + PNG (JPG is saved as maximum quality, 100, same for PNG)
> 
> JPG: [GrayscaleJPG.zip](https://github.com/buildaworldnet/IrrlichtBAW/files/3167298/GrayscaleJPG.zip)
> image
> 
> PNG: [GrayscalePNG.zip](https://github.com/buildaworldnet/IrrlichtBAW/files/3167300/GrayscalePNG.zip)
> image

In this case it looks like every image loader on earth expects input and output to be `EF_R8_SRGB` when working with grayscale.

Maybe we need a `createInConvertedFormat` method for `CImageData` that can create a clone of the image in a different pixel format using the format conversion functions (otherwise you won't be able to test the display of grayscale in OpenGL).

examples_tests/09.ColorSpaceTest/main.cpp
```diff
@@ -96,24 +96,32 @@ bool ReadLine(io::IReadFile* file, std::string &line)
...
 	
 	if (cputex[0])
 	{
-		io::path filename,extension ;
+		io::path filename, extension;
 		core::splitFilename(path.c_str(), nullptr, &filename, &extension);
 		filename += "."; filename += extension;
-
+		
+		bool writeable = (extension != "dds") && (extension != "bmp");
```
devshgraphicsprogramming, line 88: so you're nuking the BMP writer, correct?

devshgraphicsprogramming:
@Florastamine you may want to think about whether you want to deal with DDS directly or wish to solve the problem via GLI #253

examples_tests/09.ColorSpaceTest/main.cpp
ghost: Yes, I don't really see a reason why one would have a reason to write to BMP. Give me the final signal and I'll wipe the writer from the codebase.

devshgraphicsprogramming:
Regarding DDS, you can refuse to load all DirectX 9 DDS files that don't have a DX10-style extended header with proper format information (DXGI).

This will make it so that the only loadable DDS files can only be created with Photoshop+Intel Plugin or via the AMD Compressonator stand-alone tool.

The GIMP DDS plugin is very old and creates DirectX 9 DDS files which completely omit gamma information (they actually seem to be SRGB by default), I don't want our loader to play a guessing game here.

ghost:
A better thorough look at DDS and I thought I would advocate for GLI. Which means there are plenty of benefits beside not having to deal with the format from scratch:
* All formats (sRGB DXT1-5, ETC1-2, DX10 BC1-7, +compressed) plus cubemap & texture arrays are available from the get-go
* (Basically) free writer
* Accelerated decoding routines (GLI sits on top of GLM which already has SSE/AVX and NEON on ARM), also out of the box
* KTX, two birds with one stone (GLI)

devshgraphicsprogramming:
is GLI without GLM (or with GLM hidden) possible?

ghost:
Partly, it's integrated from within GLI and not exposed as a external dependency.

devshgraphicsprogramming:
ok, but does GLI's CMake look for GLM with `find_package`?

devshgraphicsprogramming:
Ok GLI (DDS & KTX) separate issue and PR, can you update your checklist and handle all greayscale as EF_R8_SRGB correctly (plus do a convert the unit test so OpenGL can display)?

src/irr/asset/CImageWriterJPG.cpp
```diff
@@ -155,6 +155,7 @@ static bool writeJPEGFile(io::IWriteFile* file, const asset::CImageData* image,
 				case asset::EF_A1R5G5B5_UNORM_PACK16:
 					video::convertColor<EF_A1R5G5B5_UNORM_PACK16, EF_R8G8B8_SRGB>(src_container, dest, 1, dim.X, dim);
 					break;
+				case asset::EF_R8_SRGB:
 				case asset::EF_R8_UNORM:
```
devshgraphicsprogramming, line 158: you cannot write it out with memcpy if its not SRGB! conversion needed

src/irr/asset/CImageWriterPNG.cpp
```diff
@@ -169,6 +171,7 @@ bool CImageWriterPNG::writeAsset(io::IWriteFile* _file, const SAssetWriteParams&
 	for (uint32_t i = 0; i < dim.Y; ++i)
 	{
 		switch (format) {
+			case asset::EF_R8_SRGB:
 			case asset::EF_R8_UNORM:
```
devshgraphicsprogramming, line 174: CONVERSION NEEDED, OR ERROR OUT

src/irr/asset/CImageWriterTGA.cpp
```diff
@@ -85,6 +81,7 @@ bool CImageWriterTGA::writeAsset(io::IWriteFile* _file, const SAssetWriteParams&
 				imageHeader.ImageDescriptor |= 1;
 			}
 			break;
+		case asset::EF_R8_SRGB:
 		case asset::EF_R8_UNORM:
```
devshgraphicsprogramming, line 85: convert or error out

src/irr/asset/CImageWriterTGA.cpp
```diff
@@ -122,6 +119,7 @@ bool CImageWriterTGA::writeAsset(io::IWriteFile* _file, const SAssetWriteParams&
 	{
 		switch (format) {
 			// Do a pass-through, byte-wise copy for everything else except EF_R8G8B8(A8)_SRGB, which would need swizzling to EF_B8G8R8(A8)_SRGB.
+			case asset::EF_R8_SRGB:
 			case asset::EF_R8_UNORM:
```
devshgraphicsprogramming: convert or error out

manhnt9:
I'm having build errors on Linux:
src/irr/asset/CImageWriterTGA.cpp:140:133: error: no matching function for call to ‘convertColor<EF_R8G8B8A8_SRGB, EF_B8G8R8A8_SRGB>(const void* [4], uint8_t*&, int, uint16_t&, irr::core::vector3d<unsigned int>)’
 SRGB>(src_container, row_pointer, 1, imageHeader.ImageWidth, image->getSize());

manhnt9:
Build log:

[ 63%] Building CXX object source/Irrlicht/CMakeFiles/Irrlicht.dir/__/__/src/irr/asset/CImageWriterTGA.cpp.o
/IrrlichtBAW/src/irr/asset/CImageWriterTGA.cpp: In member function ‘virtual bool irr::asset::CImageWriterTGA::writeAsset(irr::io::IWriteFile*, const irr::asset::IAssetWriter::SAssetWriteParams&, irr::asset::IAssetWriter::IAssetWriterOverride*)’:
/IrrlichtBAW/src/irr/asset/CImageWriterTGA.cpp:133:129: error: no matching function for call to ‘convertColor<EF_R8G8B8_SRGB, EF_B8G8R8_SRGB>(const void* [4], uint8_t*&, int, uint16_t&, irr::core::vector3d<unsigned int>)’
 SRGB>(src_container, row_pointer, 1, imageHeader.ImageWidth, image->getSize());
                                                                              ^
In file included from /IrrlichtBAW/source/Irrlicht/pch.h:162:0:
/IrrlichtBAW/include/irr/video/convertColor.h:398:17: note: candidate: template<irr::asset::E_FORMAT sF, irr::asset::E_FORMAT dF> void irr::video::convertColor(const void**, void*, uint64_t, uint32_t, uint32_t)
     inline void convertColor(const void* srcPix[4], void* dstPix, uint64_t _scale, uint32_t _blockX, uint32_t _blockY)
                 ^~~~~~~~~~~~
/IrrlichtBAW/include/irr/video/convertColor.h:398:17: note:   template argument deduction/substitution failed:
/IrrlichtBAW/src/irr/asset/CImageWriterTGA.cpp:133:127: note:   cannot convert ‘irr::asset::CImageData::getSize() const()’ (type ‘irr::core::vector3d<unsigned int>’) to type ‘uint32_t {aka unsigned int}’
 SRGB>(src_container, row_pointer, 1, imageHeader.ImageWidth, image->getSize());
                                                              ~~~~~~~~~~~~~~^~
In file included from /IrrlichtBAW/source/Irrlicht/pch.h:162:0:
/IrrlichtBAW/include/irr/video/convertColor.h:447:17: note: candidate: template<irr::asset::E_FORMAT sF, irr::asset::E_FORMAT dF> void irr::video::convertColor(const void**, void*, uint64_t, size_t, irr::core::vector3d<unsigned int>&)
     inline void convertColor(const void* srcPix[4], void* dstPix, uint64_t _scale, size_t _pixOrBlockCnt, core::vector3d<uint32_t>& _imgSize)
                 ^~~~~~~~~~~~
/IrrlichtBAW/include/irr/video/convertColor.h:447:17: note:   template argument deduction/substitution failed:
/IrrlichtBAW/src/irr/asset/CImageWriterTGA.cpp:133:127: note:   cannot convert ‘irr::asset::CImageData::getSize() const()’ (type ‘irr::core::vector3d<unsigned int>’) to type ‘irr::core::vector3d<unsigned int>&’
 SRGB>(src_container, row_pointer, 1, imageHeader.ImageWidth, image->getSize());
                                                              ~~~~~~~~~~~~~~^~
/IrrlichtBAW/src/irr/asset/CImageWriterTGA.cpp:140:133: error: no matching function for call to ‘convertColor<EF_R8G8B8A8_SRGB, EF_B8G8R8A8_SRGB>(const void* [4], uint8_t*&, int, uint16_t&, irr::core::vector3d<unsigned int>)’
 SRGB>(src_container, row_pointer, 1, imageHeader.ImageWidth, image->getSize());
                                                                              ^
In file included from /IrrlichtBAW/source/Irrlicht/pch.h:162:0:
/IrrlichtBAW/include/irr/video/convertColor.h:398:17: note: candidate: template<irr::asset::E_FORMAT sF, irr::asset::E_FORMAT dF> void irr::video::convertColor(const void**, void*, uint64_t, uint32_t, uint32_t)
     inline void convertColor(const void* srcPix[4], void* dstPix, uint64_t _scale, uint32_t _blockX, uint32_t _blockY)
                 ^~~~~~~~~~~~
/IrrlichtBAW/include/irr/video/convertColor.h:398:17: note:   template argument deduction/substitution failed:
/IrrlichtBAW/src/irr/asset/CImageWriterTGA.cpp:140:131: note:   cannot convert ‘irr::asset::CImageData::getSize() const()’ (type ‘irr::core::vector3d<unsigned int>’) to type ‘uint32_t {aka unsigned int}’
 SRGB>(src_container, row_pointer, 1, imageHeader.ImageWidth, image->getSize());
                                                              ~~~~~~~~~~~~~~^~
In file included from /IrrlichtBAW/source/Irrlicht/pch.h:162:0:
/IrrlichtBAW/include/irr/video/convertColor.h:447:17: note: candidate: template<irr::asset::E_FORMAT sF, irr::asset::E_FORMAT dF> void irr::video::convertColor(const void**, void*, uint64_t, size_t, irr::core::vector3d<unsigned int>&)
     inline void convertColor(const void* srcPix[4], void* dstPix, uint64_t _scale, size_t _pixOrBlockCnt, core::vector3d<uint32_t>& _imgSize)
                 ^~~~~~~~~~~~
/IrrlichtBAW/include/irr/video/convertColor.h:447:17: note:   template argument deduction/substitution failed:
/IrrlichtBAW/src/irr/asset/CImageWriterTGA.cpp:140:131: note:   cannot convert ‘irr::asset::CImageData::getSize() const()’ (type ‘irr::core::vector3d<unsigned int>’) to type ‘irr::core::vector3d<unsigned int>&’
 SRGB>(src_container, row_pointer, 1, imageHeader.ImageWidth, image->getSize());
                                                              ~~~~~~~~~~~~~~^~
source/Irrlicht/CMakeFiles/Irrlicht.dir/build.make:868: recipe for target 'source/Irrlicht/CMakeFiles/Irrlicht.dir/__/__/src/irr/asset/CImageWriterTGA.cpp.o' failed
make[2]: *** [source/Irrlicht/CMakeFiles/Irrlicht.dir/__/__/src/irr/asset/CImageWriterTGA.cpp.o] Error 1
CMakeFiles/Makefile2:2856: recipe for target 'source/Irrlicht/CMakeFiles/Irrlicht.dir/all' failed
make[1]: *** [source/Irrlicht/CMakeFiles/Irrlicht.dir/all] Error 2
Makefile:151: recipe for target 'all' failed
make: *** [all] Error 2

devshgraphicsprogramming:
> I'm having build errors on Linux:
> src/irr/asset/CImageWriterTGA.cpp:140:133: error: no matching function for call to ‘convertColor<EF_R8G8B8A8_SRGB, EF_B8G8R8A8_SRGB>(const void* [4], uint8_t*&, int, uint16_t&, irr::core::vector3d)’
> SRGB>(src_container, row_pointer, 1, imageHeader.ImageWidth, image->getSize());

`getSize` returns a uint32_t[3] not a vector3d&

ghost:
Putting Linux-related build fixes to the list.

ghost:
@devshgraphicsprogramming Please give final confirmation if everything looks correct.

By the way, I'm putting together a test suite, which will be in `./examples_tests/media/color_space_test`. For sample TGAs, textures from the sponza scene will be used.

Florastamine:
I'm getting together a test suite, which will be put in `./examples_tests/media/color_space_test`. For sample TGAs, textures from the sponza scene will be used instead.

ghost:
> Maybe we need a `createInConvertedFormat` method for `CImageData` that can create a clone of the image in a different pixel format using the format conversion functions (otherwise you won't be able to test the display of grayscale in OpenGL).

Also please give your confirmation on this one

devshgraphicsprogramming:
Alpha SRGB loading and writing Alpha non-SRGB and SRGB works for PNG/TGA?

devshgraphicsprogramming:
One last thing left to do
https://github.com/buildaworldnet/IrrlichtBAW/issues/234#issuecomment-494325927

I propose here https://github.com/buildaworldnet/IrrlichtBAW/blob/315172ea3d1b9b51da4c1ef30412c02b99f1066a/include/SColor.h#L404

something like (within the scope of SColorf)
```cpp
inline static SColorf fromSRGB(SColorf&& input)
{
// ... implementation (use Crissp's functions)
}
inline static SColorf toSRGB(SColorf&& input)
{
// ... implementation (use Crissp's functions)
}
```

devshgraphicsprogramming:
> 
> 
> > Maybe we need a `createInConvertedFormat` method for `CImageData` that can create a clone of the image in a different pixel format using the format conversion functions (otherwise you won't be able to test the display of grayscale in OpenGL).
> 
> Also please give your confirmation on this one

Yes, this method should come in two variants, first inline and with the output format as a template parameter and the other with the output format as a argument.

It has to create a clone of the `CImageData` (if it can, else return nullptr, not every format can be encoded into or supports the original texture's dimensions, i.e. you cannot convert TO a compressed format) that has EXACTLY IDENTICAL:
- type
- dimensions
- everything

except for the pixel format ;)

You can actually re-use that function in your writers and loaders after refactoring ;)

ghost:
> Alpha SRGB loading and writing Alpha non-SRGB and SRGB works for PNG/TGA?

@devshgraphicsprogramming reading & writing-wise, yes. This is what the writers spit out, a R8G8B8A8_SRGB:
image


which is [100% correct against the input image](http://www.schaik.com/pngsuite/tm3n3p02.png).

However this is what the framebuffer got instead (clear color 100% RGB(0,0,0)):
image

Crisspl:
did u enable blending?

devshgraphicsprogramming:
its more like, "did you download from framebuffer with alpha"

manhnt9:
I think "did you download from framebuffer with alpha" means that Flor should create the framebuffer having format that supports alpha.

devshgraphicsprogramming:
I think thats why there was originally a Render-To-Texture before presenting to screen.

The non-transparent blue look is ok when viewed in a window, its expected.

ghost:
Ouch.

manhnt9:
Please avoid force push, I was getting merge conflicts and had to revert back 100 commits to pull from this branch again. Anyways it builds fine.

manhnt9:
I don't have anything generated in bin folder after running the test. Is this normal?

manhnt9:
The console output is:

Irrlicht Engine version 0.3.0-beta2
Linux 4.15.0-46-generic #49-Ubuntu SMP Wed Feb 6 09:33:07 UTC 2019 x86_64
Creating X window...
Using renderer: OpenGL 4.6.0
GeForce GTX 1050 Ti/PCIe/SSE2: NVIDIA Corporation
GLSL version: 46.0

ghost:
Fixed it. My own `ReadLine()` apparently caused issues reading EOL bytes, so I scrapped the method completely and just use C++'s `std::getline()`.

include/SColor.h
```diff
@@ -416,6 +417,22 @@ namespace video
 			const float inv = 1.0f / 255.0f;
 			*this *= inv;
 		}
+		
+		inline static SColorf fromSRGB(SColorf&& input)
+		{
+			float color[3] = {input.r, input.g, input.b};
+			impl::SRGB2lin<float>(color);
```
devshgraphicsprogramming, line 424: you could just pass `reinterpret_cast<float[3]>(input.pointer)` or `reinterpret_cast<float[3]>(&input.r)` instead

devshgraphicsprogramming: then an `std::move()` of the `input` into the `SColorf` constructor, like this
```cpp
return SColorf(std::move(input));
```

include/irr/video/decodePixels.h
```diff
@@ -1283,6 +1286,17 @@ namespace irr { namespace video
                 else s = std::pow((s + 0.055) / 1.055, 2.4);
             }
         }
+		
+        template<>
+        inline void lin2SRGB<double>(double _lin[3])
```
devshgraphicsprogramming, line 1291: @Crisspl  has this very function in `encodePixels.h` I believe?

include/irr/video/decodePixels.h
```diff
@@ -1293,6 +1307,16 @@ namespace irr { namespace video
             for (uint32_t i = 0; i < 3u; ++i)
                 lin[i] = s[i] * 255.;
         }
+
+        template<typename T>
```
devshgraphicsprogramming, line 1311: templating is OK, but you need a `static_assert` which ensures that `T` can only be `float` or `double`

devshgraphicsprogramming:
I think we're nearly done, @Florastamine do you need https://github.com/buildaworldnet/IrrlichtBAW/pull/273#issuecomment-494343988 ?

include/irr/video/decodePixels.h
ghost: Yes, but `lin2SRGB()` on `encodePixels.h` only works with `double`, and downcasting is ugly. I can move my conversion functions there though.

manhnt9:
I still have same output on Linux, @Florastamine did you test on your Linux too?

devshgraphicsprogramming:
@manhnt9 when @Florastamine is done, how shall we handle it:

A) I merge this into IrrlichtBAW:sRGB, then you merge that into branches of manhnt9 ?
B) You merge this into manhtnt9:sRGB, we close this PR and it gets merged when the PR #291 gets merged?

manhnt9:
We should follow A if there is something that needs those loaders.
Or else let's go B and merge after we have the sRGB working for the Explorer, that way we would be able to test and fix the loaders more if needed, and guarantee the compatibility when plugging those behind a CEGUI img loader.

manhnt9:
Console output when running this example:

Irrlicht Engine version 0.3.0-beta2
Linux 4.15.0-46-generic #49-Ubuntu SMP Wed Feb 6 09:33:07 UTC 2019 x86_64
Creating X window...
Using renderer: OpenGL 4.6.0
GeForce GTX 1050 Ti/PCIe/SSE2: NVIDIA Corporation
GLSL version: 46.0
Reading: ../../media/color_space_test/R8G8B8_1.jpg
Reading: ../../media/color_space_test/R8G8B8_2.jpg
Reading: ../../media/color_space_test/R8_1.jpg
Unsupported upload format
FBO has one or several incomplete image attachments
FBO error
FBO incomplete
FBO revalidation failed!
Unsupported upload format
Unsupported color format, operation aborted.
Reading: ../../media/color_space_test/R8_2.jpg
Unsupported upload format
FBO has one or several incomplete image attachments
FBO error
FBO incomplete
FBO revalidation failed!
Unsupported upload format
Unsupported color format, operation aborted.
Reading: ../../media/color_space_test/R8G8B8_1.png
PNG warning: Interlace handling should be turned on when using png_read_image
Reading: ../../media/color_space_test/R8G8B8_2.png
Reading: ../../media/color_space_test/R8G8B8A8_1.png
Reading: ../../media/color_space_test/R8G8B8A8_2.png
Reading: ../../media/color_space_test/R8_1.png
Unsupported upload format
FBO has one or several incomplete image attachments
FBO error
FBO incomplete
FBO revalidation failed!
Unsupported upload format
Unsupported color format, operation aborted.
Reading: ../../media/color_space_test/R8_2.png
Unsupported upload format
FBO has one or several incomplete image attachments
FBO error
FBO incomplete
FBO revalidation failed!
Unsupported upload format
Unsupported color format, operation aborted.
Reading: ../../media/sponza/textures/background.tga
Gamma information is not present!
Reading: ../../media/color_space_test/R8G8B8_RLE.tga
Gamma information is not present!
Reading: ../../media/color_space_test/R8G8B8A8.tga
Gamma information is not present!
Reading: ../../media/color_space_test/R8G8B8A8_RLE.tga
Gamma information is not present!
Reading: ../../media/color_space_test/R8.tga
Gamma information is not present!
Unsupported upload format
FBO has one or several incomplete image attachments
FBO error
FBO incomplete
FBO revalidation failed!
Unsupported upload format
Unsupported color format, operation aborted.

include/SColor.h
ghost: ```impl::lin2SRGB<float>(reinterpret_cast<float[3]>(input.pointer));```
```impl::lin2SRGB<float>(reinterpret_cast<float[3]>(&input.r));```

Both failed on me (both MSVC and GCC): ```error: invalid cast from type ‘float*’ to type ‘float [3]’```

**Comment by ghost** [comment]
> > > Maybe we need a `createInConvertedFormat` method for `CImageData` that can create a clone of the image in a different pixel format using the format conversion functions (otherwise you won't be able to test the display of grayscale in OpenGL).
> > 
> > 
> > Also please give your confirmation on this one
> 
> Yes, this method should come in two variants, first inline and with the output format as a template parameter and the other with the output format as a argument.
> 
> It has to create a clone of the `CImageData` (if it can, else return nullptr, not every format can be encoded into or supports the original texture's dimensions, i.e. you cannot convert TO a compressed format) that has EXACTLY IDENTICAL:
> 
>     * type
> 
>     * dimensions
> 
>     * everything
> 
> 
> except for the pixel format ;)
> 
> You can actually re-use that function in your writers and loaders after refactoring ;)



> I think we're nearly done, @Florastamine do you need [#273 (comment)](https://github.com/buildaworldnet/IrrlichtBAW/pull/273#issuecomment-494343988) ?

I'm still not so sure on how I would tackle the matter most efficiently... given there are a bunch of pixel formats to work with + all the possible combinations on what would/wouldn't work, plus (as you said) texture dimensions and whatnot. Right now, to get the testing going I have only got the thing to convert R8_SRGB -> R8_UNORM to display the grayscale images.

```
// Templating this thing is tricky since the source & target format can often only be determined during runtime, hence the if(...), which is cumbersome. Ideas?
CImageData* createInConvertedFormat(E_FORMAT destFormat)
{
	asset::E_FORMAT format = getColorFormat();

	uint32_t nullOffset[3] = {0, 0, 0};
	uint32_t imageSize[3] = {maxCoord[0] - minCoord[0], maxCoord[1] - minCoord[1], maxCoord[2] - minCoord[2]};
	
	CImageData* image = new CImageData(nullptr, nullOffset, imageSize, 0, destFormat);
    
	uint8_t* in = (uint8_t*) getData();
	uint8_t* out = (uint8_t*) image->getData();

        if (format == asset::EF_R8_SRGB && destFormat == asset::EF_R8_UNORM)
        {
            // video::convertColor<EF_R8_SRGB, EF_R8_UNORM>(...)
        }
	
	return image;
}
```

**Comment by devshgraphicsprogramming** [comment]
You're doing it "wrong"
```cpp
template<E_FORMAT inFormat, E_FORMAT outFormat>
CImageData* createInConvertedFormat()
{
	E_FORMAT format = getColorFormat();
        assert(format==inFormat);
        E_FORMAT destFormat = outFormat;
        if (isDepthOrStencilFormat(destFormat)||
            return nullptr;
        // check if new format can handle the alignment
        {
            auto dims = getBlockDimensions(destFormat);
            for (uint32_t i=0u; i<3u; i++)
            if ((minCoord[i]%dims[i])!=0u || (maxCoord[i]%dims[i])!=0u)
                return nullptr;
        }
        uint32_t blockByteSize = getTexelOrBlockSize(destFormat);
	
	CImageData* image = new CImageData(nullptr, minCoord, maxCoord, mipLevelHint, destFormat, 0x1u<<core::findLSB(blockByteSize));

        for (auto z=minCoord[2]; z<maxCoord[2];  z++)
        for (auto y=minCoord[1]; y<maxCoord[1];  y++)
        {
            // get appropriate row pointers
	    uint8_t* in = reinterpret_cast<uint8_t*>(getData())+...;
	    uint8_t* out = reinterpret_cast<uint8_t*>(image->getData())+...;

            video::convertColor<inFormat, outFormat>(...)
        }
	
	return image;
}
```

devshgraphicsprogramming:
Then the non templated version would look like this

```cpp
CImageData* createInConvertedFormat(E_FORMAT destFormat)
{
	E_FORMAT format = getColorFormat();
        if (isDepthOrStencilFormat(destFormat)||isBlockCompressionFormat(destFormat))
            return nullptr;
        // check if new format can handle the alignment
        {
            auto dims = getBlockDimensions(destFormat);
            for (uint32_t i=0u; i<3u; i++)
            if ((minCoord[i]%dims[i])!=0u || (maxCoord[i]%dims[i])!=0u)
                return nullptr;
        }
        uint32_t blockByteSize = getTexelOrBlockSize(destFormat);
	
	CImageData* image = new CImageData(nullptr, minCoord, maxCoord, mipLevelHint, destFormat, 0x1u<<core::findLSB(blockByteSize));

        for (auto z=minCoord[2]; z<maxCoord[2];  z++)
        for (auto y=minCoord[1]; y<maxCoord[1];  y++)
        {
            // get appropriate row pointers
	    uint8_t* in = reinterpret_cast<uint8_t*>(getData())+...;
	    uint8_t* out = reinterpret_cast<uint8_t*>(image->getData())+...;

            video::convertColor(format,destFormat...)
        }
	
	return image;
}
```

devshgraphicsprogramming:
Then I'd advise pulling `irr/video/convertColor.h`, `irr/video/encodePixels.h`, `irr/video/decodePixels.h` out of the `video` namespace and putting them in the `asset` namespace instead

src/irr/asset/CImageLoaderTGA.cpp
manhnt9: @devshgraphicsprogramming @Florastamine: is this solved?

devshgraphicsprogramming: no idea where that function is now

ghost:
Non-templated `createInConvertedFormat()` isn't viable as non-templated `convertColor()` is currently unimplemented (only the prototype) for now.

https://github.com/buildaworldnet/IrrlichtBAW/blob/d193ec992870247d86ff5dad09f66f3a89af9676/include/irr/video/convertColor.h#L492

devshgraphicsprogramming:
@Crisspl ?

ghost:
By the way, since the loaders/writers themshelves are now working, would you like to merge the PR as-is so @manhnt9 can continue with sRGB + BRDF explorer testing? I'll try to finish the rest of the testing routines and that could come at a later PR.

devshgraphicsprogramming:
I think a non-templated `convertColor` could be generated via a script or something (maybe via CMake configure file thing).

It would just be a massive switch statement that calls an `impl::convertColor<SWITCH_CASE_SOURCE_FORMAT>(_dfmt,...)` which itself is a massive switch statement.

------------------------------------------------------------------------------------------------------------------

Another implementation would be to have two independent (non-nested) switches, one for encode and one for decode. The only bizzare thing would be 3 declarations of temporary pixel values as uint64_t[4], int64_t[4] and double[4], of which 2 are always unused. That would be better on the compiler (time), code gen (time and space) and library size.

I think such a function could be put together by simply 2 calls, `decodePixels` followed by `encodePixels`.

devshgraphicsprogramming:
@Florastamine great idea, open a separate PR called "WIP: CImageData::createInConvertedFormat"

src/irr/asset/CImageLoaderTGA.cpp
devshgraphicsprogramming: @Florastamine?

ghost: It's solved - I fixed the stride as instructed.
https://github.com/buildaworldnet/IrrlichtBAW/blob/e28d94463b1e9b758b83a50a79c77ac6cb329677/src/irr/asset/CImageLoaderTGA.cpp#L121

devshgraphicsprogramming: still not fixed, using same stride for both the `in` and `out` pointer

include/irr/video/decodePixels.h
devshgraphicsprogramming: you can make his templated, and slam an assert about the template type (onyl float or double) in there

include/SColor.h
devshgraphicsprogramming: ah, I forgot what the syntax for this is
https://stackoverflow.com/questions/17258749/is-there-a-way-of-casting-a-pointer-to-an-array-type

include/irr/video/decodePixels.h
devshgraphicsprogramming: @Florastamine, we need a static_assert, dont want someone using that with ints or vector classes

Crisspl:
@Florastamine

>non-templated convertColor() is currently unimplemented (only the prototype) for now.

???? It should be in convertColor.cpp, but in fact I don't see such file on github repo. Maybe unpushed file... I'll make sure when I get home in about an hour

devshgraphicsprogramming:
oopsie ;)

P.S. Better add it to the CMake as well XD

Crisspl:
Ok, so it was in the repo but in `/source/Irrlicht/` however it was not added to project. I moved it to `/src/irr/video` and added in cmake script. Note: turns out it needs `/bigobj` option (msvc) to compile because it generates too much code... Check whether it compiles on gcc without any extra options. @devshgraphicsprogramming so maybe decide whether you even want this
Fixed in https://github.com/buildaworldnet/IrrlichtBAW/pull/304

ghost:
Alright, I'll be testing out on GCC soon, thank you. Now that the file is included in the build, which explains why I kept getting linker errors despite the file is already there.

Crisspl:
Well, nobody ever noticed because it was never used

manhnt9:
Please notify me when the loaders are done. Seems like still having some minor issues left.

src/irr/asset/CImageLoaderTGA.cpp
manhnt9: @Florastamine

ghost: `src` is a raw pixel buffer loaded via libjpeg, libpng or `IReadFile::read()`. `convertColorFlip()`'s job is just filling a `CImageData` with that raw buffer, doing color conversion line-by-line in between.

It doesn't do `CImageData`-to-`CImageData` type of conversion, hence only 1 stride

ghost: plus if you look at `CImageLoaderTGA.cpp`, dest and src formats on where `convertColorFlip<>()` is used is almost the same, because it's mostly only used for flipping images/shuffling RGBA order

devshgraphicsprogramming: > It doesn't do CImageData-to-CImageData type of conversion, hence only 1 stride

If you expect those kind of things, then you need to place `assert`s in the code

PR #282, opened 2019-04-29
Unified and synced up all branches

devshgraphicsprogramming:
lets fix this build mess once and for all


manhnt9:
Well you tagged the wrong person 3 days ago, that's why I didn't have any notifications about this. My account is manhnt9, manhnt is someone else.

manhnt9:
set(PNG_BUILD_ZLIB ON CACHE BOOL "Custom zlib Location, else find_package is used")

Maybe your CMake cache is already set and the set command is ignored.

Issue #279, opened 2019-04-29
BRDF Explorer project


manhnt9:
This issue is meant to discuss software functions of BRDF Explorer.

Rendering algorithms will be discussed in #276.
- [x] ~Include [script file](https://github.com/mnauw/git-remote-hg) to clone my CEGUI branch as git submodule~
- [x] ~Apply @Florastamine's theme patch~
- [x] Push submodules online
- [x] Build CEGUI dependencies
- [x] Build CEGUI
- [x] Add new IrrlichtBAW example `13.BRDF_Explorer`
- [x] Write CEGUI management class in `ext namespace` to init/destroy/render GUI
- [x] Install CEGUI headers to INSTALL_PATH/include/irr/ext/GUI/CEGUI
- [x] Port GUI
- [x] Relocate submodules into 3rdparty
- [x] Compile cegui ext as static lib
- [x] Build dependencies on Linux instead of requiring user to install via apt
- [x] Refactor `Helpers` by @Florastamine
- [ ] Implement IrrBAW CEGUI texture loader in `ext namespace` 
- [ ] Patch CEGUI to upload sRGB (issue #263)
- [x] Render a simple model
- [x] Solve comment https://github.com/buildaworldnet/IrrlichtBAW/pull/291#discussion_r282012784, https://github.com/buildaworldnet/IrrlichtBAW/pull/291#discussion_r282012245

manhnt9:
Note: For cloning CEGUI on Windows, Python folder should be added into PATH environment variable. For example it's OK if you can call `python` from CMD. Python 2 is required for this script to treat Mercurial repo with Git.

Asking approval from @devshgraphicsprogramming and @Crisspl, or else we can download a latest, compressed CEGUI branch before building from these [links](https://bitbucket.org/manhnt9/cegui/downloads/?tab=branches), but that's a little hard for updating CEGUI.

devshgraphicsprogramming:
Python 2 (I have 3 but cmake is not finding it) seems to be necessary for some of our SPIR-V dependencies to not create disturbing errors when building with MSVC ( #269 ).

I'm all for Python 2 being a dependency, but please please please make CMake on Windows find it if its installed.

If possible I'd use and require Python 3 instead of 2 and solve #269 with Python 3, not 2.

[...]

PR #291, opened 2019-05-01
WIP: Example 13.BRDF_Explorer

manhnt9:
#279 
MSVC 2019 will be supported first, then Linux GCC.

Changes to irrbaw:
- Add IRR_BUILD_CEGUI option, default is OFF to save time
- Root CMake detect and set `${IRR_PARALLEL_BUILD_JOBS}`, you can use for subprojects too
- New global variables: CEGUI_INCLUDE_PATH, CEGUI_LIBS

DevSH's CMake guidelines (copied from Discord chat)

my main guideline for CMake and modules etc
is that build system is not allowed to make files etc outside of the build directory
with the SOLE exception of OpenSSL
also stuff installed SHOULD only be irrlicht headers and irrlicht static lib, there might be a legitimate reason to go against this
dependencies should not install together with that (which means they must be statically linked so they are inside irrlicht lib) - example, none of libpng, libjpeg etc are visible to the outside nor are they installed
the only exception is a dependency of an irr::ext (since irr::ext can be enabled and disabled separately from the main lib, and main lib never actually includes and links extensions)
then headers CAN install (if necessary) to $INSTALL_PREFIX/include/irr/ext/YOUR_EXT_NAME/3rdparty/YOUR_EXTS_3RDPARTY_DEPENDENCY
the library CAN install to $INSTALL_PREFIX/lib/irr/ext/YOUR_EXT_NAME/3rdparty/YOUR_EXTS_3RDPARTY_DEPENDENCY (but why do it if it can be statically linked?)
Example, Bullet with an $INSTALL_PREFIX=./install/win64-msvc/
Sample Header

./install/win64-gcc/include/irr/ext/Bullet/3rdparty/include/bullet/Bullet-C-Api.h

Sample lib

./install/win64-gcc/lib/irr/ext/Bullet/3rdparty/libBulletCollision.so

however if your irr::ext::something requires 3rdparty libraries you should really build it as a static library itself, install the irr::ext::something library and not install the dependencies.
Also think if you "really" need the top level API of your irr::ext to expose the 3rdparty dependency headers... unless you expect the user to want to have access to the internals, maybe don't expose
For example it may be good for users to have access to all of Bullet's API, that saves irr::ext::bullet being a silly pointless wrapper of the entire Bullet API


manhnt9:
Changes to irrbaw:
- Add IRR_BUILD_CEGUI option, default is OFF to save time

3rdparty/CEGUI/CMakeLists.txt
```diff
@@ -0,0 +1,73 @@
...
+          -DCEGUI_BUILD_TINYXML=OFF
+          -DCEGUI_BUILD_TOLUAPP=OFF
+          -DCEGUI_BUILD_XERCES=OFF
+          -DCEGUI_BUILD_ZLIB=OFF
+          -DCEGUI_MSVC_STATIC_RUNTIME=ON
+        WORKING_DIRECTORY ${CEGUI_DEPS_BUILD_PATH}
+    )
+
+    execute_process(
+        COMMAND ${CMAKE_COMMAND} --build ${CEGUI_DEPS_BUILD_PATH} --config Release -j 4
```
ghost: I would use `$ENV{NUMBER_OF_PROCESSORS}` instead of hardcoding the number of jobs to 4.

`NUMBER_OF_PROCESSORS` (exists on Windows only) defines the number of available processors

devshgraphicsprogramming: just make it a `max($ENV,8)` ;)

manhnt9: thanks for suggestions, I will update

manhnt9: updated

manhnt9:
CEGUI build is integrated in to irrbaw, but don't try it yet, still has some errors.

The building process is done, I will have to set correct configurations for CEGUI.

manhnt9:
Proposed changes, please review @devshgraphicsprogramming @Crisspl.
(Approved by devsh on Discord).

In order to link more libs to an example, we'd have to modify `macro(irr_create_executable_project)`.
CMake complains if I use `target_link_libraries` in CMake of example, after `irr_create_executable_project("" "")` since it's called before in the `macro(irr_create_executable_project)`.

So `target_link_libraries` should be called once.

To let example add more libs, the macro can be changed to: `macro(irr_create_executable_project _EXTRA_SOURCES _EXTRA_OPTIONS _EXTRA_INCLUDES _EXTRA_LIBS)`.

Further changes required:
- Find and replace `irr_create_executable_project("" "")` with `irr_create_executable_project("" "" "" "")`

devshgraphicsprogramming:
Double approved.

manhnt9:
I pushed the changes several hours ago, it appears before the comment because my Windows time is f**ked by Linux. Have just fixed my LVM partitions, relocated and resized, etc so still no code yet. My environment is stable now. From here we can expect frequent code updates.

manhnt9:
Linux can now build CEGUI and base project.

manhnt9:
In the waiting of the IMG loaders, I will use STB image for the time being.

manhnt9:
The progress is still new, I will notify when it's ready to be reviewed.

manhnt9:
According to @Florastamine, Native File Dialogs library has bugs on Linux so he had to replace with [Portable File Dialogs](https://github.com/samhocevar/portable-file-dialogs), and it works well.

Bug description:
The implementation of NDF which uses zenity (nfd_zenity.c) on a few occasions would skip right through the file dialog, which would return garbage because the window just flashes and closes right after without the user be able to choose anything.
PDF also uses zenity and doesn't exhibit the same problem, plus its single header.
It's in active development (more than NDF) and is made by a OSS supporter Sam Hocevar, the guy who invented WTFPL license.

examples_tests/13.BRDF_Explorer/main.cpp
```diff
@@ -0,0 +1,99 @@
+#define _IRR_STATIC_LIB_
+
+/* GCC compile Flags
```
devshgraphicsprogramming: remove this comment, also from other examples if you desire to do so

examples_tests/CMakeLists.txt
```diff
@@ -9,6 +9,7 @@ add_subdirectory(09.ColorSpaceTest EXCLUDE_FROM_ALL)
 add_subdirectory(10.AllocPerfTest EXCLUDE_FROM_ALL)
 add_subdirectory(11.RayCastCollision EXCLUDE_FROM_ALL)
 add_subdirectory(12.Splines EXCLUDE_FROM_ALL)
+add_subdirectory(13.BRDF_Explorer EXCLUDE_FROM_ALL)
```
devshgraphicsprogramming, line 13: only if irrbaw is configured to be built with CEGUI

.gitmodules
```diff
@@ -47,3 +47,9 @@
 [submodule "3rdparty/CEGUI/cegui"]
 	path = 3rdparty/CEGUI/cegui
 	url = git@gitlab.com:InnerPieceOSS/cegui.git
+[submodule "ext/CEGUI/portable-file-dialogs"]
+	path = ext/CEGUI/portable-file-dialogs
+	url = git@github.com:samhocevar/portable-file-dialogs.git
+[submodule "ext/CEGUI/stb"]
```
devshgraphicsprogramming, line 51: if we're not gonna keep something around, don't make it a submodule... submodules are hard to get rid of later on

.gitignore
```diff
@@ -7,6 +7,9 @@
 *.objs*
 *.d
 *.o
+*.gch
+# PCH generated by build
+source/Irrlicht/server_pch.h
```
devshgraphicsprogramming: @Crisspl fix please.

3rdparty/CEGUI/CMakeLists.txt
```diff
@@ -71,6 +72,9 @@ if (MSVC) # only MSVC needs to download and build dependencies of CEGUI
     )
     message(STATUS "Installing done")
 
+else()
+    find_package(EXPAT REQUIRED)
```
devshgraphicsprogramming: the final deliverable (when we actually merge) needs those build on Linux as well, or at the very least EXPAT (since freetype is ubiquitous).

If for some reason its not appropriate to build these ourselves, then they should at the very least be statically linked.

examples_tests/13.BRDF_Explorer/main.cpp
manhnt9: Tried removing other examples, my editor produces ^M as EOL so there're more changes in a commit. I'm only committing for my example for now.

examples_tests/CMakeLists.txt
manhnt9: updated.

3rdparty/CEGUI/CMakeLists.txt
manhnt9: I can just remove MSVC check and let all compilers build the deps. Since the deps source is already portable.

.gitmodules
manhnt9: submodule will introduce much less changes in your git history than directly put the code files here.

manhnt9:
I had to remove this include, it produces conflicts with glew that CEGUI uses.
#include "../source/Irrlicht/COpenGLExtensionHandler.h"

manhnt9:
Current CMake warning:
CMake Error at cmake/common.cmake:12 (add_executable):
  Target "brdf_explorer" links to item
  "IrrlichtBAW/build/3rdparty/CEGUI/build/install/lib/CEGUIOpenGLRenderer-0_Static.lib
  " which has leading or trailing whitespace.  This is now an error according
  to policy CMP0004.
Call Stack (most recent call first):
  examples_tests/13.BRDF_Explorer/CMakeLists.txt:13 (irr_create_executable_project)

I have to keep this because of the CMake's nature. The build still works correctly though. In fact, it needs this warning to work.

For example:
```
set(VAR
  lib1
  lib2
)
message(STATUS ${VAR}) # will print lib1lib2

# so I have to use
set(VAR
  "lib1 "
  "lib2 "
)
```

.gitignore
Crisspl: https://github.com/buildaworldnet/IrrlichtBAW/blob/master/source/Irrlicht/CMakeLists.txt#L376

Crisspl: IrrlichtServer has to make its own pch because of different flags etc. However maybe it could be better to have in repo pch_server.h with just `#include "pch.h"` instead of making cmake creating it

examples_tests/13.BRDF_Explorer/BRDF.cpp
```diff
@@ -0,0 +1,49 @@
...
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
+
+*/
+
+#include "BRDF.h"
+#include "../../ext/CEGUI/CEGUI.h"
+
+namespace irr
```
devshgraphicsprogramming: dont use namespace `irr` outside the library source, ever

examples_tests/13.BRDF_Explorer/CMakeLists.txt
```diff
@@ -4,8 +4,14 @@ if(NOT RES)
 	message(FATAL_ERROR "common.cmake not found. Should be in {repo_root}/cmake directory")
 endif()
 
+set(BRDF_SOURCES
```
devshgraphicsprogramming: rename the BRDF prefix to BRDF_EXPLORER, I don't want it colliding with actual pure-BRDF stuff later on or to be misunderstood

examples_tests/13.BRDF_Explorer/CMakeLists.txt
```diff
@@ -4,8 +4,14 @@ if(NOT RES)
...
+	"../../ext/CEGUI/CEGUI.cpp"
```
devshgraphicsprogramming: because of the pure insane number of dependencies, `irr::ext::cegui` should be its own statically linked library

examples_tests/13.BRDF_Explorer/main.cpp
```diff
@@ -37,7 +36,7 @@ int main()
     video::IVideoDriver* driver = device->getVideoDriver();
     scene::ISceneManager* smgr = device->getSceneManager();
 
-    auto cegui = ext::cegui::createGUIManager(driver);
+    auto* brdf = new BRDF(driver);
```
devshgraphicsprogramming: what is the `BRDF` object going to be responsible for?

examples_tests/13.BRDF_Explorer/BRDF.h
```diff
@@ -48,6 +48,8 @@ class BRDF {
         BRDF(video::IVideoDriver* driver);
```
devshgraphicsprogramming: ok you need to rename this object to `BRDFExplorerApp` or something similar, simply calling it `BRDF` is really confusing

ext/CEGUI/CEGUI.cpp
```diff
@@ -97,6 +98,15 @@ void GUIManager::init()
     ));
 }
 
+void GUIManager::render()
+{
+    auto renderState = irr::video::COpenGLState::collectGLState();
```
devshgraphicsprogramming: `collectGLState` and `executeGLDiff` take a list of booleans about what state it should care about.

CEGUI uses OpenGL 3 so things like SSBOs, Storage Images, etc. shouldn't matter

Also if you care about current FBO, Viewport, etc then it will change those, sometimes that is not what you want, this is why in the silverlining example I told it to not care about FBO, depth test, viewport etc. etc. because I wanted to render it into a currently texture with currently set viewport size and depth testing.

ext/CEGUI/CEGUI.cpp
```diff
@@ -97,6 +98,15 @@ void GUIManager::init()
...
+    irr::video::COpenGLState guiState;
```
devshgraphicsprogramming: default constructor will make an `COpenGLState` object that is as-if for a default OpenGL state that you get at startup, just pointing this out.. don't know if thats the state you want CEGUI to start out in

3rdparty/CMakeLists.txt
```diff
@@ -144,4 +144,9 @@ if (IRR_BUILD_CEGUI)
 			PARENT_SCOPE
 		)
 	endif()
+
+	install(
+		DIRECTORY ${CEGUI_INCLUDE_PATH}/CEGUI
```
devshgraphicsprogramming: do we need to expose CEGUI in the public headers of your `irr::ext::cegui` ?

devshgraphicsprogramming:
> 
> 
> Current CMake warning:
> CMake Error at cmake/common.cmake:12 (add_executable):
> Target "brdf_explorer" links to item
> "IrrlichtBAW/build/3rdparty/CEGUI/build/install/lib/CEGUIOpenGLRenderer-0_Static.lib
> " which has leading or trailing whitespace. This is now an error according
> to policy CMP0004.
> Call Stack (most recent call first):
> examples_tests/13.BRDF_Explorer/CMakeLists.txt:13 (irr_create_executable_project)
> 
> I have to keep this because of the CMake's nature. The build still works correctly though. In fact, it needs this warning to work.
> 
> For example:
> 
> ```
> set(VAR
>   lib1
>   lib2
> )
> message(STATUS ${VAR}) # will print lib1lib2
> 
> # so I have to use
> set(VAR
>   "lib1 "
>   "lib2 "
> )
> ```

would it work as
```
set(VAR
  "lib1"
  "lib2"
)
```
?

devshgraphicsprogramming:
Last thing, why is Portable File Dialog a submodule in irr::ext?

btw you have permission to put 3rdparty dependencies (even CEGUI) of any extension in the `./3rdparty` folder as this simplifies life when multiple extensions or parts of the library need the same extension.
Just don't install them

examples_tests/13.BRDF_Explorer/BRDF.h
manhnt9: Yes, I had the same idea but was just testing it quickly before refactoring. I often work that way, usually I'll refactor things later.

3rdparty/CMakeLists.txt
manhnt9: yes, because ext will only help you set up. In order to develop GUI, you'll have to use CEGUI API too.

manhnt9:
```
set(VAR
  "lib1"
  "lib2"
)
```
Doesn't work.

I'll relocate the submodules.

manhnt9:
Still doesn't build so I postpone the refactoring.

examples_tests/13.BRDF_Explorer/BRDF.cpp
manhnt9: All other examples have it, could you explain further?

examples_tests/13.BRDF_Explorer/CMakeLists.txt
manhnt9: Will compile ext as lib and link later, when things render on the screen.

examples_tests/13.BRDF_Explorer/CMakeLists.txt
manhnt9: updated.

examples_tests/13.BRDF_Explorer/main.cpp
manhnt9: refactored

examples_tests/13.BRDF_Explorer/CMakeLists.txt
manhnt9: Added to task list in #279

examples_tests/13.BRDF_Explorer/BRDF.h
manhnt9: Refactored.

ext/CEGUI/CEGUI.cpp
manhnt9: Added to task list, will review this when things render.

ext/CEGUI/CEGUI.cpp
manhnt9: Added to task list, will review this when things render.

manhnt9:
Well, GitLab tracks summary of unsolved comments so you can check without having to scroll, just saying.

examples_tests/13.BRDF_Explorer/BRDF.cpp
devshgraphicsprogramming: examples have `using namespace irr` they don't put random code in the namespace itself

manhnt9: OK, like only put `using namespace` in example having consistent code context.

manhnt9:
About the clip control.
Currently not calling `resetOpenGLClip()` because I call `setOpenGLClip()` after `saveOpenGLState()` so `restoreOpenGLState()` should also restore the clip.

manhnt9:
RFC about the default texture slot, looks like it's from Windows.
Will we be able to use this?

DefaultEmpty

devshgraphicsprogramming:
> RFC about the default texture slot, looks like it's from Windows.
> Will we be able to use this?

I don't understand ? clarify.

manhnt9:
The icon seems like it is designed for Windows. Would you like to change to something else?

manhnt9:
Like a free for commercial use icon, maybe.

devshgraphicsprogramming:
> Like a free for commercial use icon, maybe.

I'm open to any icon that has a license that lets us use it.

examples_tests/13.BRDF_Explorer/main.cpp
```diff
@@ -0,0 +1,66 @@
...
+    uint64_t lastFPSTime = 0;
+
+    while(device->run())
+    if (device->isWindowActive())
+    {
+        driver->beginScene(true, false, video::SColor(255,255,0,0) ); //this gets 11k FPS
+        brdfExplorerApp->renderGUI();
+        driver->endScene();
+
+        // display frames per second in window title
```
devshgraphicsprogramming: remove this, this drags down performance... we should display the Frame time in Miliseconds as measured by an OpenGL timer query with freetype font by CEGUI

manhnt9:
CEGUI ext is now a static library, @devshgraphicsprogramming let's review.
Only texture loader and minor refactoring left then we'll be able to do the 3D rendering.

examples_tests/13.BRDF_Explorer/main.cpp
manhnt9: removed.

manhnt9:
Replaced image icon to FontAwesome's icon.
DefaultEmpty
@Florastamine be alert about this issue, we won't just take any icon to use in our work.

Screenshot_2019-05-16_11-14-21

manhnt9:
I'm having this on Windows (release mode).

Capture

manhnt9:
MSVC can't mix Debug/Release libs (can't build Debug example with Release CEGUI libs). So I'm changing CEGUI build process to use `add_subdirectory`.

3rdparty/CMakeLists.txt
```diff
@@ -117,31 +117,44 @@ if (IRR_BUILD_CEGUI)
...
 	if (MSVC)
 		set(CEGUI_LIBS
-			"${CEGUI_INSTALL_PATH}/lib/CEGUIBase-0_Static.lib "
-			"${CEGUI_INSTALL_PATH}/lib/CEGUICommonDialogs-0_Static.lib "
-			"${CEGUI_INSTALL_PATH}/lib/CEGUICoreWindowRendererSet_Static.lib "
-			"${CEGUI_INSTALL_PATH}/lib/CEGUIExpatParser_Static.lib "
-			"${CEGUI_INSTALL_PATH}/lib/CEGUIOpenGLRenderer-0_Static.lib "
-			"${CEGUI_DEPS_INSTALL_PATH}/lib/Expat.lib "
-			"${CEGUI_DEPS_INSTALL_PATH}/lib/Freetype.lib "
+			${CEGUI_INSTALL_PATH}/lib/CEGUIOpenGLRenderer-0_Static.lib
```
devshgraphicsprogramming: not quoting anymore?

3rdparty/CMakeLists.txt
```diff
@@ -117,31 +117,44 @@ if (IRR_BUILD_CEGUI)
...
+			${CEGUI_INSTALL_PATH}/lib/libCEGUICommonDialogs-0_Static.a
+			${CEGUI_DEPS_INSTALL_PATH}/lib/static/libexpat.a
+			${CEGUI_DEPS_INSTALL_PATH}/lib/static/libpcre.a
 			PARENT_SCOPE
 		)
 	endif()
 
+	install(
+		DIRECTORY ${CEGUI_DEPS_INSTALL_PATH}/include
+		DESTINATION ./include/irr/ext/GUI
```
devshgraphicsprogramming: is the ext called `irr::ext::GUI` or `irr::ext::CEGUI` ?

3rdparty/CMakeLists.txt
```diff
@@ -152,4 +165,14 @@ if (IRR_BUILD_CEGUI)
 		DESTINATION ./debug/include/irr/ext/GUI
 		CONFIGURATIONS Debug
 	)
+	install(
+		FILES ${CEGUI_LIBS}
+		DESTINATION ./lib
```
devshgraphicsprogramming: what does this do?

3rdparty/CMakeLists.txt
```diff
@@ -117,31 +117,44 @@ if (IRR_BUILD_CEGUI)
...
+			${CEGUI_INSTALL_PATH}/lib/libCEGUICoreWindowRendererSet_Static.a
+			${CEGUI_INSTALL_PATH}/lib/libCEGUICommonDialogs-0_Static.a
+			${CEGUI_DEPS_INSTALL_PATH}/lib/static/libexpat.a
+			${CEGUI_DEPS_INSTALL_PATH}/lib/static/libpcre.a
 			PARENT_SCOPE
 		)
 	endif()
 
+	install(
+		DIRECTORY ${CEGUI_DEPS_INSTALL_PATH}/include
```
devshgraphicsprogramming: what does this do?

cmake/common.cmake
```diff
@@ -48,6 +48,59 @@ macro(irr_create_executable_project _EXTRA_SOURCES _EXTRA_OPTIONS _EXTRA_INCLUDE
...
+	target_compile_options(${EXECUTABLE_NAME} PUBLIC ${_EXTRA_OPTIONS})
+
+	if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
+		add_compile_options(
+			"$<$<CONFIG:DEBUG>:-fstack-protector-all>"
+		)
+
+		set(COMMON_LINKER_OPTIONS "-msse4.2 -mfpmath=sse -fuse-ld=gold")
+		set(CMAKE_EXE_LINKER_FLAGS_RELEASE "${COMMON_LINKER_OPTIONS}")
+		set(CMAKE_EXE_LINKER_FLAGS_DEBUG "${COMMON_LINKER_OPTIONS} -fstack-protector-strong -fsanitize=address")
```
devshgraphicsprogramming, line 77: those are only enabled by a special IRRBAW option when building the main library, because they inject into process and make it impossible to use Renderdoc (the sanitizers in particular)

maybe you could hijack the main irrlibrary compile options (they are separate and different for Relaese and Debug targets)

cmake/common.cmake
```diff
@@ -48,6 +48,59 @@ macro(irr_create_executable_project _EXTRA_SOURCES _EXTRA_OPTIONS _EXTRA_INCLUDE
...
+			PROPERTIES
+			RUNTIME_OUTPUT_DIRECTORY_DEBUG "${PROJECT_SOURCE_DIR}/bin"
+			RUNTIME_OUTPUT_DIRECTORY_RELEASE "${PROJECT_SOURCE_DIR}/bin"
+			VS_DEBUGGER_WORKING_DIRECTORY "${PROJECT_SOURCE_DIR}/bin" # seems like has no effect
+		)
+	endif()
+
+	install(
+		TARGETS ${EXECUTABLE_NAME}
+		DESTINATION ./lib
```
devshgraphicsprogramming: wrong destination, need the install prefix as well

examples_tests/13.BRDF_Explorer/CMakeLists.txt
```diff
@@ -4,17 +4,9 @@ if(NOT RES)
...
-
 irr_create_executable_project(
-	"${BRDF_EXPLORER_SOURCES}"
-	"-DCEGUI_STATIC -DCEGUI_BUILD_STATIC_FACTORY_MODULE"
-	"${CEGUI_INCLUDE_PATH}"
-	"${CEGUI_LIBS}"
+	BRDFExplorerApp.cpp
+  "-DCEGUI_STATIC -DCEGUI_BUILD_STATIC_FACTORY_MODULE"
+	"${IRR_CEGUI_EXT_INCLUDE_DIRS}"
+	IrrCEGUIExt
```
devshgraphicsprogramming: I'd call it IrrExtCEGUI to keep consistency with the namespace `irr::ext::cegui`

ext/CEGUI/CMakeLists.txt
```diff
@@ -0,0 +1,40 @@
...
+  IrrCEGUIExt
+  "${IRR_CEGUI_EXT_SRC}"
+  "-DCEGUI_STATIC -DCEGUI_BUILD_STATIC_FACTORY_MODULE"
+  "${CEGUI_INCLUDE_PATH}"
+  "${CEGUI_LIBS}"
+)
+
+install(
+  FILES ${IRR_CEGUI_EXT_H}
+  DESTINATION ./include/irr/ext/CEGUI
```
devshgraphicsprogramming: install prefix?

devshgraphicsprogramming:
> FontAwesome's

What's the license for that image?

manhnt9:
https://fontawesome.com/license/free

manhnt9:
CEGUI deps are not meant to be built on Linux, so Linux users have to install packages manually.
This is the message when I use add_subdirectory with CEGUI deps:

  This dependency pack is intended for use on Apple OS X and MS Windows only.
  For other systems, such as GNU/Linux, please consult your package manager
  or distribution maintainer to discover how to install the libraries you
  wish to use with CEGUI.  Thanks!

manhnt9:
CEGUI deps must be built before adding cegui as subdir to find those deps.
So I don't have other options than to build CEGUI deps when configuring (running CMake).

manhnt9:
CEGUI's project added.

Capture

manhnt9:
CEGUI deps was downloaded into build folder. In order to reset the build folder easier (just deleting build folder without caring about anything inside), deps is downloaded into 3rdparty/CEGUI folder and ignored in git. Then it's extracted to build folder and will be built there.

manhnt9:
Only this left and we'll have a perfect MSVC build.
Looks strange.

Capture

manhnt9:
Ah typo. It's good now.

manhnt9:
Looks what your macro resolves to. I'm not sure that function instead of macro will make a difference but I always stay away from macro. It's as dangerous as variables in CMake. I'm wasting the whole day to make it link for MSVC.

-- optimizedC:/Users/manh/Desktop/IrrlichtBAW/build/3rdparty/CEGUI/cegui/lib/CEGUIOpenGLRenderer-0_Static.liboptimizedC:/Users/manh/Desktop/IrrlichtBAW/build/3rdparty/CEGUI/cegui/lib/CEGUIBase-0_Static.liboptimizedC:/Users/manh/Desktop/IrrlichtBAW/3rdparty/CEGUI/cegui/dependencies/lib/glew.liboptimizedC:/Users/manh/Desktop/IrrlichtBAW/build/3rdparty/CEGUI/cegui/lib/CEGUIExpatParser_Static.liboptimizedC:/Users/manh/Desktop/IrrlichtBAW/build/3rdparty/CEGUI/cegui/lib/CEGUISTBImageCodec_Static.liboptimizedC:/Users/manh/Desktop/IrrlichtBAW/3rdparty/CEGUI/cegui/dependencies/lib/freetype.liboptimizedC:/Users/manh/Desktop/IrrlichtBAW/build/3rdparty/CEGUI/cegui/lib/CEGUICoreWindowRendererSet_Static.liboptimizedC:/Users/manh/Desktop/IrrlichtBAW/build/3rdparty/CEGUI/cegui/lib/CEGUICommonDialogs-0_Static.liboptimizedC:/Users/manh/Desktop/IrrlichtBAW/3rdparty/CEGUI/cegui/dependencies/lib/libexpat.liboptimizedC:/Users/manh/Desktop/IrrlichtBAW/3rdparty/CEGUI/cegui/dependencies/lib/pcre.libdebugC:/Users/manh/Desktop/IrrlichtBAW/build/3rdparty/CEGUI/cegui/lib/CEGUIOpenGLRenderer-0_Static_d.libdebugC:/Users/manh/Desktop/IrrlichtBAW/build/3rdparty/CEGUI/cegui/lib/CEGUIBase-0_Static_d.libdebugC:/Users/manh/Desktop/IrrlichtBAW/3rdparty/CEGUI/cegui/dependencies/lib/glew_d.libdebugC:/Users/manh/Desktop/IrrlichtBAW/build/3rdparty/CEGUI/cegui/lib/CEGUIExpatParser_Static_d.libdebugC:/Users/manh/Desktop/IrrlichtBAW/build/3rdparty/CEGUI/cegui/lib/CEGUISTBImageCodec_Static_d.libdebugC:/Users/manh/Desktop/IrrlichtBAW/3rdparty/CEGUI/cegui/dependencies/lib/freetype_d.libdebugC:/Users/manh/Desktop/IrrlichtBAW/build/3rdparty/CEGUI/cegui/lib/CEGUICoreWindowRendererSet_Static_d.libdebugC:/Users/manh/Desktop/IrrlichtBAW/build/3rdparty/CEGUI/cegui/lib/CEGUICommonDialogs-0_Static_d.libdebugC:/Users/manh/Desktop/IrrlichtBAW/3rdparty/CEGUI/cegui/dependencies/lib/libexpat_d.libdebugC:/Users/manh/Desktop/IrrlichtBAW/3rdparty/CEGUI/cegui/dependencies/lib/pcre_d.libIrrCEGUIExt

manhnt9:
MSVC will always be considered a disease to me. CMake with GCC is very great. For example if you link IrrCEGUIExt, you have all IrrCEGUIExt as well. Guess what you get with MSVC? A IrrCEGUIExt string.

devshgraphicsprogramming:
> 
> 
> CEGUI deps are not meant to be built on Linux, so Linux users have to install packages manually.
> This is the message when I use add_subdirectory with CEGUI deps:
> 
> This dependency pack is intended for use on Apple OS X and MS Windows only.
> For other systems, such as GNU/Linux, please consult your package manager
> or distribution maintainer to discover how to install the libraries you
> wish to use with CEGUI. Thanks!

That's literally beyond annoying, but it can stay like this for the time being so you're not wasting your time.

3rdparty/CEGUI/CMakeLists.txt
```diff
@@ -1,122 +1,75 @@
...
+set(CEGUI_BUILD_XMLPARSER_RAPIDXML OFF CACHE BOOL "")
+set(CEGUI_BUILD_RENDERER_OPENGL OFF CACHE BOOL "")
+set(CEGUI_SAMPLES_ENABLED OFF CACHE BOOL "")
+
+# Enable required CEGUI options
+set(CEGUI_STRING_CLASS "2" CACHE STRING "")
+set(CEGUI_HAS_FREETYPE ON CACHE BOOL "")
+set(CEGUI_HAS_PCRE_REGEX ON CACHE BOOL "")
+set(CEGUI_BUILD_XMLPARSER_EXPAT ON CACHE BOOL "")
+set(CEGUI_BUILD_IMAGECODEC_STB ON CACHE BOOL "") # needs at least 1 image codec module
```
devshgraphicsprogramming, line 99: eventually after hooking in irrbaw image loaders, this will disappear?

devshgraphicsprogramming:
> CEGUI deps must be built before adding cegui as subdir to find those deps.
> So I don't have other options than to build CEGUI deps when configuring (running CMake).

> CEGUI deps was downloaded into build folder. In order to reset the build folder easier (just deleting build folder without caring about anything inside), deps is downloaded into 3rdparty/CEGUI folder and ignored in git. Then it's extracted to build folder and will be built there.

Sounds like the CEGUI dependency pack is a little bit fishy ;)

I think we've had the "had to be built upon Configure" issue with OpenSSL (and then it turned out that you can actually get the exotic dependency building outside of Configure), @Crisspl how did we resolve it?

examples_tests/13.BRDF_Explorer/CMakeLists.txt
```diff
@@ -4,9 +4,16 @@ if(NOT RES)
 	message(FATAL_ERROR "common.cmake not found. Should be in {repo_root}/cmake directory")
 endif()
 
+add_compile_definitions(CEGUI_STATIC CEGUI_BUILD_STATIC_FACTORY_MODULE)
+
+set(BRDF_EXPLORER_LIB
```
devshgraphicsprogramming: BRDF Explorer will be a lib?

devshgraphicsprogramming:
> 
> 
> Looks what your macro resolves to. I'm not sure that function instead of macro will make a difference but I always stay away from macro. It's as dangerous as variables in CMake. I'm wasting the whole day to make it link for MSVC.
> 
> -- optimizedC:/Users/manh/Desktop/IrrlichtBAW/build/3rdparty/CEGUI/cegui/lib/CEGUIOpenGLRenderer-0_Static.liboptimizedC:/Users/manh/Desktop/IrrlichtBAW/build/3rdparty/CEGUI/cegui/lib/CEGUIBase-0_Static.liboptimizedC:/Users/manh/Desktop/IrrlichtBAW/3rdparty/CEGUI/cegui/dependencies/lib/glew.liboptimizedC:/Users/manh/Desktop/IrrlichtBAW/build/3rdparty/CEGUI/cegui/lib/CEGUIExpatParser_Static.liboptimizedC:/Users/manh/Desktop/IrrlichtBAW/build/3rdparty/CEGUI/cegui/lib/CEGUISTBImageCodec_Static.liboptimizedC:/Users/manh/Desktop/IrrlichtBAW/3rdparty/CEGUI/cegui/dependencies/lib/freetype.liboptimizedC:/Users/manh/Desktop/IrrlichtBAW/build/3rdparty/CEGUI/cegui/lib/CEGUICoreWindowRendererSet_Static.liboptimizedC:/Users/manh/Desktop/IrrlichtBAW/build/3rdparty/CEGUI/cegui/lib/CEGUICommonDialogs-0_Static.liboptimizedC:/Users/manh/Desktop/IrrlichtBAW/3rdparty/CEGUI/cegui/dependencies/lib/libexpat.liboptimizedC:/Users/manh/Desktop/IrrlichtBAW/3rdparty/CEGUI/cegui/dependencies/lib/pcre.libdebugC:/Users/manh/Desktop/IrrlichtBAW/build/3rdparty/CEGUI/cegui/lib/CEGUIOpenGLRenderer-0_Static_d.libdebugC:/Users/manh/Desktop/IrrlichtBAW/build/3rdparty/CEGUI/cegui/lib/CEGUIBase-0_Static_d.libdebugC:/Users/manh/Desktop/IrrlichtBAW/3rdparty/CEGUI/cegui/dependencies/lib/glew_d.libdebugC:/Users/manh/Desktop/IrrlichtBAW/build/3rdparty/CEGUI/cegui/lib/CEGUIExpatParser_Static_d.libdebugC:/Users/manh/Desktop/IrrlichtBAW/build/3rdparty/CEGUI/cegui/lib/CEGUISTBImageCodec_Static_d.libdebugC:/Users/manh/Desktop/IrrlichtBAW/3rdparty/CEGUI/cegui/dependencies/lib/freetype_d.libdebugC:/Users/manh/Desktop/IrrlichtBAW/build/3rdparty/CEGUI/cegui/lib/CEGUICoreWindowRendererSet_Static_d.libdebugC:/Users/manh/Desktop/IrrlichtBAW/build/3rdparty/CEGUI/cegui/lib/CEGUICommonDialogs-0_Static_d.libdebugC:/Users/manh/Desktop/IrrlichtBAW/3rdparty/CEGUI/cegui/dependencies/lib/libexpat_d.libdebugC:/Users/manh/Desktop/IrrlichtBAW/3rdparty/CEGUI/cegui/dependencies/lib/pcre_d.libIrrCEGUIExt

This is still a problem or did your last two commits get around it?

Also what do you propose instead of that macro?

3rdparty/CMakeLists.txt
manhnt9: Yes, not sure why it doesn't work before although I tried different formats including non-quotes and quotes. But right now Windows and Linux are working fine anyways.

3rdparty/CMakeLists.txt
manhnt9: irr::ext::GUI means GUI dependencies used by ext.

3rdparty/CMakeLists.txt
manhnt9: It copies CEGUI's libraries to installation folder for end users to link.

3rdparty/CMakeLists.txt
manhnt9: It copies CEGUI's libraries to installation folder for end users to link.

manhnt9: On Windows, end users need headers of CEGUI's dependencies to compile their app since CEGUI include those too.

cmake/common.cmake
manhnt9: About the install prefix, I replied once before. Maybe you missed the mail (please use mail filter to classify GitHub mail). `install()` knows the folder. Destination is also correct since this is for a library, the ${EXECUTABLE_NAME} is the one which causes misunderstanding, will refactor.

examples_tests/13.BRDF_Explorer/CMakeLists.txt
manhnt9: Changed to IrrExtCEGUI.

3rdparty/CEGUI/CMakeLists.txt
manhnt9: No, it needs to compile because the code must access to the image codec interface.
Unless I patched CEGUI to have a null image codec or embed irrbaw image loader code into CEGUI (not 100% certain if that will be accepted into original CEGUI repo).

examples_tests/13.BRDF_Explorer/CMakeLists.txt
manhnt9: BRDF_EXPLORER_LIB variable stores the library list (which the app needs to link).

manhnt9:
It links properly now. Macro is like find and replace (same as #define in C++) so passed parameters could be messed up in code generation. Also it bloats up the CMake code and increases configuration time a little bit.
I would use CMake's function syntax.

cmake/common.cmake
manhnt9: Could you tell me the involved line numbers?

3rdparty/CMakeLists.txt
devshgraphicsprogramming: https://github.com/manhnt9/IrrlichtBAW/tree/sRGB/ext

A. There's no GUI folder here. 
B. With different GUI system out there, we cannot just use `GUI` as a name

devshgraphicsprogramming: The only folders that should be in `./include` should match namespace names 100%

manhnt9: Mis-typed, there's no namespace like irr::ext::GUI, only a GUI folder.

devshgraphicsprogramming: i.e. if there's a `./include/irr/ext/VisibilityBuffer` then there is a `irr::ext::VisiblityBuffer`

manhnt9: understood, see my other comment to solve the actual problem about this.

manhnt9:
I'm also giving irrbaw an ext framework, including:
- A macro to create a static ext library by providing header, source list, external include path and compile options
- That macro will build, install headers, install ext lib to correct location according to `Extension Rules`
- And set global CMake variables for any example to use, e.g: IRR_EXT_CEGUI_INCLUDE_DIRS and IRR_EXT_CEGUI_LIB

To use the ext framework:
- Refer to ext/CMakeLists.txt, then add_subdirectory and copy 2 simple variables
- In ext's CMake code, call irr_create_ext_library_project
- Now you can use IRR_EXT_NAME_INCLUDE_DIRS and IRR_EXT_NAME_LIB

manhnt9:
CMake code is now super clean.

manhnt9:
Discovered and fixed another issue affecting the bad boy MSVC.
Also reported to CMake [repo](https://gitlab.kitware.com/cmake/cmake/issues/19278).

manhnt9:
Commit date is messed again because of Linux/Windows time mismatch.

ghost:
> Commit date is messed again because of Linux/Windows time mismatch.

You can configure Linux to use local time as the stored time format
```
timedatectl set-local-rtc 1 --adjust-system-clock
```

Linux and Windows use different conventions when writing current time/date to BIOS upon booting, hence the mistmatch

manhnt9:
I was lazy to find the solution although I've done before. Used your above command, Thanks!

devshgraphicsprogramming:
Hi, I met with our contact, tried to show him the example and ran into this when doing a fresh clone

```
Cloning into 'C:/work/IrrlichtBAW/manhnt/3rdparty/CEGUI/cegui'...
Host key verification failed.
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
fatal: clone of 'git@gitlab.com:InnerPieceOSS/cegui.git' into submodule path 'C:/work/IrrlichtBAW/manhnt/3rdparty/CEGUI/cegui' failed
Failed to clone '3rdparty/CEGUI/cegui'. Retry scheduled

Cloning into 'C:/work/IrrlichtBAW/manhnt/3rdparty/portable-file-dialogs'...
Host key verification failed.
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
fatal: clone of 'git@github.com:samhocevar/portable-file-dialogs.git' into submodule path 'C:/work/IrrlichtBAW/manhnt/3rdparty/portable-file-dialogs' failed
Failed to clone '3rdparty/portable-file-dialogs'. Retry scheduled

Cloning into 'C:/work/IrrlichtBAW/manhnt/examples_tests/media/cegui_alfisko'...
Host key verification failed.
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
fatal: clone of 'git@gitlab.com:InnerPieceOSS/cegui_alfisko.git' into submodule path 'C:/work/IrrlichtBAW/manhnt/examples_tests/media/cegui_alfisko' failed
Failed to clone 'examples_tests/media/cegui_alfisko'. Retry scheduled

Cloning into 'C:/work/IrrlichtBAW/manhnt/ext/CEGUI/stb'...
Host key verification failed.
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
fatal: clone of 'git@github.com:nothings/stb.git' into submodule path 'C:/work/IrrlichtBAW/manhnt/ext/CEGUI/stb' failed
Failed to clone 'ext/CEGUI/stb'. Retry scheduled

Cloning into 'C:/work/IrrlichtBAW/manhnt/3rdparty/CEGUI/cegui'...
Host key verification failed.
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
fatal: clone of 'git@gitlab.com:InnerPieceOSS/cegui.git' into submodule path 'C:/work/IrrlichtBAW/manhnt/3rdparty/CEGUI/cegui' failed
Failed to clone '3rdparty/CEGUI/cegui' a second time, aborting

CMake Error at 3rdparty/CMakeLists.txt:7 (add_subdirectory):
  The source directory

    C:/work/IrrlichtBAW/manhnt/3rdparty/SPIRV-Headers

  does not contain a CMakeLists.txt file.


CMake Error at 3rdparty/CMakeLists.txt:11 (add_subdirectory):
  The source directory

    C:/work/IrrlichtBAW/manhnt/3rdparty/SPIRV-Tools

  does not contain a CMakeLists.txt file.


CMake Error at 3rdparty/CMakeLists.txt:16 (add_subdirectory):
  The source directory

    C:/work/IrrlichtBAW/manhnt/3rdparty/glslang

  does not contain a CMakeLists.txt file.


CMake Error at 3rdparty/CMakeLists.txt:26 (add_subdirectory):
  The source directory

    C:/work/IrrlichtBAW/manhnt/3rdparty/shaderc

  does not contain a CMakeLists.txt file.


CMake Error at 3rdparty/CMakeLists.txt:36 (add_subdirectory):
  The source directory

    C:/work/IrrlichtBAW/manhnt/3rdparty/libjpeg

  does not contain a CMakeLists.txt file.


CMake Error at 3rdparty/CMakeLists.txt:45 (add_subdirectory):
  The source directory

    C:/work/IrrlichtBAW/manhnt/3rdparty/libpng

  does not contain a CMakeLists.txt file.
```

manhnt9:
Network problems, @Crisspl met these too. Were you using coffee's internet?

manhnt9:
BTW you could ask for a pre-built version. Are you still there or not?

manhnt9:
Also did you try cloning your original repo too?

manhnt9:
Ah I discovered something, we should use HTTPS for those modules. I was using SSH url because I always use SSH. Sorry, this is very unexpected. Will update.

devshgraphicsprogramming:
when can I pull and try again?

manhnt9:
in several mins

manhnt9:
If you want to save build time I can send a prebuilt version.

manhnt9:
I pushed, try git reset --hard and pull, then git submodule sync

Crisspl:
Not sure if it's the same problem but in my case it started running when i created gitlab account, created ssh key and registered it in gitlab account

manhnt9:
I think that's it. The only odd thing I've found out is that the failed submodules were using SSH URLs.

manhnt9:
Shit, I was hurrying and sleepy, pasted STB URL for CEGUI module.

devshgraphicsprogramming:
Having CMake `Configure` build your CEGUI dependencies is a nightmare (takes so long one thinks its broken), we should fix it at some point in the future (before 2020)

I'm getting linker errors, the glew dependency
```
38>LINK : warning LNK4098: defaultlib 'MSVCRT' conflicts with use of other libs; use /NODEFAULTLIB:library
38>CEGUIOpenGLRenderer-0_Static.lib(GL3Renderer.obj) : error LNK2001: unresolved external symbol __imp___glewUseProgram
38>CEGUIOpenGLRenderer-0_Static.lib(GL3Renderer.obj) : error LNK2001: unresolved external symbol __imp___glewUniform1i
38>CEGUIOpenGLRenderer-0_Static.lib(GL3Renderer.obj) : error LNK2001: unresolved external symbol __imp___glewBlendFuncSeparate
38>CEGUIOpenGLRenderer-0_Static.lib(GL3Renderer.obj) : error LNK2001: unresolved external symbol __imp___glewActiveTexture
38>CEGUIOpenGLRenderer-0_Static.lib(RendererBase.obj) : error LNK2001: unresolved external symbol __imp_glewGetErrorString
38>CEGUIOpenGLRenderer-0_Static.lib(RendererBase.obj) : error LNK2001: unresolved external symbol __imp_glewExperimental
38>CEGUIOpenGLRenderer-0_Static.lib(RendererBase.obj) : error LNK2001: unresolved external symbol __imp_glewInit
38>CEGUIOpenGLRenderer-0_Static.lib(GL3FBOTextureTarget.obj) : error LNK2001: unresolved external symbol __imp___glewFramebufferTexture2D
38>CEGUIOpenGLRenderer-0_Static.lib(GL3FBOTextureTarget.obj) : error LNK2001: unresolved external symbol __imp___glewCheckFramebufferStatus
38>CEGUIOpenGLRenderer-0_Static.lib(GL3FBOTextureTarget.obj) : error LNK2001: unresolved external symbol __imp___glewDeleteFramebuffers
38>CEGUIOpenGLRenderer-0_Static.lib(GL3FBOTextureTarget.obj) : error LNK2001: unresolved external symbol __imp___glewGenFramebuffers
38>CEGUIOpenGLRenderer-0_Static.lib(GL3FBOTextureTarget.obj) : error LNK2001: unresolved external symbol __imp___glewBindFramebuffer
38>CEGUIOpenGLRenderer-0_Static.lib(Shader.obj) : error LNK2001: unresolved external symbol __imp___glewGetShaderiv
38>CEGUIOpenGLRenderer-0_Static.lib(Shader.obj) : error LNK2001: unresolved external symbol __imp___glewCompileShader
38>CEGUIOpenGLRenderer-0_Static.lib(Shader.obj) : error LNK2001: unresolved external symbol __imp___glewGetAttribLocation
38>CEGUIOpenGLRenderer-0_Static.lib(Shader.obj) : error LNK2001: unresolved external symbol __imp___glewCreateShader
38>CEGUIOpenGLRenderer-0_Static.lib(Shader.obj) : error LNK2001: unresolved external symbol __imp___glewCreateProgram
38>CEGUIOpenGLRenderer-0_Static.lib(Shader.obj) : error LNK2001: unresolved external symbol __imp___glewGetShaderInfoLog
38>CEGUIOpenGLRenderer-0_Static.lib(Shader.obj) : error LNK2001: unresolved external symbol __imp___glewGetProgramiv
38>CEGUIOpenGLRenderer-0_Static.lib(Shader.obj) : error LNK2001: unresolved external symbol __imp___glewBindFragDataLocation
38>CEGUIOpenGLRenderer-0_Static.lib(Shader.obj) : error LNK2001: unresolved external symbol __imp___glewGetUniformLocation
38>CEGUIOpenGLRenderer-0_Static.lib(Shader.obj) : error LNK2001: unresolved external symbol __imp___glewAttachShader
38>CEGUIOpenGLRenderer-0_Static.lib(Shader.obj) : error LNK2001: unresolved external symbol __imp___glewDeleteProgram
38>CEGUIOpenGLRenderer-0_Static.lib(Shader.obj) : error LNK2001: unresolved external symbol __imp___glewShaderSource
38>CEGUIOpenGLRenderer-0_Static.lib(Shader.obj) : error LNK2001: unresolved external symbol __imp___glewGetProgramInfoLog
38>CEGUIOpenGLRenderer-0_Static.lib(Shader.obj) : error LNK2001: unresolved external symbol __imp___glewDeleteShader
38>CEGUIOpenGLRenderer-0_Static.lib(Shader.obj) : error LNK2001: unresolved external symbol __imp___glewLinkProgram
38>CEGUIOpenGLRenderer-0_Static.lib(StateChangeWrapper.obj) : error LNK2001: unresolved external symbol __imp___glewBindVertexArray
38>CEGUIOpenGLRenderer-0_Static.lib(StateChangeWrapper.obj) : error LNK2001: unresolved external symbol __imp___glewBindBuffer
38>CEGUIOpenGLRenderer-0_Static.lib(GL.obj) : error LNK2001: unresolved external symbol __imp___GLEW_VERSION_2_0
38>CEGUIOpenGLRenderer-0_Static.lib(GL.obj) : error LNK2001: unresolved external symbol __imp___GLEW_VERSION_3_1
38>CEGUIOpenGLRenderer-0_Static.lib(GL.obj) : error LNK2001: unresolved external symbol __imp___glewGetStringi
38>CEGUIOpenGLRenderer-0_Static.lib(GL.obj) : error LNK2001: unresolved external symbol __imp___GLEW_VERSION_3_2
38>CEGUIOpenGLRenderer-0_Static.lib(GL.obj) : error LNK2001: unresolved external symbol __imp___GLEW_VERSION_1_3
38>CEGUIOpenGLRenderer-0_Static.lib(GL.obj) : error LNK2001: unresolved external symbol __imp___GLEW_ARB_texture_non_power_of_two
38>CEGUIOpenGLRenderer-0_Static.lib(GL3GeometryBuffer.obj) : error LNK2001: unresolved external symbol __imp___glewEnableVertexAttribArray
38>CEGUIOpenGLRenderer-0_Static.lib(GL3GeometryBuffer.obj) : error LNK2001: unresolved external symbol __imp___glewVertexAttribPointer
38>CEGUIOpenGLRenderer-0_Static.lib(GL3GeometryBuffer.obj) : error LNK2001: unresolved external symbol __imp___glewBufferSubData
38>CEGUIOpenGLRenderer-0_Static.lib(GL3GeometryBuffer.obj) : error LNK2001: unresolved external symbol __imp___glewDeleteVertexArrays
38>CEGUIOpenGLRenderer-0_Static.lib(GL3GeometryBuffer.obj) : error LNK2001: unresolved external symbol __imp___glewBufferData
38>CEGUIOpenGLRenderer-0_Static.lib(GL3GeometryBuffer.obj) : error LNK2001: unresolved external symbol __imp___glewGenVertexArrays
38>CEGUIOpenGLRenderer-0_Static.lib(GL3GeometryBuffer.obj) : error LNK2001: unresolved external symbol __imp___glewUniformMatrix4fv
38>CEGUIOpenGLRenderer-0_Static.lib(GL3GeometryBuffer.obj) : error LNK2001: unresolved external symbol __imp___glewGenBuffers
38>CEGUIOpenGLRenderer-0_Static.lib(GL3GeometryBuffer.obj) : error LNK2001: unresolved external symbol __imp___glewDeleteBuffers
38>CEGUIOpenGLRenderer-0_Static.lib(Texture.obj) : error LNK2001: unresolved external symbol __imp___glewGetCompressedTexImage
38>CEGUIOpenGLRenderer-0_Static.lib(Texture.obj) : error LNK2001: unresolved external symbol __imp___glewCompressedTexImage2D
38>CEGUIOpenGLRenderer-0_Static.lib(Texture.obj) : error LNK2001: unresolved external symbol __imp___glewCompressedTexSubImage2D
```

manhnt9:
Is this the latest commit?

devshgraphicsprogramming:
yes, after hard reset, pull etc.

manhnt9:
What is your CEGUI module's latest commit?

manhnt9:
About the slowness, I don't thing it's gonna improve much when using git modules, because:
- Downloading a zip file is faster than each uncompressed submodule
- Build time of dependencies and cegui won't improve

devshgraphicsprogramming:
Build time is fine, its just the CMake "Configure" stage thats killing me

manhnt9:
Did the linker errors go away? I think it's because you wasn't having latest CEGUI patch for static linking.

devshgraphicsprogramming:
if we need a later version of CEGUI can you commit that you've pulled a later version of the submodule?
(if you pull a different submodule commit/branch, you can commit that fact)

manhnt9:
Yes I know that, just wondering if you've got the correct CEGUI version. Because it (glew linking errors) is strange since that is fixed in latest CEGUI.

devshgraphicsprogramming:
well I reset and force intialized my submodules, what else can i do?

manhnt9:
Could you paste last 5 commits of git log after cd into `3rdparty/CEGUI/cegui`?

devshgraphicsprogramming:
```
Revision: fcc8d6cdc1992ba532db865ec56056e930850cc4
Author: ManhNT <manhnt.9@outlook.com>
Date: 20/05/2019 11:35:35
Message:
Update CEGUI for static PCRE fix

----
Modified: 3rdparty/CEGUI/cegui

Revision: 7b86b751cec96158bf9214f3e49ab9467fd63dd4
Author: ManhNT <manhnt.9@outlook.com>
Date: 20/05/2019 11:04:05
Message:
Update patched CEGUI

----
Modified: 3rdparty/CEGUI/CMakeLists.txt
Modified: 3rdparty/CEGUI/cegui

Revision: 3d5d0224eb7aaacf88d7f931325f0a13615b7675
Author: Manh NT <manhnt.9@outlook.com>
Date: 02/05/2019 22:17:39
Message:
Update CEGUI submodule to ignore dependencies folder

----
Modified: 3rdparty/CEGUI/cegui

Revision: 272422997a98184f39a0945279f9c95d2deec785
Author: ManhNT <manhnt.9@outlook.com>
Date: 01/05/2019 11:34:00
Message:
Add submodules

----
Modified: .gitmodules
Added: 3rdparty/CEGUI/cegui
Added: examples_tests/media/cegui_alfisko

```

manhnt9:
Yeah that's an old module. My log is below, notice that the first commit hash is the same as what I pushed: https://github.com/manhnt9/IrrlichtBAW/tree/sRGB/3rdparty/CEGUI
```
commit 160161abafdafe2b033b1d84c3df10eb6f45cbda (HEAD -> saga, origin/saga, origin/HEAD)
Author: ManhNT <manhnt.9@outlook.com>
Date:   Mon May 20 16:33:55 2019 +0700

    Add utility files

commit 1795c87bbb250c9b10322a0f10f3bfb731cc3f68
Author: Manh NT <manhnt.9@outlook.com>
Date:   Mon May 20 16:31:38 2019 +0700

    Add PCRE to static dependencies config

commit 8c1906537df085f46ac962856cd8ef881a470ae8
Author: Manh NT <manhnt.9@outlook.com>
Date:   Mon May 20 14:54:58 2019 +0700

    Initial static dependencies patch

commit a19029730a24ceb7ff1eec522d7b18b1bddc1fd2
Author: Manh NT <manhnt.9@outlook.com>
Date:   Wed Apr 17 10:59:45 2019 +0700

    Revert missing patch

commit 18802fcdbbaa864d8d9bd95c8ff420100d471854
Author: Manh NT <manhnt.9@outlook.com>
Date:   Wed Apr 17 10:55:59 2019 +0700

    Revert incompatible change
```

manhnt9:
Oooh, strange, but the latest commit message should be `Add utility files` instead of `Update CEGUI for static PCRE fix` shown in https://github.com/manhnt9/IrrlichtBAW/tree/sRGB/3rdparty/CEGUI.

I think Git is messed up somehow. I'm checking.

manhnt9:
Solved, besides Git URL a submodule needs the branch name if it's not using master as default branch. My CEGUI branch is `saga`. Adding submodule without branch name works for the first time but it can't update without knowing which branch later. Git is dumb in this issue, not producing a sensible output and just assume the updated status somehow.

manhnt9:
You can pull and try updating the CEGUI module.

devshgraphicsprogramming:
I couldn't pull because despite the pull of the main repo (where the branch is specified) git told me I'm not on a branch when pulling a submodule

So I had to pull explicitly, and it told me I'm up to date
```
C:\work\IrrlichtBAW\manhnt\3rdparty\CEGUI\cegui>git pull origin saga
From https://gitlab.com/InnerPieceOSS/cegui
 * branch              saga       -> FETCH_HEAD
Already up to date.

C:\work\IrrlichtBAW\manhnt\3rdparty\CEGUI\cegui>
```

Why do I need to mess around with the module manually if you update the submodules file?
Its not like I had any tracked or untrack file changes there.


And after all this I get extra compiler errors + cannot find glew.h .

manhnt9:
For any submodule, you'll have to checkout its branch before pulling.
```
git checkout saga (or master for other modules)
git pull origin saga
```

devshgraphicsprogramming:
I've done that manually, question is why on earth do I have to do that by myself, shouldn't we all be synced up nicely regarding what branch and revision/commit of a submodule we're using?

Especially after a `hard` `reset`?

devshgraphicsprogramming:
I was all up to date and proper.

Seems that because the CEGUI deps are not proper targets, after a change of cegui the following happens:
1) `ZERO_CHECK` doesn't work, as in it doesn't detect dependency
2) "Configure" again doesn't work because of CMake cache

So I needed to delete cache, and do "Configure" with correct extra options from the ground up again.

manhnt9:
The submodule hassle is due to the git mirroring.
The git version has extra changes (.gitignore for example) which are not reasonable to put into original svn repo. And the sync process will break completely if the git repo has manually made commit via git.
So each time I sync, my sync script does some tricks that re-writing some git commits, I'd have to revert git changes, sync from svn, re-add git changes and commit, and push -f. Maybe the commit hashes are messed up.
For now, I can eliminate git changes from the repo (leaving out .gitignore and then dependencies folder generated by build will be detected by irrbaw's git, but we already have shaderc and openssl polluted the same way, shaderc needs a touch of new lib file before building and openssl only builds in the source folder).

About the CEGUI CMake cache, yes because CEGUI version has changed. I advise deleting the CEGUI folder generated in build/3rdparty and configure again. Also note that I've tried deleting my build folder and rebuild several times on both Windows and Linux, so it shouldn't meet any serious build problems now.

devshgraphicsprogramming:
Actually OpenSSL building in the source dir is not a problem, because the submodule gets reset and pulled the latest version from the 1.0.2 branch every time it builds ;)

Yeah the shaderc thing is a drag, they were supposed to have already fixed it :(

I think CEGUI deps should be properly added and statically linked on Linux after we ship BRDF Explorer completely.

devshgraphicsprogramming:
Update: I've experienced #299 when trying to build example 13 in the Release configuration straight away.

Nope still same glew linker errors

devshgraphicsprogramming:
Even in Debug build
```
12>CEGUIOpenGLRenderer-0_Static_d.lib(RendererBase.obj) : error LNK2019: unresolved external symbol __imp_glewInit referenced in function "protected: void __cdecl CEGUI::OpenGLRendererBase::init(bool,bool)" (?init@OpenGLRendererBase@CEGUI@@IEAAX_N0@Z)
12>CEGUIOpenGLRenderer-0_Static_d.lib(RendererBase.obj) : error LNK2019: unresolved external symbol __imp_glewGetErrorString referenced in function "protected: void __cdecl CEGUI::OpenGLRendererBase::init(bool,bool)" (?init@OpenGLRendererBase@CEGUI@@IEAAX_N0@Z)
12>CEGUIOpenGLRenderer-0_Static_d.lib(RendererBase.obj) : error LNK2001: unresolved external symbol __imp_glewExperimental
12>CEGUIOpenGLRenderer-0_Static_d.lib(GL3Renderer.obj) : error LNK2001: unresolved external symbol __imp___glewActiveTexture
12>CEGUIOpenGLRenderer-0_Static_d.lib(GL3Renderer.obj) : error LNK2001: unresolved external symbol __imp___glewBlendFuncSeparate
12>CEGUIOpenGLRenderer-0_Static_d.lib(StateChangeWrapper.obj) : error LNK2001: unresolved external symbol __imp___glewBlendFuncSeparate
12>CEGUIOpenGLRenderer-0_Static_d.lib(GL3Renderer.obj) : error LNK2001: unresolved external symbol __imp___glewUniform1i
12>CEGUIOpenGLRenderer-0_Static_d.lib(GL3Renderer.obj) : error LNK2001: unresolved external symbol __imp___glewUseProgram
12>CEGUIOpenGLRenderer-0_Static_d.lib(Shader.obj) : error LNK2001: unresolved external symbol __imp___glewUseProgram
12>CEGUIOpenGLRenderer-0_Static_d.lib(GL.obj) : error LNK2001: unresolved external symbol __imp___glewGetStringi
12>CEGUIOpenGLRenderer-0_Static_d.lib(GL.obj) : error LNK2001: unresolved external symbol __imp___GLEW_VERSION_1_3
12>CEGUIOpenGLRenderer-0_Static_d.lib(GL.obj) : error LNK2001: unresolved external symbol __imp___GLEW_VERSION_2_0
12>CEGUIOpenGLRenderer-0_Static_d.lib(GL.obj) : error LNK2001: unresolved external symbol __imp___GLEW_VERSION_3_1
12>CEGUIOpenGLRenderer-0_Static_d.lib(GL.obj) : error LNK2001: unresolved external symbol __imp___GLEW_VERSION_3_2
12>CEGUIOpenGLRenderer-0_Static_d.lib(GL.obj) : error LNK2001: unresolved external symbol __imp___GLEW_ARB_texture_non_power_of_two
12>CEGUIOpenGLRenderer-0_Static_d.lib(Texture.obj) : error LNK2001: unresolved external symbol __imp___glewCompressedTexImage2D
12>CEGUIOpenGLRenderer-0_Static_d.lib(Texture.obj) : error LNK2001: unresolved external symbol __imp___glewCompressedTexSubImage2D
12>CEGUIOpenGLRenderer-0_Static_d.lib(Texture.obj) : error LNK2001: unresolved external symbol __imp___glewGetCompressedTexImage
12>CEGUIOpenGLRenderer-0_Static_d.lib(Texture.obj) : error LNK2001: unresolved external symbol __imp___glewBindBuffer
12>CEGUIOpenGLRenderer-0_Static_d.lib(GL3GeometryBuffer.obj) : error LNK2001: unresolved external symbol __imp___glewBindBuffer
12>CEGUIOpenGLRenderer-0_Static_d.lib(StateChangeWrapper.obj) : error LNK2001: unresolved external symbol __imp___glewBindBuffer
12>CEGUIOpenGLRenderer-0_Static_d.lib(Texture.obj) : error LNK2001: unresolved external symbol __imp___glewBindFramebuffer
12>CEGUIOpenGLRenderer-0_Static_d.lib(GL3FBOTextureTarget.obj) : error LNK2001: unresolved external symbol __imp___glewBindFramebuffer
12>CEGUIOpenGLRenderer-0_Static_d.lib(Texture.obj) : error LNK2001: unresolved external symbol __imp___glewDeleteFramebuffers
12>CEGUIOpenGLRenderer-0_Static_d.lib(GL3FBOTextureTarget.obj) : error LNK2001: unresolved external symbol __imp___glewDeleteFramebuffers
12>CEGUIOpenGLRenderer-0_Static_d.lib(Texture.obj) : error LNK2001: unresolved external symbol __imp___glewFramebufferTexture2D
12>CEGUIOpenGLRenderer-0_Static_d.lib(GL3FBOTextureTarget.obj) : error LNK2001: unresolved external symbol __imp___glewFramebufferTexture2D
12>CEGUIOpenGLRenderer-0_Static_d.lib(Texture.obj) : error LNK2001: unresolved external symbol __imp___glewGenFramebuffers
12>CEGUIOpenGLRenderer-0_Static_d.lib(GL3FBOTextureTarget.obj) : error LNK2001: unresolved external symbol __imp___glewGenFramebuffers
12>CEGUIOpenGLRenderer-0_Static_d.lib(Shader.obj) : error LNK2001: unresolved external symbol __imp___glewAttachShader
12>CEGUIOpenGLRenderer-0_Static_d.lib(Shader.obj) : error LNK2001: unresolved external symbol __imp___glewCompileShader
12>CEGUIOpenGLRenderer-0_Static_d.lib(Shader.obj) : error LNK2001: unresolved external symbol __imp___glewCreateProgram
12>CEGUIOpenGLRenderer-0_Static_d.lib(Shader.obj) : error LNK2001: unresolved external symbol __imp___glewCreateShader
12>CEGUIOpenGLRenderer-0_Static_d.lib(Shader.obj) : error LNK2001: unresolved external symbol __imp___glewDeleteProgram
12>CEGUIOpenGLRenderer-0_Static_d.lib(Shader.obj) : error LNK2001: unresolved external symbol __imp___glewDeleteShader
12>CEGUIOpenGLRenderer-0_Static_d.lib(Shader.obj) : error LNK2001: unresolved external symbol __imp___glewGetAttribLocation
12>CEGUIOpenGLRenderer-0_Static_d.lib(Shader.obj) : error LNK2001: unresolved external symbol __imp___glewGetProgramInfoLog
12>CEGUIOpenGLRenderer-0_Static_d.lib(Shader.obj) : error LNK2001: unresolved external symbol __imp___glewGetProgramiv
12>CEGUIOpenGLRenderer-0_Static_d.lib(Shader.obj) : error LNK2001: unresolved external symbol __imp___glewGetShaderInfoLog
12>CEGUIOpenGLRenderer-0_Static_d.lib(Shader.obj) : error LNK2001: unresolved external symbol __imp___glewGetShaderiv
12>CEGUIOpenGLRenderer-0_Static_d.lib(Shader.obj) : error LNK2001: unresolved external symbol __imp___glewGetUniformLocation
12>CEGUIOpenGLRenderer-0_Static_d.lib(Shader.obj) : error LNK2001: unresolved external symbol __imp___glewLinkProgram
12>CEGUIOpenGLRenderer-0_Static_d.lib(Shader.obj) : error LNK2001: unresolved external symbol __imp___glewShaderSource
12>CEGUIOpenGLRenderer-0_Static_d.lib(Shader.obj) : error LNK2001: unresolved external symbol __imp___glewBindFragDataLocation
12>CEGUIOpenGLRenderer-0_Static_d.lib(GL3GeometryBuffer.obj) : error LNK2001: unresolved external symbol __imp___glewBufferData
12>CEGUIOpenGLRenderer-0_Static_d.lib(GL3GeometryBuffer.obj) : error LNK2001: unresolved external symbol __imp___glewBufferSubData
12>CEGUIOpenGLRenderer-0_Static_d.lib(GL3GeometryBuffer.obj) : error LNK2001: unresolved external symbol __imp___glewDeleteBuffers
12>CEGUIOpenGLRenderer-0_Static_d.lib(GL3GeometryBuffer.obj) : error LNK2001: unresolved external symbol __imp___glewGenBuffers
12>CEGUIOpenGLRenderer-0_Static_d.lib(GL3GeometryBuffer.obj) : error LNK2001: unresolved external symbol __imp___glewEnableVertexAttribArray
12>CEGUIOpenGLRenderer-0_Static_d.lib(GL3GeometryBuffer.obj) : error LNK2001: unresolved external symbol __imp___glewUniformMatrix4fv
12>CEGUIOpenGLRenderer-0_Static_d.lib(GL3GeometryBuffer.obj) : error LNK2001: unresolved external symbol __imp___glewVertexAttribPointer
12>CEGUIOpenGLRenderer-0_Static_d.lib(GL3GeometryBuffer.obj) : error LNK2001: unresolved external symbol __imp___glewBindVertexArray
12>CEGUIOpenGLRenderer-0_Static_d.lib(StateChangeWrapper.obj) : error LNK2001: unresolved external symbol __imp___glewBindVertexArray
12>CEGUIOpenGLRenderer-0_Static_d.lib(GL3GeometryBuffer.obj) : error LNK2001: unresolved external symbol __imp___glewDeleteVertexArrays
12>CEGUIOpenGLRenderer-0_Static_d.lib(GL3GeometryBuffer.obj) : error LNK2001: unresolved external symbol __imp___glewGenVertexArrays
12>CEGUIOpenGLRenderer-0_Static_d.lib(GL3FBOTextureTarget.obj) : error LNK2001: unresolved external symbol __imp___glewCheckFramebufferStatus
```

Are you actually getting your glew from vcpkg or something and not the repo?
Otherwise I have no idea how you manage to build this

3rdparty/CEGUI/CMakeLists.txt
```diff
@@ -0,0 +1,109 @@
...
+                -DCEGUI_BUILD_FREEIMAGE=OFF
+                -DCEGUI_BUILD_GLFW=OFF
+                -DCEGUI_BUILD_LUA=OFF
+                -DCEGUI_BUILD_MINIZIP=OFF
+                -DCEGUI_BUILD_SILLY=OFF
+                -DCEGUI_BUILD_TINYXML=OFF
+                -DCEGUI_BUILD_TOLUAPP=OFF
+                -DCEGUI_BUILD_XERCES=OFF
+                -DCEGUI_BUILD_ZLIB=OFF
+                -DCEGUI_BUILD_EXPAT=ON
```
devshgraphicsprogramming, line 48: see #308

devshgraphicsprogramming:
I think I found the issue, you're including the dependencies like this
```
..\..\..\3rdparty\CEGUI\cegui\dependencies\lib\XXXXX
```

When in-fact they are
```
..\..\3rdparty\CEGUI\cegui-cegui-dependencies-0ecdf3a9e49b\build\dependencies\lib\static\XXXXX
```

devshgraphicsprogramming:
Damn it, that's not it

devshgraphicsprogramming:
This might be definitely it
```
if (CEGUI_BUILD_SHARED_LIBS_WITH_STATIC_DEPENDENCIES AND CEGUI_USE_GLEW)
    add_definitions( -DGLEW_STATIC )
endif()
```

sources:
https://github.com/nigels-com/glew/issues/161
https://github.com/nigels-com/glew/issues/88
https://stackoverflow.com/questions/22213558/unresolved-external-symbol-imp-glewinit-vs-imp-glewinit0

devshgraphicsprogramming:
> This might be definitely it

Now it works, CEGUI needs `CEGUI_BUILD_SHARED_LIBS_WITH_STATIC_DEPENDENCIES = ON`

manhnt9:
CEGUI_BUILD_SHARED_LIBS_WITH_STATIC_DEPENDENCIES
is not intended. It's for dynamic libs with static dependencies. That's why I needed to patch CEGUI for static libs + static dependencies. Mine works without that option, I'm not able to dig into your case soon but in general that fix is discouraged.

devshgraphicsprogramming:
>  It's for dynamic libs with static dependencies. 

Seems not, also the only thing it seems to control is whether to define `GLEW_STATIC`

> Mine works without that option

No idea how thats possible, maybe you're actually loading glew from a system installed dll.
`GLEW_STATIC` must be defined when compiling anything that used glew headers for a static glew to work.

cmake/common.cmake
devshgraphicsprogramming: we actually have these beautiful options https://github.com/buildaworldnet/IrrlichtBAW/blob/master/CMakeLists.txt#L32

devshgraphicsprogramming:
Also funny story `CEGUI_BUILD_SHARED_LIBS_WITH_STATIC_DEPENDENCIES` does not show on the Linux CMake-GUI config screen ;)

devshgraphicsprogramming:
Tested on Windows and Linux.

Merged as its 99% complete, further work should be done on a new PR (the helper refactoring etc.)

ext/CEGUI/CEGUI.cpp
devshgraphicsprogramming: done in latest commit

ext/CEGUI/CEGUI.cpp
devshgraphicsprogramming: done in latest commit

PR #297, opened 2019-05-11
Smooth Normal Computation

devshgraphicsprogramming:
Awaiting a spatial hash close-vertex search and its ready for merging

image
image


devshgraphicsprogramming:
You can nuke whatever is guarded by an `#ifndef NEW_MESHES` that is to do with normals, tangents, etc.

Przemog1:
Since we have margin of tolerance for vertex position comparison, there may be "connected" vertices in different cells, this problem is not resolved yet.
Also hashTableSize and cellSize parameters need to be adjusted.

manhnt9:
A little note: this feature also presents in Assimp library, which we will revive in the future.

devshgraphicsprogramming:
yes but:
1. ASSIMP is an extension not a core feature
2. We need to be able to do this on our own meshes (not in ASSIMP data structures)

src/irr/asset/CMeshManipulator.cpp
```diff
@@ -698,19 +700,27 @@ asset::ICPUMeshBuffer* CMeshManipulator::createMeshBufferUniquePrimitives(asset:
...
-	if (makeTriangleSoup)
+	//will get rid of that
+	if (true)
 		outbuffer = createMeshBufferUniquePrimitives(inbuffer);
 	else
 		outbuffer = createMeshBufferDuplicate(inbuffer);
 
-	CSmoothNormalGenerator::calculateNormals(outbuffer,creaseAngle);
+
+	std::cout << "\n\n computation time: ";
```
devshgraphicsprogramming: spit out the profiling only in `_IRR_DEBUG` if defined

src/irr/asset/CSmoothNormalGenerator.cpp
```diff
@@ -7,6 +7,12 @@ namespace irr
 namespace asset
 {
 
+static inline bool compareVertexPosition(const core::vectorSIMDf& a, const core::vectorSIMDf& b, float epsilon)
```
devshgraphicsprogramming, line 21: this needs to be the default to the `calculateSmoothNormals` function's std::function parameter, need to support user define-able vertex comparison criterions

src/irr/asset/CMeshManipulator.h
```diff
@@ -66,7 +66,7 @@ class CMeshManipulator : public IMeshManipulator
 	virtual asset::ICPUMeshBuffer* createMeshBufferUniquePrimitives(asset::ICPUMeshBuffer* inbuffer) const;
 
 	//
-	virtual asset::ICPUMeshBuffer* calculateSmoothNormals(asset::ICPUMeshBuffer* inbuffer, const float creaseAngle, bool makeTriangleSoup = true) const override;
+	virtual asset::ICPUMeshBuffer* calculateSmoothNormals(asset::ICPUMeshBuffer* inbuffer, const float creaseAngle, float epsilon = 1.525e-5f) const override;
```
devshgraphicsprogramming: instead of epsilon I want a `std::function` vertex comparison predicate

src/irr/asset/CSmoothNormalGenerator.cpp
```diff
@@ -76,22 +84,127 @@ void CSmoothNormalGenerator::processConnectedVertices(asset::ICPUMeshBuffer* buf
...
+
+uint32_t CSmoothNormalGenerator::VertexHashMap::hash(const CSmoothNormalGenerator::Vertex& vertex) const
+{
+	static constexpr uint32_t primeNumber1 = 73856093;
+	static constexpr uint32_t primeNumber2 = 19349663;
+	static constexpr uint32_t primeNumber3 = 83492791;
+
+	const core::vector3df_SIMD position = vertex.position / cellSize;
+
+	return	(((uint32_t)position.x * primeNumber1) ^
```
devshgraphicsprogramming: This is 100% correct ;)

src/irr/asset/CSmoothNormalGenerator.cpp
```diff
@@ -76,22 +84,127 @@ void CSmoothNormalGenerator::processConnectedVertices(asset::ICPUMeshBuffer* buf
...
+
+#pragma region hashing
+
+CSmoothNormalGenerator::VertexHashMap::VertexHashMap(size_t _hashTableSize, float _cellSize)
+	:hashTableSize(_hashTableSize), cellSize(_cellSize)
+{
+	hashTable.reserve(hashTableSize);
+
+	for (int i = 0; i < hashTableSize; i++)
+		hashTable.emplace_back();
```
devshgraphicsprogramming: you know you could just use a std::vertex::resize instead of a reserve+emplace loop

devshgraphicsprogramming:
> 
> 
> I'm pretty sure it is correct.
> 
> > The way as it is now, multiple positions will be getting same hash easily
> 
> Thats the point of it, with this hash function you want to find vertices that reside in the same "cube"

It is 100% correct, it makes integer voxel coordinates, no bit cast

src/irr/asset/CSmoothNormalGenerator.cpp
```diff
@@ -0,0 +1,254 @@
...
+	processConnectedVertices(buffer, vertexArray, epsilon, normalAttrID, vxcmp);
+
+	return buffer;
+}
+
+CSmoothNormalGenerator::VertexHashMap::VertexHashMap(size_t _vertexCount, uint32_t _hashTableMaxSize, float _cellSize)
+	:hashTableMaxSize(_hashTableMaxSize), 
+	cellSize(_cellSize)
+{
+	_IRR_DEBUG_BREAK_IF((!core::isPoT(_hashTableMaxSize)));
```
devshgraphicsprogramming: this needs to be an actual `assert(!core::isPoT(_hashTableMaxSize))`, we want the program to crash (with a message) 

`_IRR_DEBUG_BREAK_IF` only breaks the debugger if one is used, otherwise execution continues.

src/irr/asset/CMeshManipulator.cpp
```diff
@@ -718,6 +334,19 @@ asset::ICPUMeshBuffer* CMeshManipulator::createMeshBufferUniquePrimitives(asset:
...
 }
 
+//
+asset::ICPUMeshBuffer* CMeshManipulator::calculateSmoothNormals(asset::ICPUMeshBuffer* inbuffer, float epsilon,
+	asset::E_VERTEX_ATTRIBUTE_ID normalAttrID, VxCmpFunction vxcmp) const
+{
+	if (inbuffer == nullptr)
+		return nullptr;
+
+	asset::ICPUMeshBuffer* outbuffer = createMeshBufferDuplicate(inbuffer);
```
devshgraphicsprogramming: Maybe take heed from 
https://github.com/buildaworldnet/IrrlichtBAW/blob/839516417231e3a47dcb1781b9414965ca74a85b/include/irr/asset/IMeshManipulator.h#L105

because the mesh does not always need to be created as a duplicate (in-fact that can be harmful to other optimizations).

src/irr/asset/CMeshManipulator.cpp
```diff
@@ -718,6 +334,19 @@ asset::ICPUMeshBuffer* CMeshManipulator::createMeshBufferUniquePrimitives(asset:
 	return clone;
 }
 
+//
+asset::ICPUMeshBuffer* CMeshManipulator::calculateSmoothNormals(asset::ICPUMeshBuffer* inbuffer, float epsilon,
+	asset::E_VERTEX_ATTRIBUTE_ID normalAttrID, VxCmpFunction vxcmp) const
+{
+	if (inbuffer == nullptr)
```
devshgraphicsprogramming, line 341: you also need to check whether the index type is `EIT_UNKNOWN` 
(that the mesh has no shared vertices)

devshgraphicsprogramming:
@Florastamine or @manhnt9 could you check out this branch and just see if it builds on Linux?
(I don't need to know whether it works)

PR #301, opened 2019-05-23
Derivative Mapping and GLSL Include Handling

devshgraphicsprogramming:


include/irr/asset/IIncluder.h
```diff
@@ -0,0 +1,45 @@
...
+    virtual ~IIncluder() = default;
+
+public:
+    IIncluder() : m_searchDirectories{""} {}
+
+    virtual void addSearchDirectory(const std::string& _searchDir) { m_searchDirectories.push_back(_searchDir); }
+
+    std::string getInclude(const std::string& _path) const
+    {
+        for (const std::string& searchDir : m_searchDirectories)
```
devshgraphicsprogramming, line 26: order in which you try them is important, you dont want to try a parent directory before a deeper (more matching) one

devshgraphicsprogramming:
> 
> 
> I know,
> TODO here:
> 
>     * hardcore warp size to 32 and get rid of `IVideoCapabilityReporter` dependency
> 
>     * move `CGLSLScanBuiltinIncludeLoader` to `asset` ns
> 
>     * bring back in-source include guards

should we have a `glsl` namespace for this?

After all this is crap that will never get serialized to a .baw file, and `asset` doesn't really need this glsl include processing, glsl preprocessor, nor a compiler... only SPIR-V reflection/introspection

And if SPIR-V reflection => you need GLSL compiler as well (shaderc) then `asset` could depend on `core`, `io` (future `system`) and `glsl`

devshgraphicsprogramming:
> 
> 
> not a good idea (discontinuity).
> 
> I'd fit a quadratic surface to a 3x3 sample neighbourhood of control points, and get the derivative analytically at the central point.
> 
> Example in 1D.
> 
> ```c++
> float samples[3] = ...;
> // x=-1 => a-b+c = samples[0]
> // x= 0 => c = samples[1]
> // x=+1 => a+b+c = samples[2]
> // a = (samples[0]+samples[2])*0.5-samples[1]
> // b = (samples[2]-samples[0])*0.5
> // dy/dx = 2ax+b
> // dy/dx(0) = b
> float dydx = (samples[2]-samples[0])*0.5;
> ```


btw equation for 2D biquadratic is
```cpp
vec2 gradient = vec2(texelFetchOffset(tex,current,ivec2(1,0)).r-texelFetchOffset(tex,current,ivec2(-1,0)).r,texelFetchOffset(tex,current,ivec2(0,1)).r-texelFetchOffset(tex,current,ivec2(0,-1)).r)*0.5;
```

which is actually identical to an average of the gradient of 4 bilinear surfaces around a pixel

source/Irrlicht/CIrrDeviceStub.cpp
```diff
@@ -48,7 +48,7 @@ CIrrDeviceStub::CIrrDeviceStub(const SIrrlichtCreationParameters& params)
 
 	checkVersion(params.SDK_version_do_not_use);
 
-    IncludeHandler = core::make_smart_refctd_ptr<asset::IIncludeHandler>(new asset::CIncludeHandler(FileSystem), core::dont_grab);
+    IncludeHandler = new asset::CIncludeHandler(FileSystem);
```
devshgraphicsprogramming: why ditch the smart_ptr ?

Crisspl: gives compile errors with abstract classes

devshgraphicsprogramming: has the same syntax as 
https://en.cppreference.com/w/cpp/memory/unique_ptr/make_unique
you don't need the new, just the args

devshgraphicsprogramming: `IncludeHandler = core::make_smart_refctd_ptr<asset::CIncludeHandler>(FileSystem);`

Crisspl:
>btw equation for 2D biquadratic is
>```
>vec2 gradient = vec2(texelFetchOffset(tex,current,ivec2(1,0)).r->texelFetchOffset(tex,current,ivec2(-1,0)).r,texelFetchOffset(tex,current,ivec2(0,1)).r->texelFetchOffset(tex,current,ivec2(0,-1)).r)*0.5;
>```



and it's sufficient for continuity?

source/Irrlicht/CIrrDeviceStub.cpp
Crisspl: ah, my bad, it was because i wrote core::make_smart_refctd_ptr<asset:: **IIncludeHandler** >

Crisspl: But `core::make_smart_refctd_ptr<asset::CIncludeHandler>(FileSystem, core::dont_grab);` doesnt work, it wants to put dont_grab into CIncludeHandler's constructor :/

devshgraphicsprogramming:
> and it's sufficient for continuity?

should be ok for generating derivative maps of the same resolution as the height map

source/Irrlicht/CIrrDeviceStub.cpp
devshgraphicsprogramming: you don't need a `core::dont_grab` I don't think so

devshgraphicsprogramming: if you do then I've fucked up the definition of `template <class T> ...... core::make_smart_refctd_ptr`

src/irr/asset/IGLSLCompiler.cpp
```diff
@@ -6,13 +6,42 @@
 namespace irr { namespace asset
 {
 
+    class MyIncluder : public shaderc::CompileOptions::IncluderInterface
```
devshgraphicsprogramming: and you've pushed :D

source/Irrlicht/CIrrDeviceStub.cpp
Crisspl: >you don't need a core::dont_grab I don't think so

well, it's easy to workaround (just drop after creating smart ptr), but there's
```
	template< class T, class... Args >
	inline smart_refctd_ptr<T> make_smart_refctd_ptr(Args&& ... args, dont_grab_t t)
```
so seems like it's supposed to work. But I don't know why it doesn't or how to do it other way

`dont_grab` is needed because while creating a ref-counted object it obviously has ref-count=1 and after creating smart ptr, it then has ref-count=2. But smart pointer's dtor drops only once so it's a memleak

src/irr/asset/IGLSLCompiler.cpp
Crisspl: fuuuu.. i'll remove and force push

source/Irrlicht/CIrrDeviceStub.cpp
devshgraphicsprogramming: `make_smart_refctd_ptr` should just `dont_grab` by default

devshgraphicsprogramming: I fucked up bigtime, the only definition of `make_smart_refctd_ptr` should have been:

```cpp
template< class T, class... Args >
	inline smart_refctd_ptr<T> make_smart_refctd_ptr(Args&& ... args)
	{
		T* obj = new T(std::forward<Args>(args)...);
		smart_refctd_ptr<T> smart(obj,dont_grab);
		return smart;
}
```

Crisspl: ah, ok, so i'll push it to this branch

examples_tests/15.LayeredCubeMapShadows/deriv_map_gen.comp
```diff
@@ -1,97 +1,46 @@
...
-//place for #defines to put with sprintf
-%s
+layout (local_size_x = 16, local_size_y = 16) in;
 
-#if 0
-#extension GL_NV_compute_shader_derivatives : enable
-#if defined(GL_NV_compute_shader_derivatives)
-layout (derivative_group_quadsNV) in;
+layout (binding = 7) uniform sampler2D bumpMapSampler;
+layout (binding = 0, rg16_snorm) uniform image2D derivativeMapImage;
```
devshgraphicsprogramming: rg8_snorm

examples_tests/15.LayeredCubeMapShadows/deriv_map_gen.comp
```diff
@@ -1,97 +1,46 @@
...
 
-#define IRR_DDX_IS_DEFINED 1
-#define IRR_DDY_IS_DEFINED 1
+layout (location = 0) uniform float uHeightScaleFactor = 0.001;
 
-float DDX(in float x)
-{
-	return dFdx(x);
-}
+shared float smem[324];//18*18
```
devshgraphicsprogramming, line 11: use a macro to make it configurable

examples_tests/15.LayeredCubeMapShadows/deriv_map_gen.comp
Crisspl: thx, no idea how it worked

examples_tests/15.LayeredCubeMapShadows/deriv_map_gen.comp
Crisspl: smem size? But what fot?

examples_tests/15.LayeredCubeMapShadows/deriv_map_gen.comp
```diff
@@ -1,97 +1,46 @@
...
 
-float DDY(in float y)
+int getAddr(in ivec2 threadID)
```
devshgraphicsprogramming, line 14: macro like `(18*(threadID.y+1) + threadID.x+1)` should work much better

Crisspl: what's the difference?

examples_tests/15.LayeredCubeMapShadows/deriv_map_gen.comp
```diff
@@ -1,97 +1,46 @@
...
-	if ((G_IDX.x & 1) == 0 && (G_IDX.y & 1) == 0)
-		imageStore(derivativeMapImage, G_IDX/2, vec4(d, 0.0, 0.0, 0.0));
-#else
-	const uint LC_IDX = gl_LocalInvocationIndex;
-	smem[LC_IDX] = textureLod(bumpMapSampler, vec2(1.0, 1.0)/vec2(RESOLUTION) * vec2(G_IDX), 0.0).r * uHeightScaleFactor;
+	vec2 bumpMapSz = vec2(textureSize(bumpMapSampler, 0));
+
+	const ivec2 LC_IDX = ivec2(gl_LocalInvocationID.xy);
+	smem[getAddr(LC_IDX)] = textureLod(bumpMapSampler, vec2(1.0, 1.0)/vec2(RESOLUTION) * vec2(G_IDX), 0.0).r * uHeightScaleFactor;
+	if (LC_IDX.x == 0 || LC_IDX.x == 15)
```
devshgraphicsprogramming: don't do the if-statements

examples_tests/15.LayeredCubeMapShadows/deriv_map_gen.comp
```diff
@@ -1,97 +1,46 @@
...
 	
-	const uint SMEM_IDX = LC_IDX/2u * 2u;
-	float d = smem[SMEM_IDX+1u] - smem[SMEM_IDX];
-	if ((G_IDX.x & 1) == 0 && (G_IDX.y & 1) == 0)
-		imageStore(derivativeMapImage, G_IDX/2, vec4(d, 0.0, 0.0, 0.0));
-#endif
+	vec2 d = vec2(
+		smem[getAddr(LC_IDX+ivec2(1, 0))] - smem[getAddr(LC_IDX-ivec2(1, 0))],
+		smem[getAddr(LC_IDX+ivec2(0, 1))] - smem[getAddr(LC_IDX-ivec2(0, 1))]
+	) * bumpMapSz;
```
devshgraphicsprogramming: you're missing the factor of half (because you're taking the gradient across 2 pixels not 1)

examples_tests/15.LayeredCubeMapShadows/deriv_map_gen.comp
```diff
@@ -1,97 +1,46 @@
...
+	if (LC_IDX.x == 0 || LC_IDX.x == 15)
+	{
+		ivec2 offset = ivec2(mix(ivec2(1, 0), ivec2(-1, 0), LC_IDX.x == 0));
+		smem[getAddr(LC_IDX+offset)] = textureLod(bumpMapSampler, vec2(1.0, 1.0)/vec2(RESOLUTION) * vec2(G_IDX+offset), 0.0).r * uHeightScaleFactor;
+	}
+	if (LC_IDX.y == 0 || LC_IDX.y == 15)
+	{
+		ivec2 offset = ivec2(mix(ivec2(0, 1), ivec2(0, -1), LC_IDX.y == 0));
+		smem[getAddr(LC_IDX+offset)] = textureLod(bumpMapSampler, vec2(1.0, 1.0)/vec2(RESOLUTION) * vec2(G_IDX+offset), 0.0).r * uHeightScaleFactor;
+	}
```
devshgraphicsprogramming, line 37: also why not apply `uHeightScaleFactor` at the end?

examples_tests/15.LayeredCubeMapShadows/main.cpp
```diff
@@ -230,46 +228,22 @@ int main()
...
-
-    free(deriv_map_gen_glsl);
-    free(deriv_map_gen_glsl_spec);
-
-    GLuint merge_deriv_xy_cs = createComputeShaderFromFile("../merge_deriv_xy.comp");
-
-    video::COpenGLExtensionHandler::extGlBindImageTexture(0, static_cast<const video::COpenGL2DTexture*>(derivMap_x)->getOpenGLName(),
-        0, GL_FALSE, 0, GL_WRITE_ONLY, GL_R16_SNORM);
-    video::COpenGLExtensionHandler::extGlUseProgram(deriv_map_gen_cs_x);
+    video::COpenGLExtensionHandler::extGlBindImageTexture(0, static_cast<const video::COpenGL2DTexture*>(derivMap)->getOpenGLName(),
```
devshgraphicsprogramming: unbind this later!

examples_tests/15.LayeredCubeMapShadows/main.cpp
```diff
@@ -230,46 +228,22 @@ int main()
...
+        0, GL_FALSE, 0, GL_WRITE_ONLY, GL_RG8_SNORM);
+    video::COpenGLExtensionHandler::extGlUseProgram(deriv_map_gen_cs);
```
devshgraphicsprogramming, line 240: this needs to be reset to previous value

include/irr/core/IReferenceCounted.h
```diff
@@ -266,11 +266,19 @@ namespace core
 			inline bool operator>(const smart_refctd_ptr<U>& other) const { return ptr > other.ptr; }
 	};
 
+    // create and object and make a smart pointer without increasing reference count above 1
```
devshgraphicsprogramming: why did the fix get reverted?

source/Irrlicht/CIrrDeviceStub.cpp
```diff
@@ -49,6 +49,9 @@ CIrrDeviceStub::CIrrDeviceStub(const SIrrlichtCreationParameters& params)
 	checkVersion(params.SDK_version_do_not_use);
 
     IncludeHandler = core::make_smart_refctd_ptr<asset::CIncludeHandler>(FileSystem);
+    //TODO it should be core::make_smart_refctd_ptr<asset::CIncludeHandler>(FileSystem, core::dont_grab);
+    //but doesn't work (on MSVC at least), wants to put dont_grab into CIncludeHandler's ctor
+    IncludeHandler->drop();//remove this drop() when above works
```
devshgraphicsprogramming: fix got reverted?

devshgraphicsprogramming:
Can you post a screenie of the derivative mapping?

examples_tests/15.LayeredCubeMapShadows/deriv_map_gen.comp
devshgraphicsprogramming: you've coded the function as-if there's real SIMD in a single GPU thread... so there are useless multiplies here also a cast to int for no reason.

The upside of a macro is that it can function like a template ;) (doesn't care if you use uint or int)

examples_tests/15.LayeredCubeMapShadows/deriv_map_gen.comp
devshgraphicsprogramming: one day when all devices support 1024-sized work groups....

include/irr/core/IReferenceCounted.h
Crisspl: Commit message ;) crashes on fence-related code deep in the engine and can't figure why

devshgraphicsprogramming: because the fence stuff worked with a fucked up reference count ;)

devshgraphicsprogramming: shit, we're leaking fences

examples_tests/15.LayeredCubeMapShadows/deriv_map_gen.comp
Crisspl: well, i thought `dot` will make it more SIMD. The cast is because my driver is a little retarded and while compiling without cast, gives errors like `dot` is returning always float, same thing with `mix`.

>(doesn't care if you use uint or int)

well, that's actually source of i-dont-know-whats-going-on bugs :P

examples_tests/15.LayeredCubeMapShadows/deriv_map_gen.comp
Crisspl: ?? how?

examples_tests/15.LayeredCubeMapShadows/deriv_map_gen.comp
Crisspl: oh, yea, thanks

examples_tests/15.LayeredCubeMapShadows/deriv_map_gen.comp
Crisspl: because now it works on smaller numbers = possibly better precision. Generally i think that way, but here no point

Crisspl:
d

examples_tests/15.LayeredCubeMapShadows/deriv_map_gen.comp
devshgraphicsprogramming: if your compiler is complaining about dot being a float then you shouldn't use it... it might actually calc the index as a float XD

Crisspl: it probably does, but they're integer's small enough to be representable on 32bit float

examples_tests/15.LayeredCubeMapShadows/deriv_map_gen.comp
devshgraphicsprogramming: ok:
1) Your texture coordinate is wrong, you'd need to offset by 0.5 of a pixel
2) You should use texelFetch instead anyway.
3) Best way to do this is like this

```cpp
for (uint vid=gl_LocalInvocationIndex; vid<SMEM_SIZE; vid+=WORK_GROUP_TOTAL_SIZE)
{
   ivec2 texel = gl_WorkGroupID.xy*gl_WorkGroupSize.xy-ivec2(1,1);
   int row = vid/SMEM_DIM_X;
   texel.x += vid-row*SMEM_DIM_X;
   texel.y += row;
   smem[vid] = texelFetch(bumpMapSampler,texel,0).r;
}
```

compiler should unroll the loop and you will just get the texel calculation math + 1 conditional texel fetch

examples_tests/15.LayeredCubeMapShadows/deriv_map_gen.comp
devshgraphicsprogramming: float precision is only an issue if your operands to ADD or SUB differ vastly in magnitude

examples_tests/15.LayeredCubeMapShadows/deriv_map_gen.comp
devshgraphicsprogramming: but converting from float to int and back is not free

examples_tests/15.LayeredCubeMapShadows/deriv_map_gen.comp
Crisspl: hm smart
But `texelFetch` require that bump map and derivative map are the same size + nothing-doing threads if the size is not multiple of 16 (if-statement needed here). Filtering is so costly?

devshgraphicsprogramming:
Nice screenshots, the derivative map needs some mip-mapping XD

You should do a reduction in shared memory, you can recast the 18*18 smem buffer as a 16*16
With 16x16 WG size you can compute extra 4 mip levels from the base 

Then you can make a multi-pass solution that reduces 4 mips at a time:
Initial Run: 0 (already made by creating the map), 1, 2, 3, 4
Iteration 1: 5,6,7,8
Iteration 2: 9,10,11,12
...etc...

Don't try to be clever with Z-order curves and nice reductions as the functions will probably need to be updated at a certain point to use a nice kernel and modulate roughness maps when implementing Toksvig AA in the future.

But do add the `derivativeMapComputeMipMap` functions to a built-in header.
(disclaimer: there's probably no way to pass shared memory to a function, so requiring the user to define a macro might be in order)

Crisspl:
Currently `regenerateMipMapLevels()` is called after running derivative map gen CS. (but screenshot is done with bilinear filtering on derivative map sampler)
Do you know if there's a way to bind more than one mip level to one `image2D`?

edit:
>glBindImageTexture binds a single level of a texture to an image unit

i think not..

devshgraphicsprogramming:
> i think not..

Irrlicht has the downside of only using the multi-bind for storage images, see FAQ point for 18
https://www.khronos.org/registry/OpenGL/extensions/ARB/ARB_multi_bind.txt

But yes, you can only attach a single mip-map you'd need 5 image bind points

BTW: don't use texture sampler as input for subsequent mip-map calculation passes, because you'll need 2 memory barrier bits instead of just one (GL_SHADER_IMAGE_ACCESS_BARRIER_BIT). OBviously you'd need all the barrier bits after the very last mipmap.

devshgraphicsprogramming:
I changed the branch that this will get merged into.

Also if you want there's a new sponza with bump maps (instead of normal maps) all packed nicely in a zip
https://github.com/buildaworldnet/IrrlichtBAW/commit/1b9b426022f8c1a97f2d175a9428ea0f6ec93347

examples_tests/15.LayeredCubeMapShadows/deriv_map_gen.comp
```diff
@@ -4,10 +4,12 @@ layout (local_size_x = 16, local_size_y = 16) in;
 
 layout (binding = 7) uniform sampler2D bumpMapSampler;
 layout (binding = 0, rg8_snorm) uniform image2D derivativeMapImage;
+layout (binding = 1, rg8_snorm) uniform image2D derivativeMapImage_mm[4];
 
 layout (location = 0) uniform float uHeightScaleFactor = 0.001;
 
 shared float smem[324];//18*18
+shared vec2 smem2[256];//16*16
```
devshgraphicsprogramming, line 12: re-use `smem`, by basically doubling the smem required you'll reduce occupancy greatly

examples_tests/15.LayeredCubeMapShadows/deriv_map_gen.comp
devshgraphicsprogramming: bump map and drivative map should absolutely be the same size, because from frequency analysis theory (mixing fourier and Shannon Nyquist):
1. A bump-map of resolution N contains frequencies of at most K Hz
2. A derivative applied before a fourier transform does introduces a mutliplicative term to the original fourier transform, hence if the original did not contain frequencies above K, then the fourier of the derivative will also not contain any frequencies above K
3. Hence if derivative map is lower resolution them it looses detail or if its higher its wasting pixels because the signal can be reconstructed anyway from a lower amount of samples

PR #305, opened 2019-05-25
WIP: CImageData::createInConvertedFormat()

ghost:
Plus a bunch of post-#273 fixes that are not really essential/related.

- [ ] `CImageData::createInConvertedFormat()`
- [ ] Make `EF_R8_SRGB` testing working as expected
- [ ] Misc. fixes (@devshgraphicsprogramming 's comments)


ghost:
I'm still not sure on when the actual conversion process would happen, i. e. where would we call `createInConvertedFormat()`. `loadAsset()` returns a `asset::IAsset*`, so no image manipulation outside asset loading - maybe do convert during image buffer filling and such, and turn it off if we're not doing image testing.

devshgraphicsprogramming:
> loadAsset() returns a asset::IAsset*, so no image manipulation outside asset loading - maybe do convert during image buffer filling and such,

`asset:IAsset` has a `getType` or similar function which tells you that it is safe to `reinterpret_cast` an `asset::IAsset*` to an `asset::CImageData*` etc.

include/irr/video/encodePixels.h
```diff
@@ -2333,11 +2333,24 @@ namespace irr { namespace video
...
+            
+            for (uint32_t i = 0u; i < 3u; ++i)
+            {
+                T& s = _lin[i];
+                if (s <= 0.0031308) s *= 12.92;
+                else s = 1.055 * std::pow(s, 1./2.4) - 0.055;
+            }
+        }
+        
+        inline double lin2srgb(double _lin)
```
devshgraphicsprogramming, line 2349: why are there two identical functions? the number of components to convert could also bea template parameter

source/Irrlicht/CMakeLists.txt
```diff
@@ -130,12 +130,9 @@ set(IRRLICHT_SRCS_COMMON
 	CColorConverter.cpp
 	CImage.cpp
 	${CMAKE_SOURCE_DIR}/src/irr/video/convertColor.cpp
-	${CMAKE_SOURCE_DIR}/src/irr/asset/CImageLoaderBMP.cpp
-	${CMAKE_SOURCE_DIR}/src/irr/asset/CImageLoaderDDS.cpp
```
devshgraphicsprogramming, line 134: so you're nuking the DDS loader without providing a replacement?

ghost: darn, I wiped it out thinking there wasn't any code using the DDS loader... apparently example 04 is still making some use of it.

putting the DDS one back in for the time being until a more robust solution appears (#253)

devshgraphicsprogramming:
@Przemog1 take this over at some point in July.

PR #307, opened 2019-05-27
SSE4.2 implementation of aabbox3d

Przemog1:


include/aabbox2dSIMD.h
```diff
@@ -0,0 +1,28 @@
...
+
+#include "irr/core/math/irrMath.h"
+
+namespace irr { namespace core {
+
+class aabbox2dSIMDf
+{
+public:
+	inline aabbox2dSIMDf()
+		: bounds(-1.0f, -1.0f, 1.0f, 1.0f) {};
```
devshgraphicsprogramming, line 12: FLT_MAX, otherwise this won't work for boxes with max edge <(1,1) or min edge > (-1,-1)

include/aabbox2dSIMD.h
```diff
@@ -0,0 +1,28 @@
...
+
+	inline aabbox2dSIMDf(const core::vectorSIMDf& _bounds)
```
devshgraphicsprogramming, line 14: nope, bad API

you only want
```cpp
inline aabb2dSIMDf(const core::vectorSIMDf& _point) : bound(_point.xyxy()^core::vectorSIMDu32(0x80000000u,0x80000000u,0u,0u)) {}
```

include/aabbox2dSIMD.h
```diff
@@ -0,0 +1,28 @@
+#ifndef __IRR_AABBOX_2D_SIMD_H_INCLUDED__
+#define __IRR_AABBOX_2D_SIMD_H_INCLUDED__
+
+#include "irr/core/math/irrMath.h"
+
+namespace irr { namespace core {
+
+class aabbox2dSIMDf
```
devshgraphicsprogramming, line 8: rename to `aabb2dSIMDf`

include/aabbox2dSIMD.h
```diff
@@ -0,0 +1,28 @@
...
+{
+public:
+	inline aabbox2dSIMDf()
+		: bounds(-1.0f, -1.0f, 1.0f, 1.0f) {};
+
+	inline aabbox2dSIMDf(const core::vectorSIMDf& _bounds)
+		: bounds(_bounds) {};
+
+	inline aabbox2dSIMDf(float minx, float miny, float maxx, float maxy)
```
devshgraphicsprogramming, line 17: do it from two `core::vectorSIMDf`s instead

include/aabbox2dSIMD.h
```diff
@@ -0,0 +1,28 @@
...
+		: bounds(core::vectorSIMDf(minx, miny, maxx, maxy)) {};
+
+private:
+	core::vectorSIMDf bounds; // xy - min, zw - max
```
devshgraphicsprogramming, line 21: y is negative min

include/aabbox3dSIMD.h
```diff
@@ -0,0 +1,17 @@
+#ifndef __IRR_AABBOX_3D_SIMD_H_INCLUDED__
+#define __IRR_AABBOX_3D_SIMD_H_INCLUDED__
+
+#include "irr/core/math/irrMath.h"
+
+namespace irr {
+namespace core {
+
+class aabbox3dSIMDf
```
devshgraphicsprogramming, line 9: `aabb3dSIMDf`

include/aabbox2dSIMD.h
```diff
@@ -0,0 +1,28 @@
...
+{
+public:
+	inline aabbox2dSIMDf()
+		: bounds(-1.0f, -1.0f, 1.0f, 1.0f) {};
+
+	inline aabbox2dSIMDf(const core::vectorSIMDf& _bounds)
+		: bounds(_bounds) {};
+
+	inline aabbox2dSIMDf(float minx, float miny, float maxx, float maxy)
+		: bounds(core::vectorSIMDf(minx, miny, maxx, maxy)) {};
```
devshgraphicsprogramming, line 18: you need to negate the minimums

include/aabbox2dSIMD.h
```diff
@@ -0,0 +1,28 @@
...
+
+private:
```
devshgraphicsprogramming, line 20: because `bounds` is private (rightfully so because it has a weird and packed encoding), you need public getter functions.
```cpp
inline core::vectorSIMDf getMinEdge() const
{
   auto retval = -bounds;
   retval.makeSafe2D();
   return retval;
}
inline core::vectorSIMDf getMaxEdge() const
{
   auto retval = bounds.zwzw(); // can be anything zw__()
   retval.makeSafe2D();
   return retval;
}
```

devshgraphicsprogramming:
One more thing, because this is nice and new clean API... the files should be in `include/irr/namespace/...`

In the case of Axis Aligned Bounding Boxes, they should be in `include/irr/core/math`

include/irr/core/math/aabb2dSIMD.h
```diff
@@ -0,0 +1,169 @@
+#ifndef __IRR_AABB_SIMD_2D_H_INCLUDED__
+#define __IRR_AABB_SIMD_2D_H_INCLUDED__
+
+#include "irr/core/math/irrMath.h"
+#include "irr/core/memory/memory.h"
+#include "irr/core/alloc/AlignedBase.h"
+
+#include <iostream>
+
+#define SIGN_FLIP_MASK_XY	core::vector3du32_SIMD(0x80000000, 0x80000000, 0x00000000, 0x00000000)
```
devshgraphicsprogramming, line 10: those need to be undefined at the end of the file to not pollute the preprocessor

include/irr/core/math/aabb2dSIMD.h
```diff
@@ -0,0 +1,169 @@
...
+#define SIGN_FLIP_MASK_XYZW core::vector3du32_SIMD(0x80000000, 0x80000000, 0x80000000, 0x80000000)
+
+
+namespace irr { namespace core {
+
+class aabb2dSIMDf
+{
+public:
+	inline aabb2dSIMDf()
+		: box( FLT_MAX, FLT_MAX, FLT_MAX, FLT_MAX ) {};
```
devshgraphicsprogramming: negative on everything! you're using `core::max_` to join boxes and points

include/irr/core/math/aabb2dSIMD.h
```diff
@@ -0,0 +1,169 @@
...
+
+	inline aabb2dSIMDf(const core::vectorSIMDf& _box)
```
devshgraphicsprogramming: add the keyword `explicit` we don't want a conversion to happen by accident

include/irr/core/math/aabb2dSIMD.h
```diff
@@ -0,0 +1,169 @@
...
+		: box(_box ^ SIGN_FLIP_MASK_XY) {};
+
+	inline void addBox(const aabb2dSIMDf& _box)		
+	{ 
+		box = core::max_(box, _box.getBounds()); 
+	}
+
+	inline void addPoint(const core::vectorSIMDf& point)	
+	{
+		addBox(point);
```
devshgraphicsprogramming: make construction of a aabbox explicit here (construct the object)

include/irr/core/math/aabb2dSIMD.h
```diff
@@ -0,0 +1,169 @@
...
+	{ 
+		box = core::max_(box, _box.getBounds()); 
+	}
+
+	inline void addPoint(const core::vectorSIMDf& point)	
+	{
+		addBox(point); 
+	}
+
+	inline core::vectorSIMDf getBounds() const
```
devshgraphicsprogramming: there's no such function in the original., there shoudl be a separate `getMinEdge` and `getMaxEdge`

include/irr/core/math/aabb2dSIMD.h
```diff
@@ -0,0 +1,169 @@
...
+	{
+		return box ^ SIGN_FLIP_MASK_XY;
+	}
+
+
+	//! Resets the bounding box to a one-point box.
+	/** \param x X coord of the point.
+	\param y Y coord of the point. */
+	inline void reset(float x, float y)
```
devshgraphicsprogramming: reset from a `vectorSIMDf` point

include/irr/core/math/aabb2dSIMD.h
```diff
@@ -0,0 +1,169 @@
...
+	{
+		box = core::vectorSIMDf(-x, -y, x, y);
+	}
+
+	//! Resets the bounding box.
+	/** \param initValue New box to set this one to. */
+	inline void reset(const aabb2dSIMDf& initValue)
```
devshgraphicsprogramming: don't need this, just provide copy-assignment

include/irr/core/math/aabb2dSIMD.h
Przemog1: should this function stay as an overload?

include/irr/core/math/aabb2dSIMD.h
```diff
@@ -0,0 +1,169 @@
...
+
+	//! Resets the bounding box to a one-point box.
+	/** \param initValue New point. */
+	//void reset(const aabb2dSIMDf& initValue);
+
+	//! Get center of the bounding box
+	/** \return Center of the bounding box. */
+	inline core::vectorSIMDf getCenter() const
+	{
+		core::vectorSIMDf minToCenterVec = (box + box.zwzw()) * 0.5f;
```
devshgraphicsprogramming: you could have just did the following
```
return (box.zwzw()-box.xyxy())*vectorSIMDf(0.5f,0.5f,0.f,0.f);
```

include/irr/core/math/aabb2dSIMD.h
```diff
@@ -0,0 +1,169 @@
...
+		return ((box ^ SIGN_FLIP_MASK_XY) + minToCenterVec);
+	}
+
+	//! Get extent of the box (maximal distance of two points in the box)
+	/** \return Extent of the bounding box. */
+	core::vectorSIMDf getExtent() const
+	{
+		return box.zwzw() + box;
```
devshgraphicsprogramming: I think its important to have zeros in the zw output coordinate so save to a temporary and call `makeSafe2D` before returning

include/irr/core/math/aabb2dSIMD.h
```diff
@@ -0,0 +1,169 @@
...
+	}
+
+	//! Check if the box is empty.
+	/** This means that there is no space between the min and max edge.
+	\return True if box is empty, else false. */
+	bool isEmpty() const
+	{
+		core::vectorSIMDf extent = getExtent();
+			//epsilon?
+		return (extent.x == 0.0f) || (extent.y == 0.0f);
```
devshgraphicsprogramming: just ```return all((box^SIGN_FLIP_MASK_XY)==box.zwzw());```

include/irr/core/math/aabb2dSIMD.h
```diff
@@ -0,0 +1,169 @@
...
+	}
+
+	//! Get the surface area of the box in squared units
+	inline float getArea() const
+	{
+		core::vectorSIMDf a = box + box.zwxx();	// ( abs(x2-x1), abs(y2-y1) )
```
devshgraphicsprogramming: there's no abs there, update the comment

include/irr/core/math/aabb2dSIMD.h
```diff
@@ -0,0 +1,169 @@
...
+		return a.x * a.y;						// multiplication of both sides of the square
+	}
+
+	//! Stores all 8 edges of the box into an array
```
devshgraphicsprogramming: 4 edges

include/irr/core/math/aabb2dSIMD.h
```diff
@@ -0,0 +1,169 @@
...
+	/** \param edges: Pointer to array of 8 edges. */
+	void getEdges(core::vectorSIMDf& edges) const;
+
+	//! Repairs the box.
+	/** Necessary if for example MinEdge and MaxEdge are swapped. */
+	//TODO
+	void repair();
```
devshgraphicsprogramming: extract the min and max point and turn them into solitary aabb2d then join the two together

include/irr/core/math/aabb2dSIMD.h
Przemog1: I know, its copy pasted from aabbox3d

include/irr/core/math/aabb2dSIMD.h
```diff
@@ -0,0 +1,169 @@
...
+		return result.allBits();
+	}
+
+	//! Check if this box is completely inside the 'other' box.
+	/** \param other: Other box to check against.
+	\return True if this box is completly inside the other box,
+	otherwise false. */
+	bool isFullInside(const aabb2dSIMDf& other) const
+	{
+		core::vector4db_SIMD a = box <= other.box;
```
devshgraphicsprogramming: provide an explanation as to why this function works

include/irr/core/math/aabb2dSIMD.h
```diff
@@ -0,0 +1,169 @@
...
+		core::vector4db_SIMD b = (box.zwxy() ^ SIGN_FLIP_MASK_XYZW) <= other.box;
```
devshgraphicsprogramming: why do you need to check this as well?

include/irr/core/math/aabb2dSIMD.h
```diff
@@ -0,0 +1,169 @@
...
+	}
+
+	//! Determines if the axis-aligned box intersects with another axis-aligned box.
+	/** \param other: Other box to check a intersection with.
+	\return True if there is an intersection with the other box,
+	otherwise false. */
+
+	bool intersectsWithBox(const aabb2dSIMDf& other) const
+	{
+		return
```
devshgraphicsprogramming: I swear there has to be a faster check

include/irr/core/math/aabb2dSIMD.h
```diff
@@ -0,0 +1,169 @@
...
+			isPointInside(other.box) ||
+			isPointInside(other.box.zwxx()) ||
+			isPointInside(other.box.xwxx()) ||
+			isPointInside(other.box.zyxx());
+	}
+
+	//! Tests if the box intersects with a line
+	/** \param line: Line to test intersection with.
+	\return True if there is an intersection , else false. */
+	bool intersectsWithLine(const line3d<float>& line) const;
```
devshgraphicsprogramming: there are no 2d line classes in the engine so you can nuke those

include/irr/core/math/aabb2dSIMD.h
```diff
@@ -0,0 +1,169 @@
...
+	\param linevect Vector of the line.
+	\param halflength Half length of the line.
+	\return True if there is an intersection, else false. */
+	bool intersectsWithLine(const core::vectorSIMDf& linemiddle,
+		const core::vectorSIMDf& linevect, float/*T*/ halflength) const;
+
+
+
+private:
+	core::vectorSIMDf box;
```
devshgraphicsprogramming: this needs a better name

Przemog1: suggestions?

include/irr/core/math/aabb2dSIMD.h
devshgraphicsprogramming: nope because if one wants to do float they can use implicit vectorSIMDf constructor like this
`someBox.reset({x,y});`

include/irr/core/math/aabb2dSIMD.h
devshgraphicsprogramming: `internal_representation`

include/vectorSIMD.h
```diff
@@ -365,8 +365,8 @@ namespace core
 		// functions
 		//! zeroes out out of range components (useful before performing a dot product so it doesnt get polluted with random values)
 		//! WARNING IT DOES COST CYCLES
-		inline void makeSafe2D(void) {_mm_store_si128(pointer,_mm_and_si128(_mm_load_si128(pointer),_mm_set_epi32(0,0,-1,-1)));}
-		inline void makeSafe3D(void) {_mm_store_si128(pointer,_mm_and_si128(_mm_load_si128(pointer),_mm_set_epi32(0,-1,-1,-1)));}
+		inline void makeSafe2D(void) {_mm_store_si128((__m128i*)pointer,_mm_and_si128(_mm_load_si128((__m128i*)pointer),_mm_set_epi32(0,0,-1,-1)));}
```
devshgraphicsprogramming, line 369: reinterpret_cast

include/irr/core/math/aabb2diSIMD.h
```diff
@@ -0,0 +1,155 @@
...
+#include "irr/core/memory/memory.h"
+#include "irr/core/alloc/AlignedBase.h"
+
+namespace irr { namespace core {
+
+class aabb2dSIMDi
+{
+public:
+	inline aabb2dSIMDi()
+		: internalBoxRepresentation(-INT_MAX, -INT_MAX, -INT_MAX, -INT_MAX) {};
```
devshgraphicsprogramming, line 16: just INT_MIN instead

include/irr/core/math/aabb2diSIMD.h
```diff
@@ -0,0 +1,155 @@
...
+	inline void addBox(const aabb2dSIMDi& _box)
+	{
+		//because operator < isn't defined for core::vector32_SIMD core::max_(const core::vector32_SIMD&, const core::vector32_SIMD&) doesn't work yet
+		//internalBoxRepresentation = core::max_(internalBoxRepresentation, _box.getBounds());
+
+		__m128i result = _mm_max_epi32(_box.internalBoxRepresentation.getAsRegister(), internalBoxRepresentation.getAsRegister());
+		_mm_store_si128((__m128i*)internalBoxRepresentation.pointer, result);
+	}
+
+	inline void addPoint(const core::vector4di32_SIMD& point)
```
devshgraphicsprogramming: this could be templated

include/irr/core/math/aabb2dSIMD.h
```diff
@@ -8,90 +8,92 @@
...
-		: box(_box ^ SIGN_FLIP_MASK_XY) {};
+	inline explicit aabb2dSIMDf(const core::vectorSIMDf& _box)
+		: internalBoxRepresentation(_box^ SIGN_FLIP_MASK_XY) {};
 
-	inline void addBox(const aabb2dSIMDf& _box)		
-	{ 
-		box = core::max_(box, _box.getBounds()); 
+	inline void addBox(const aabb2dSIMDf& _box)
+	{
+		internalBoxRepresentation = core::max_(internalBoxRepresentation, _box.getBounds());
```
devshgraphicsprogramming: `internalBoxRepresentation = core::max_(internalBoxRepresentation, _box.internalBoxRepresentation);`

include/irr/core/math/aabb2dSIMD.h
```diff
@@ -8,90 +8,92 @@
 #include <iostream>
 
 #define SIGN_FLIP_MASK_XY	core::vector3du32_SIMD(0x80000000, 0x80000000, 0x00000000, 0x00000000)
-#define SIGN_FLIP_MASK_XYZW core::vector3du32_SIMD(0x80000000, 0x80000000, 0x80000000, 0x80000000)
-
 
 namespace irr { namespace core {
 
+
 class aabb2dSIMDf
```
devshgraphicsprogramming, line 15: this needs to inherit from a AlignedBase class to ensure alignment and `operator new`+`operator delete` overrides

devshgraphicsprogramming: same for the other vector type

include/irr/core/math/aabb2dSIMD.h
```diff
@@ -8,90 +8,92 @@
...
+	inline void addBox(const aabb2dSIMDf& _box)
+	{
+		internalBoxRepresentation = core::max_(internalBoxRepresentation, _box.getBounds());
 	}
 
-	inline void addPoint(const core::vectorSIMDf& point)	
+	inline void addPoint(const core::vectorSIMDf& point)
 	{
-		addBox(point); 
+		addBox(aabb2dSIMDf(point));
```
devshgraphicsprogramming, line 31: this sort of thing could be provided by a base class with CRTP (assuming the base class a CRTP template argument which is a template class itself, see the allocator code for an example of a template on a template class)
```cpp
inline void addPoint(const T& point)
{
   CRTP<T>::addPoint(CRTP<T>(point));
}
```

include/irr/core/math/aabb2dSIMD.h
```diff
@@ -8,90 +8,92 @@
...
 	}
 
-	inline core::vectorSIMDf getBounds() const
+	inline core::vectorSIMDf getMinEdge() const
 	{
-		return box ^ SIGN_FLIP_MASK_XY;
+		core::vectorSIMDf minEdge = getBounds();
```
devshgraphicsprogramming: be consistent and don't use `getBounds()` internally, just use `internalBoxRepresentation`

include/irr/core/math/aabb2dSIMD.h
```diff
@@ -8,90 +8,92 @@
...
-		box = core::vectorSIMDf(-x, -y, x, y);
+		core::vectorSIMDf maxEdge = internalBoxRepresentation.zwzw();
+		maxEdge.makeSafe2D();
+		return maxEdge;
 	}
 
-	//! Resets the bounding box.
-	/** \param initValue New box to set this one to. */
-	inline void reset(const aabb2dSIMDf& initValue)
+	inline core::vectorSIMDf getBounds() const
```
devshgraphicsprogramming: remove this method

include/aabbox2dSIMD.h
devshgraphicsprogramming: we come back to this at the end

include/irr/core/math/aabb2dSIMD.h
devshgraphicsprogramming: another note, it would be good to use the `vector2df_SIMD` typedef for places where we input or return points, its more clear what the expected input and output is

include/irr/core/math/aabb2dSIMD.h
```diff
@@ -8,90 +8,92 @@
...
-		box = initValue.box;
+		return internalBoxRepresentation ^ SIGN_FLIP_MASK_XY;
 	}
 
 	//! Resets the bounding box to a one-point box.
-	/** \param initValue New point. */
-	//void reset(const aabb2dSIMDf& initValue);
+	/** \param x X coord of the point.
+	\param y Y coord of the point. */
+	inline void reset(const core::vectorSIMDf& point)
```
devshgraphicsprogramming: this could actually be a template function of the base class
```cpp
inline void reset(const T& point)
{
   CRTP<T>::operator=(CRTP<T>(point));
}
```

include/irr/core/math/aabb2dSIMD.h
```diff
@@ -8,90 +8,92 @@
...
+	{
+		internalBoxRepresentation = point.xyxy();
+		internalBoxRepresentation ^= SIGN_FLIP_MASK_XY;
+	}
 
 	//! Get center of the bounding box
 	/** \return Center of the bounding box. */
 	inline core::vectorSIMDf getCenter() const
```
devshgraphicsprogramming: could be a template of base
```cpp
      inline T getCenter() const
      {
         return (CRTP<T>::getMinEdge()+CRTP<T>::getMaxEdge())/T(2);
      }

```

include/irr/core/math/aabb2dSIMD.h
```diff
@@ -8,90 +8,92 @@
...
 	{
-		core::vectorSIMDf minToCenterVec = (box + box.zwzw()) * 0.5f;
-		return ((box ^ SIGN_FLIP_MASK_XY) + minToCenterVec);
+		return (internalBoxRepresentation.zwzw() - internalBoxRepresentation) * vectorSIMDf(0.5f, 0.5f, 0.f, 0.f);
 	}
 
+
 	//! Get extent of the box (maximal distance of two points in the box)
 	/** \return Extent of the bounding box. */
 	core::vectorSIMDf getExtent() const
```
devshgraphicsprogramming: could also be templated

include/irr/core/math/aabb2dSIMD.h
```diff
@@ -8,90 +8,92 @@
...
 	bool isEmpty() const
 	{
-		core::vectorSIMDf extent = getExtent();
-			//epsilon?
-		return (extent.x == 0.0f) || (extent.y == 0.0f);
+		return ((internalBoxRepresentation ^ SIGN_FLIP_MASK_XY) == internalBoxRepresentation.zwzw()).all();
 	}
 
 	//! Get the surface area of the box in squared units
 	inline float getArea() const
```
devshgraphicsprogramming: could also template

include/irr/core/math/aabb2dSIMD.h
```diff
@@ -8,90 +8,92 @@
...
 	{
-		core::vectorSIMDf a = box + box.zwxx();	// ( abs(x2-x1), abs(y2-y1) )
-		return a.x * a.y;						// multiplication of both sides of the square
+		core::vectorSIMDf a = internalBoxRepresentation + internalBoxRepresentation.zwxx();
+		return a.x* a.y;
 	}
 
-	//! Stores all 8 edges of the box into an array
+	//! Stores all 4 edges of the box into an array
 	/** \param edges: Pointer to array of 8 edges. */
```
devshgraphicsprogramming: update comment

include/irr/core/math/aabb2dSIMD.h
```diff
@@ -8,90 +8,92 @@
...
-	void getEdges(core::vectorSIMDf& edges) const;
+	void getEdges(core::vectorSIMDf & edges) const;
```
devshgraphicsprogramming: it should be possible to template and inline this I think

include/irr/core/math/aabb2diSIMD.h
devshgraphicsprogramming: actually because INT_MIN is equal to -1-INT_MAX, we cannot use the floating point hack :(

include/irr/core/math/aabb2diSIMD.h
```diff
@@ -0,0 +1,155 @@
...
+	inline float getArea() const
+	{
+		core::vector4di32_SIMD a = internalBoxRepresentation + internalBoxRepresentation.zwxx();	// ( abs(x2-x1), abs(y2-y1) )
+		return a.x * a.y;																			// multiplication of both sides of the square
+	}
+
+	//! Repairs the box.
+	/** Necessary if for example MinEdge and MaxEdge are swapped. */
+	//TODO
+	void repair();
```
devshgraphicsprogramming: could be templated

include/irr/core/math/aabb2diSIMD.h
```diff
@@ -0,0 +1,155 @@
...
+
+private:
+
+	//will change that most likely
+	inline core::vector4di32_SIMD flipSignXY() const
+	{ 
+		return (internalBoxRepresentation ^ core::vector4du32_SIMD(0xFFFFFFFF, 0xFFFFFFFF, 0x00000000, 0x00000000)) + core::vector4du32_SIMD(1, 1, 0, 0);
+	}
+
+	inline core::vector4di32_SIMD flipSignXY(const core::vector4di32_SIMD& vec) const
```
devshgraphicsprogramming, line 89: keep this one and kill the one above

include/vectorSIMD.h
```diff
@@ -324,6 +324,12 @@ namespace core
 
 		inline vectorSIMD_32<T> operator-(T val) const { return (*this)-vectorSIMD_32<T>(val); }
 		inline vectorSIMD_32<T>& operator-=(T val) { return ( (*this) -= vectorSIMD_32<T>(val) ); }
+
+		inline vectorSIMD_32<T>& operator>>(uint32_t val) { _mm_store_si128(reinterpret_cast<__m128i*>(pointer), _mm_srai_epi32(Base::getAsRegister(),val)); return (*this); }
+
+		//comparison
+		inline vector4db_SIMD operator<(const vectorSIMDf& other) const { return _mm_cmplt_epi32(getAsRegister(), other.getAsRegister()); }
```
devshgraphicsprogramming: why are you doing a compare against a vectorSIMDf?

devshgraphicsprogramming: also this comparison op only works for signed ints

You'd need to use CRTP and make vectorSIMD_32<T> a common base class (as `vectorSIMD_32<VECTOR_TYPE<T> >`) for actual `vectorSIMDu32` and `vectorSIMDi32` (instead of them being typedefs as they are now)

include/vectorSIMD.h
```diff
@@ -324,6 +324,12 @@ namespace core
 
 		inline vectorSIMD_32<T> operator-(T val) const { return (*this)-vectorSIMD_32<T>(val); }
 		inline vectorSIMD_32<T>& operator-=(T val) { return ( (*this) -= vectorSIMD_32<T>(val) ); }
+
+		inline vectorSIMD_32<T>& operator>>(uint32_t val) { _mm_store_si128(reinterpret_cast<__m128i*>(pointer), _mm_srai_epi32(Base::getAsRegister(),val)); return (*this); }
```
devshgraphicsprogramming: shift right behaves differently on signed vs unsigned (whether it shifts in copies of the sign bit)

include/vectorSIMD.h
Przemog1: > why are you doing a compare against a vectorSIMDf?

oops

include/vectorSIMD.h
Przemog1: I'm going to use it only for aabbSIMDu (unsigned int aabb) in getCenter() function, where i need to divide by 2.

Przemog1: i.e. 
```cpp
(getMaxEdge() + getMinEdge) >> 2;
```

devshgraphicsprogramming: that's fine(ish)

but you've just added a >> operator to a class that provides both `vectorSIMD32u` and `vectorSIMD32i` so you've introduced a bug

include/irr/core/math/aabb2dSIMD.h
```diff
@@ -88,17 +86,39 @@ class aabb2dSIMDf
...
+	/**
+	A ----- D
+	|		|
+	|		|
+	B ----- C
+
+	array modified by this function will store edges in this order:
+	AB, BC, AD, DC
+	\param edges: Pointer to array of 4 edges. */
+	void getEdges(core::vectorSIMDf* edges) const
```
devshgraphicsprogramming: actually this function is called `getEdges` but its actually `getVertices` if you look at the definition in `aabbox3d<>`
its for getting the corners
(you can rename if you like)

include/irr/core/math/aabb2dSIMD.h
```diff
@@ -0,0 +1,180 @@
...
+	}
+
+	inline void addPoint(const core::vectorSIMDf& point)
+	{
+		addBox(aabb2dSIMDf(point));
+	}
+
+	inline core::vectorSIMDf getMinEdge() const
+	{
+		core::vectorSIMDf minEdge = internalBoxRepresentation;
```
devshgraphicsprogramming: sign flip missing

include/vectorSIMD.h
Przemog1: _mm_srli_epi32 will do for unsigned int?

devshgraphicsprogramming: correct

include/irr/core/math/aabb2dSIMDIntBase.h
```diff
@@ -0,0 +1,119 @@
...
+	VECTOR_TYPE box;
+};
+
+template<typename VECTOR_TYPE>
+class aabbSIMD : public aabbSIMDBase<aabbSIMD, VECTOR_TYPE> 
+{};
+
+
+template<>
+class aabbSIMD<core::vector4di32_SIMD> : public aabbSIMDBase<aabbSIMD,core::vector4di32_SIMD>
```
Przemog1: this class will be moved to aabbSIMDi.h ofc

devshgraphicsprogramming: you want to template on `core::vector2di32_SIMD`

Przemog1: right

include/irr/core/math/aabb2dSIMD.h
Przemog1: it can be shared between aabbSIMDi and aabbSIMDu only, aabbSIMDf needs also sign flip operation

include/irr/core/math/aabb2dSIMDIntBase.h
```diff
@@ -0,0 +1,119 @@
...
+	{
+		static_cast<CRTP<VECTOR_TYPE>*>(this)->addBox();
+	}
+
+	inline VECTOR_TYPE getExtent() const
+	{
+		return getMaxEdge() - getMinEdge();
+	}
+
+	inline VECTOR_TYPE getMinEdge() const
```
devshgraphicsprogramming: getMinEdge and getMaxEdge you'll actually have to specialize

include/irr/core/math/aabb2dSIMDIntBase.h
```diff
@@ -0,0 +1,119 @@
...
+	inline VECTOR_TYPE getMaxEdge() const
+	{
+		VECTOR_TYPE maxEdge = box.zwzw();
+		maxEdge.makeSafe2D();
+		return maxEdge;
+	}
+
+	inline CRTP<VECTOR_TYPE> reset(const VECTOR_TYPE& point) const
+	{
+		box = point.xyxy();
```
devshgraphicsprogramming: reset can be done as an move assignment (`operator=(&&)`) of a new box created from a point

include/irr/core/math/aabb2dSIMDIntBase.h
```diff
@@ -0,0 +1,119 @@
...
+public:
+	aabbSIMD(const core::vector4di32_SIMD& _box)
+		//:box(_box) - doesn't compile
+	{
+		this->box = _box;
+	}
+
+	bool isPointInside(const core::vector2di32_SIMD& point) const
+	{
+		//temporary solution
```
devshgraphicsprogramming: the unpack lo and unpack hi comparison will work here too

include/irr/core/math/aabb2diSIMD.h
```diff
@@ -16,68 +18,13 @@ class aabb2dSIMDi
 	inline explicit aabb2dSIMDi(const core::vector4di32_SIMD& _box)
 		: internalBoxRepresentation(core::vector4di32_SIMD(-_box.x, -_box.y, _box.z, _box.w)) {};
```
devshgraphicsprogramming, line 19: dont use the negative minimum convention for integers, due to range issues

include/irr/core/math/aabb2dSIMDIntBase.h
```diff
@@ -0,0 +1,119 @@
...
+			static_cast<const CRTP<VECTOR_TYPE>*>(this)->isPointInside(other.box.xwxx()) ||
+			static_cast<const CRTP<VECTOR_TYPE>*>(this)->isPointInside(other.box.zyxx());
+	}
+
+protected:
+	VECTOR_TYPE box;
+};
+
+template<typename VECTOR_TYPE>
+class aabbSIMD : public aabbSIMDBase<aabbSIMD, VECTOR_TYPE>
```
devshgraphicsprogramming: you're missing `2d` from the name

include/irr/core/math/aabb2dSIMDIntBase.h
```diff
@@ -0,0 +1,119 @@
...
+#include "irr/core/alloc/AlignedBase.h"
+#include "C:\IrrlichtBAW\IrrlichtBAW\include\vectorSIMD.h"
+
+#include <iostream>
+
+namespace irr { namespace core {
+
+//will modify this class so aabbSIMDf will be able do derive from it too
+template<template<typename> typename CRTP, typename VECTOR_TYPE>
+class aabbSIMDBase
```
devshgraphicsprogramming: you're missing `2d` from the name

include/irr/core/math/aabb2dSIMDIntBase.h
```diff
@@ -111,9 +147,83 @@ class aabbSIMD<core::vector4di32_SIMD> : public aabbSIMDBase<aabbSIMD,core::vect
...
+	otherwise false. */
+	inline bool isFullInside(const aabbSIMD<core::vector2df_SIMD>& other) const
+	{
+		core::vector4db_SIMD a = internalBoxRepresentation<= other.internalBoxRepresentation;
+		return a.allBits();
+	}
+
+};
+
+typedef aabbSIMD<core::vector2df_SIMD> aabb2dSIMDf;
```
devshgraphicsprogramming: lovely stuff @Przemog1

include/irr/core/math/aabb2diSIMD.h
Przemog1: i know, this class will be dumped anyway

include/irr/core/math/aabb2dSIMDIntBase.h
```diff
@@ -111,9 +147,83 @@ class aabbSIMD<core::vector4di32_SIMD> : public aabbSIMDBase<aabbSIMD,core::vect
...
+{
+public:
+	inline aabbSIMD()
+		//: internalBoxRepresentation(-FLT_MAX, -FLT_MAX, -FLT_MAX, -FLT_MAX)
+	{
+		internalBoxRepresentation = core::vectorSIMDf(-FLT_MAX, -FLT_MAX, -FLT_MAX, -FLT_MAX);
+	};
+
+	aabbSIMD(const core::vector2df_SIMD& _box)
+		//:inernalboxRepresentation(_box) - doesn't compile
```
Przemog1, line 123: why it doesn't compile?

include/irr/core/math/aabb2dSIMDIntBase.h
```diff
@@ -216,9 +175,80 @@ class aabbSIMD<core::vector2df_SIMD> : public aabbSIMDBase<aabbSIMD, core::vecto
...
+template<typename VECTOR_TYPE>
+class aabb2dSIMDInt : public aabbSIMDBase<aabb2dSIMDInt, VECTOR_TYPE>
+{
+	static_assert(
+		std::is_same<VECTOR_TYPE, core::vector2di32_SIMD>::value ||
+		std::is_same<VECTOR_TYPE, core::vector2du32_SIMD>::value,
+		"assertion failed: cannot use this type");
+
+public:
+	inline aabb2dSIMDInt()
```
Przemog1, line 190: only this function will vary between aabb2dSIMDi and aabb2dSIMDu i believe

include/irr/core/math/aabb2dSIMDIntBase.h
```diff
@@ -4,67 +4,87 @@
...
-		VECTOR_TYPE maxEdge = box.zwzw();
-		maxEdge.makeSafe2D();
-		return maxEdge;
-	}
-
-	inline CRTP<VECTOR_TYPE> reset(const VECTOR_TYPE& point) const
-	{
-		box = point.xyxy();
-		return *this;
+		CRTP<VECTOR_TYPE> pointBox(point.xyxy());
```
devshgraphicsprogramming: you don't need a swizzle your constructors will work with random stuff in zw

include/irr/core/math/aabb2dSIMDIntBase.h
```diff
@@ -4,67 +4,87 @@
...
+		internalBoxRepresentation= pointBox;
```
devshgraphicsprogramming: now you need an `std::move`

include/irr/core/math/aabb2dSIMDIntBase.h
```diff
@@ -13,9 +13,9 @@
 
 namespace irr { namespace core {
 
-//will modify this class so aabbSIMDf will be able do derive from it too
+
 template<template<typename> typename CRTP, typename VECTOR_TYPE>
-class aabbSIMDBase
+class aabbSIMDBase : public AlignedBase<_IRR_VECTOR_ALIGNMENT>
```
devshgraphicsprogramming: still missing `2d` in the name

include/irr/core/math/aabb2dSIMDIntBase.h
devshgraphicsprogramming: sneaky way of solving this ;)

include/irr/core/math/aabb2dSIMDIntBase.h
```diff
@@ -216,9 +175,80 @@ class aabbSIMD<core::vector2df_SIMD> : public aabbSIMDBase<aabbSIMD, core::vecto
...
+	static_assert(
+		std::is_same<VECTOR_TYPE, core::vector2di32_SIMD>::value ||
+		std::is_same<VECTOR_TYPE, core::vector2du32_SIMD>::value,
+		"assertion failed: cannot use this type");
+
+public:
+	inline aabb2dSIMDInt()
+	{
+		if constexpr (std::is_same<VECTOR_TYPE, core::vector2di32_SIMD>::value)
+			internalBoxRepresentation = core::vector4di32_SIMD(INT_MIN);
```
devshgraphicsprogramming: you need `(INT_MAX,INT_MAX,INT_MIN,INT_MIN)` I told you that beacuse of the ranges you won't be able to use the same "one SSE op" trick as for the `vectorSIMDf` based solution

include/irr/core/math/aabb2dSIMDIntBase.h
devshgraphicsprogramming: if you implement all the comparison operators in the vectorSIMD32* types then yes.

include/irr/core/math/aabb2dSIMDIntBase.h
```diff
@@ -216,9 +175,80 @@ class aabbSIMD<core::vector2df_SIMD> : public aabbSIMDBase<aabbSIMD, core::vecto
...
+
+		if constexpr (std::is_same<VECTOR_TYPE, core::vector2du32_SIMD>::value)
```
devshgraphicsprogramming: We don't quite support C++17, but we have out own workaround called `irr::static_if` if I remember correctly

include/irr/core/math/aabb2dSIMDIntBase.h
```diff
@@ -216,9 +175,80 @@ class aabbSIMD<core::vector2df_SIMD> : public aabbSIMDBase<aabbSIMD, core::vecto
...
+			internalBoxRepresentation = core::vector4du32_SIMD(0u);
```
devshgraphicsprogramming: (UINT_MAX,UINT_MAX,0u,0u)

include/irr/core/math/aabb2dSIMDIntBase.h
```diff
@@ -216,9 +175,80 @@ class aabbSIMD<core::vector2df_SIMD> : public aabbSIMDBase<aabbSIMD, core::vecto
...
+	}
+
+	inline aabb2dSIMDInt(const VECTOR_TYPE& _box)
```
devshgraphicsprogramming, line 206: `explicit`

include/irr/core/math/aabb2dSIMDIntBase.h
```diff
@@ -216,9 +175,80 @@ class aabbSIMD<core::vector2df_SIMD> : public aabbSIMDBase<aabbSIMD, core::vecto
...
+		core::vector2di32_SIMD maxEdge = internalBoxRepresentation.zwzw();
+		maxEdge.makeSafe2D();
+
+		return maxEdge;
+	} 
+
+	inline bool isPointInside(const VECTOR_TYPE& p) const
+	{
+		//need operator< and operator>
+		//but implementation will be the same for both core::vector2di_SIMD and core::vector2du_SIMD
```
devshgraphicsprogramming, line 236: great plan, remember to use `_mm_blend_*` from SSE4.2 (the same can be used to implement `core::mix` compatibile with GLSL that operates on a bvec)

include/irr/core/math/aabb2dSIMDIntBase.h
Przemog1: wtf i did xd

include/irr/core/math/aabb2dSIMDIntBase.h
Przemog1: whats the reason to not support c++17?

include/irr/core/math/aabb2dSIMDIntBase.h
devshgraphicsprogramming: actually if you did
```
all(unpacklo(internalBoxRepresentation,p) < mix(p,internalBoxRepresentation,bvec4(false,false,true,true)))
// all(  (i.min,p.xy) < (p.xy,i.max)  )
```
then you could get away just with `operator<`

include/irr/core/math/aabb2dSIMDIntBase.h
devshgraphicsprogramming: you have a typo (missing "t")?

include/irr/core/math/aabb2dSIMDIntBase.h
devshgraphicsprogramming: ?

include/irr/core/math/aabb2dSIMDIntBase.h
devshgraphicsprogramming: having to compile on old versions of GCC for edu.buildaworld.net
(we support c++17, but need fallbacks, `static_if` is a fallback for c++14)

include/irr/core/math/aabb2dSIMDIntBase.h
```diff
@@ -0,0 +1,259 @@
...
+		CRTP<VECTOR_TYPE> pointBox(point.xyxy());
+		internalBoxRepresentation= pointBox;
+	}
+
+	//! Repairs the box.
+	/** Necessary if for example MinEdge and MaxEdge are swapped. */
+	inline CRTP<VECTOR_TYPE> repair()
+	{
+		//
+		//but will it work for aabbSIMDf for sure?
```
devshgraphicsprogramming: sure because `get*Edge` returns a normal natural 2d point and `reset` + `addPoint` should work on it just fine

include/irr/core/math/aabb2dSIMDIntBase.h
```diff
@@ -0,0 +1,259 @@
...
+	\param y Y coord of the point. */
+	inline void reset(const VECTOR_TYPE& point)
+	{
+		CRTP<VECTOR_TYPE> pointBox(point.xyxy());
+		internalBoxRepresentation= pointBox;
+	}
+
+	//! Repairs the box.
+	/** Necessary if for example MinEdge and MaxEdge are swapped. */
+	inline CRTP<VECTOR_TYPE> repair()
```
devshgraphicsprogramming: void

include/irr/core/math/aabb2dSIMDIntBase.h
```diff
@@ -0,0 +1,259 @@
...
+	{
+		//
+		//but will it work for aabbSIMDf for sure?
+		//
+		//
+		reset(DWN_CAST_THIS_PTR->getMinEdge());
+		addPoint(DWN_CAST_THIS_PTR->getMaxEdge());
+
+		return *this;
```
devshgraphicsprogramming: no return

include/irr/core/math/aabb2dSIMDIntBase.h
```diff
@@ -0,0 +1,259 @@
...
+	\return True if box is empty, else false. */
+	inline bool isEmpty() const
+	{
+		return return (getMinEdge() == getMaxEdge()).all();;
+	}
+
+	//! Get the surface area of the box in squared units
+	inline float getArea() const
+	{
+		VECTOR_TYPE a = DWN_CAST_THIS_PTR->internalBoxRepresentation.getMaxEdge().x - DWN_CAST_THIS_PTR->internalBoxRepresentation.getMinEdge().x;
```
devshgraphicsprogramming: something went wrong here.
```
VECTOR_TYPE extent = DWN_CAST_THIS_PTR->getExtent();
return extent.x*extent.y;
```

devshgraphicsprogramming: made a typo, updated

include/irr/core/math/aabb2dSIMDIntBase.h
```diff
@@ -0,0 +1,259 @@
...
+		return (internalBoxRepresentation.zwzw() - internalBoxRepresentation) / VECTOR_TYPE(2);
+	}
+
+	//! Determines if the axis-aligned box intersects with another axis-aligned box.
+	/** \param other: Other box to check a intersection with.
+	\return True if there is an intersection with the other box,
+	otherwise false. */
+	bool intersectsWithBox(const CRTP<VECTOR_TYPE>& other) const
+	{
+		return
```
devshgraphicsprogramming: remember about the stack overflow algorithm

include/irr/core/math/aabb2dSIMDIntBase.h
```diff
@@ -0,0 +1,259 @@
...
+class aabbSIMD<core::vector2df_SIMD> : public aabbSIMDBase<aabbSIMD, core::vector2df_SIMD>
+{
+public:
+	inline aabbSIMD()
+		//: internalBoxRepresentation(-FLT_MAX, -FLT_MAX, -FLT_MAX, -FLT_MAX)
+	{
+		internalBoxRepresentation = core::vectorSIMDf(-FLT_MAX, -FLT_MAX, -FLT_MAX, -FLT_MAX);
+	};
+
+	aabbSIMD(const core::vector2df_SIMD& _box)
```
devshgraphicsprogramming: explicit

include/irr/core/math/aabb2diSIMD.h
```diff
@@ -0,0 +1,102 @@
...
+
+#include "irr/core/math/irrMath.h"
+#include "irr/core/memory/memory.h"
+#include "irr/core/alloc/AlignedBase.h"
+
+#include "irr/core/math/aabb2dSIMDIntBase.h"
+
+namespace irr { namespace core {
+
+class aabb2dSIMDi// : public aabb2dSIMDIntBase<aabb2dSIMDi<core::vector4di32_SIMD>>
```
devshgraphicsprogramming, line 12: this class still around?

include/irr/core/math/aabb2dSIMD.h
devshgraphicsprogramming: Wrong.

aabb2dSIMDf will perform sign flip when calling a constructor from a single point

include/irr/core/math/aabb2diSIMD.h
Przemog1: no, will clean it up later

include/irr/core/math/aabb2dSIMDIntBase.h
Przemog1: nvm

include/irr/core/math/aabb2dSIMDIntBase.h
Przemog1: thats the plan, but implementing those will take some time

include/irr/core/math/aabb2dSIMDIntBase.h
Przemog1: why not ordinary copy assigment? We are not moving anything

include/irr/core/math/aabb2dSIMDIntBase.h
Przemog1: I have some concerns about how will it work for aabbSIMDf

include/irr/core/math/aabb2dSIMDIntBase.h
Przemog1: sure

include/irr/core/math/aabb2dSIMDIntBase.h
Przemog1: description of this function says: 
/** Necessary if for example MinEdge and MaxEdge are swapped. */

so in this situation if minEdge and maxEdge are switched, then aabbSIMDf version of getMinEdge() will return maxEdge with fliped sign

include/irr/core/math/aabb2dSIMDIntBase.h
Crisspl: @devshgraphicsprogramming we are allowed to use `if constexpr` now?

devshgraphicsprogramming: no, see the other comment I left about `static_if`

include/irr/core/math/aabb2dSIMDIntBase.h
devshgraphicsprogramming: you could move the temporary `pointBox` into `this`

now what you have is incompileable, but I can assume you meant to copy the internal rep of the temporary `pointBox` into `this->internal....` which will get optimized by the compiler only if you're lucky.

include/irr/core/math/aabb2dSIMDIntBase.h
devshgraphicsprogramming: yes, but you need to anticipate `repair` being called on a totally valid aabb (that does not need switching)

devshgraphicsprogramming:
@Przemog1 , btw you need to add `IRR_FORCE_EBO` to all classes which are "empty" (don't have members) otherwise their sizes will be messed up

like this
https://github.com/buildaworldnet/IrrlichtBAW/blob/08269e8af602b0622b2b48cad785d0997f0b4551/include/irr/core/BaseClasses.h#L21

Its a very interesting MSVC bug regarding Empty Base Optimization that you can read about ;)

Przemog1:
now all I need to do is better algorithm for box intersection test

Przemog1:
Also to finish this task I will need operator<, operator>, unpackLo and unpackHi functions for SIMD vectors

include/irr/core/math/aabb2dSIMDIntBase.h
```diff
@@ -5,17 +5,16 @@
...
-#define DWN_CAST_THIS_PTR static_cast<const CRTP<VECTOR_TYPE>*>(this)
+#define DWN_CAST_THIS_PTR	static_cast<const CRTP<VECTOR_TYPE>*>(this)
 #define SIGN_FLIP_MASK_XY	core::vector3du32_SIMD(0x80000000, 0x80000000, 0x00000000, 0x00000000)
 
 namespace irr { namespace core {
 
 
 template<template<typename> typename CRTP, typename VECTOR_TYPE>
-class aabbSIMDBase : public AlignedBase<_IRR_VECTOR_ALIGNMENT>
+class aabb2dSIMDBase : public AlignedBase<_IRR_VECTOR_ALIGNMENT>
```
devshgraphicsprogramming: IRR_FORCE_EBO

include/irr/core/math/aabb2dSIMDIntBase.h
```diff
@@ -24,6 +23,9 @@ class aabbSIMDBase : public AlignedBase<_IRR_VECTOR_ALIGNMENT>
 		"assertion failed: cannot use this type");
 
 public:
+	aabb2dSIMDBase() = default;
```
devshgraphicsprogramming, line 28: make them protected, this way `aabb2dSIMDBase` will not be possible to instantiate

Przemog1: makes sense

include/irr/core/math/aabb2dSIMDIntBase.h
```diff
@@ -45,38 +47,31 @@ class aabbSIMDBase : public AlignedBase<_IRR_VECTOR_ALIGNMENT>
 	\param y Y coord of the point. */
 	inline void reset(const VECTOR_TYPE& point)
 	{
-		CRTP<VECTOR_TYPE> pointBox(point.xyxy());
-		internalBoxRepresentation= pointBox;
+		CRTP<VECTOR_TYPE> pointBox(point);
+		internalBoxRepresentation = pointBox.internalBoxRepresentation;
```
devshgraphicsprogramming: just use assignment

devshgraphicsprogramming:
```cpp
auto pseudoMin = core::unpacklo(internalRep,other.internalRep);
auto pseudoMax = core::unpackhi(internalRep,other.internalRep);
if (FLOATING_TYPE) // min <= max implies max-min >= 0
   return _mm_movemask_ps((pseudoMax+pseudoMin).getAsRegister())==0x0u;
else
   return pseudoMin<=pseudoMax;
```

include/irr/core/math/aabb2dSIMDIntBase.h
```diff
@@ -48,7 +49,7 @@ class aabb2dSIMDBase : public AlignedBase<_IRR_VECTOR_ALIGNMENT>
 	inline void reset(const VECTOR_TYPE& point)
 	{
 		CRTP<VECTOR_TYPE> pointBox(point);
-		internalBoxRepresentation = pointBox.internalBoxRepresentation;
+		*this = pointBox;
```
devshgraphicsprogramming, line 54: nope, you need to downcast the pointer and call the CRTP version of operator=

include/irr/core/math/aabb2dSIMDIntBase.h
```diff
@@ -102,7 +103,7 @@ template<typename VECTOR_TYPE>
 class IRR_FORCE_EBO aabb2dSIMD;
 
 template<>
-class aabb2dSIMD<core::vector2df_SIMD> : public aabb2dSIMDBase<aabb2dSIMD, core::vector2df_SIMD>
+class IRR_FORCE_EBO aabb2dSIMD<core::vector2df_SIMD> : public aabb2dSIMDBase<aabb2dSIMD, core::vector2df_SIMD>
```
devshgraphicsprogramming, line 113: aabb2SIMDBase decl also needs `IRR_FORCE_EBO`

Przemog1: i really need to read about it

include/irr/core/math/aabb2dSIMDIntBase.h
Przemog1: got it, just in case copy assigment is overrided

include/irr/core/math/aabb2dSIMDIntBase.h
devshgraphicsprogramming: basically put static_asserts that `sizeof(aabb2dSIMDf)==16u` and see what happens

include/vectorSIMD.h
```diff
@@ -970,6 +970,22 @@ namespace core
     inline vectorSIMDIntBase& vectorSIMDIntBase::operator|=(const vectorSIMDf &other) { _mm_store_si128((__m128i*)this,_mm_or_si128(getAsRegister(),_mm_castps_si128(other.getAsRegister()))); return *this;}
     inline vectorSIMDIntBase& vectorSIMDIntBase::operator^=(const vectorSIMDf &other) { _mm_store_si128((__m128i*)this,_mm_xor_si128(getAsRegister(),_mm_castps_si128(other.getAsRegister()))); return *this;}
 
+	inline core::vectorSIMDf unpackLo(const core::vectorSIMDf& v0, const core::vectorSIMDf& v1)
+	{
+		core::vectorSIMDf result;
+		*((__m128*)(result.pointer)) = _mm_unpacklo_ps(v0.getAsRegister(), v1.getAsRegister());
```
devshgraphicsprogramming: that's invalid, you need to use `_mm_store_ps`

Przemog1: I've changed that already, but it worked

devshgraphicsprogramming: its works because your compiler substitutes in `_mm_storeu_ps` silently

devshgraphicsprogramming:
are we done?

Przemog1:
If everything is alright then yes.
Need to add few comments and I think its time to suspend this task until SIMD vector classes rework.

PR #314, opened 2019-06-03
Fixed CMake for Smooth Normal Generators to compile properly.

khomsidakhun:


khomsidakhun:
Testing shit didn't mean to do it here.

PR #315, opened 2019-06-03
Fixed CMake for Smooth Normal Generators to compile properly.

khomsidakhun:


devshgraphicsprogramming:
Thanks a lot for the fix but this has been resolved in the sRGB and brdf_explorer branches, which I guess I should merge to master right now

PR #318, opened 2019-06-06
POSTPONED: Have all CEGUI deps build via CMake and come from submodules

devshgraphicsprogramming:
Still a recursive cmake configure and build call, but this is better than the previous solution (static linking on Linux!)

Problems are as follows:
1) `./ext/CEGUI/CMakeLists.txt` doesn't see the value of `$CEGUI_DEPS_INSTALL_PATH`
2) `_d` suffix does not get added to Debug builds of dependencies and Release builds probably overwrite them when installing

The result is that irr::ext::CEGUI doesn't build on Linux, on Windows it builds but doesn't link.

Cosmetic:
1) glew installs binary utils and shared libs, even though I tell it not to build those
2) pcre installs and builds the docs and man pages


devshgraphicsprogramming:
@Crisspl @Przemog1 if you can help fix in spare time (in July), that'd be great!

Issue #130, opened 2018-08-02
Add SPIR-V introspection and GLSL downcompile for OpenGL


devshgraphicsprogramming:
Use SPIRV-Cross
https://www.khronos.org/assets/uploads/developers/library/2016-vulkan-devday-uk/4-Using-spir-v-with-spirv-cross.pdf
https://github.com/KhronosGroup/SPIRV-Cross

to enable some introspection for asset::ISPIR_VShader

SPIRV-Cross does have some overhead, so introspection data should be gathered only on demand through `bool asset::ISPIR_VShader::initIntrospection()`

OpenGL 4.5 backend and below will need this to down-compile SPIR-V back to GLSL and will have to define their own workarounds to push_constants and descriptor sets as per
https://github.com/buildaworldnet/IrrlichtBAW/wiki/Vulkan-Limitations-due-to-OpenGL-Support

Depends on #129

devshgraphicsprogramming:
Superseded by 
https://docs.google.com/document/d/1WA8ctunP2IWM-sdi2D26c7q2FvixvEKLQ7sVo1czXqs/edit?usp=sharing

devshgraphicsprogramming:
Which is superseded by
https://repository.genmymodel.com/devshgraphicsprogramming/VulkanShaders

Crisspl:
I'll put these questions here

* what's the difference between ICPUShader and ICPUSpecializedShader? isn't specialized shader an ISPIR_VProgram + specialization data? Why isn't ICPUSpecializedShader derivative of ICPUShader then?
* Or is ICPUShader a collection of ICPUSpecializedShader-s? (i.e. whole shader program for example vs+fs) Is it complete program or just one shader/stage?
* So we want to support spir_v only as input for the engine (asset pipeline)? And optionally(?) VK GLSL (which gets compiled to spir_v) - that's what we'll use shaderc for? What about GL GLSL shader sources?
* GL GLSL (under COpenGLDriver) is compiled into machine code by downcompiling spir_v code (ISPIR_VProgram) into GLSL code (with spirv cross) and then put into OpenGL to compile, right?
* How should specialization be done? e.g.: Do introspection, change things in introspection data (bindings, wg size etc) and generate new shader based on modified introspection data?
* What is VkSpecializedMapEntry (ICPUSpecializedShader::mSpecializations); [Google doesn't know about this]? Actually what is this mSpecializations member?
* Why ISPIR_VProgram needs to be an interface?
[...]

Issue #129, opened 2018-08-02
Add an online GLSL to SPIR-V compiler


devshgraphicsprogramming:
We have a choice between shaderc and glslang.

However shaderc wraps around both glslang and SPIRV-Tools while providing `#include` functionality.

the asset::IAssetManager should load an asset::ISPIR_VProgram

manhnt9:
[Using Glslang as an Online Compiler for Vulkan](http://forestsharp.com/glslang-cpp/)

devshgraphicsprogramming:
Thanks for the link, it talks about shaderc too

I prefer shaderc because of the integration of SPIRV-Tools and the include stuff

devshgraphicsprogramming:
https://www.khronos.org/assets/uploads/developers/library/2016-vulkan-devday-uk/4-Using-spir-v-with-spirv-cross.pdf

[Under Vulkan]
It is paramount that the GLSL compiler reports all enabled Vulkan extensions in the form of '#define's to the shader (insert in the line just after '#version').

I.e. if the vulkan device was created with the supported extension `KHR_shader_subgroup_arithmetic` then the GLSL compiler must insert '#define GL_KHR_shader_subgroup_arithmetic' into every GLSL source before compilation to SPIRV.

This is so things like this can be done
```
#ifndef KHR_shader_subgroup_arithmetic
    #extension GL_NV_shader_thread_group: enable
#endif // KHR_shader_subgroup_arithmetic
```

devshgraphicsprogramming:
https://repository.genmymodel.com/devshgraphicsprogramming/VulkanShaders

devshgraphicsprogramming:
When we migrate to SPIR-V only our built-in material shaders will need converting to SPIR-V.

However they (and a bunch of other dependencies) should be embedded in the library and not depend on some file being found.
https://github.com/vector-of-bool/cmrc

devshgraphicsprogramming:
This has now turned into a duplicate of #130

PR #333, opened 2019-06-25
BRDF Explorer Alpha

devshgraphicsprogramming:


examples_tests/13.BRDF_Explorer/BRDFExplorerApp.cpp
```diff
@@ -230,29 +229,12 @@ float ReIoRfromF0andImIoR(float F0, float ImIoR)
...
-            source += "\treturn uRealIoR;";
-        else
-            source +=
-            R"(float F0 = texture(uIoRMap, texCoords).x;
-return vec3(IoRfromF0(REFLECTANCE_SCALE_FACTOR*F0*F0));
-)";
-        source += "\n}\n";
-
-        source += "vec3 getIoR_conductor(in vec2 texCoords) {\n";
+        source += "vec3 getReflectance(in vec2 texCoords) {\n";
```
devshgraphicsprogramming, line 241: you're not returning reflectance, this returns IoR when not using texture?

examples_tests/13.BRDF_Explorer/BRDFExplorerApp.cpp
```diff
@@ -395,15 +381,32 @@ void main() {
 		const float metallic = getMetallic(texCoords);
 		const vec3 baseColor = getAlbedo(texCoords);
 		const float ao = getAO(texCoords);
-
-        vec3 IoR_dielectric = getIoR_dielectric(texCoords);
-        mat2x3 IoR_conductor = mat2x3(getIoR_conductor(texCoords), uImagIoR);
+)"
+    +
+    [&_params] {
+    if (!_params.constantRI)
```
devshgraphicsprogramming, line 425: !?

Crisspl: hm?

examples_tests/13.BRDF_Explorer/BRDFExplorerApp.cpp
Crisspl: yea, there's a comment in code you commented below

examples_tests/13.BRDF_Explorer/BRDFExplorerApp.cpp
devshgraphicsprogramming: what is going on here?

examples_tests/13.BRDF_Explorer/BRDFExplorerApp.cpp
devshgraphicsprogramming: so why is it named like that?

Crisspl: well I had to name it somehow and choose between the two

examples_tests/13.BRDF_Explorer/BRDFExplorerApp.cpp
Crisspl: real ior is got in different way depending on whether it comes from texture or constant. And this code is generated in inline lambda (so that I don't have to create extra string for this code, imo it's more legible this way)

examples_tests/13.BRDF_Explorer/createDerivMapFromBumpMap.cpp
```diff
@@ -0,0 +1,56 @@
+#include <irrlicht.h>
+#include "createComputeShader.h"
+#include "../source/Irrlicht/COpenGL2DTexture.h"
+#include "../source/Irrlicht/COpenGLDriver.h"
+#include "CComputeShaderManager.h"
+
+using namespace irr;
+
+video::IVirtualTexture* createDerivMapFromBumpMap(video::IVirtualTexture* _bumpMap, IrrlichtDevice* _device, float _heightFactor)
```
devshgraphicsprogramming: shouldn't you factor this out to engine core in `video` namespace?

examples_tests/13.BRDF_Explorer/createDerivMapFromBumpMap.cpp
```diff
@@ -0,0 +1,56 @@
...
+    video::IVideoDriver* driver = _device->getVideoDriver();
+
+    const uint32_t* derivMap_sz = _bumpMap->getSize();
+    video::ITexture* derivMap = driver->createGPUTexture(video::ITexture::ETT_2D, derivMap_sz, std::log2(std::max(derivMap_sz[0], derivMap_sz[1])), asset::EF_R8G8_SNORM);
+
+    {
+        video::STextureSamplingParams params;
+        params.UseMipmaps = 0;
+        params.MaxFilter = params.MinFilter = video::ETFT_LINEAR_NO_MIP;
+        params.TextureWrapU = params.TextureWrapV = video::ETC_CLAMP_TO_EDGE;
```
devshgraphicsprogramming: you need a choice (argument to the function) whether this will simulate repeating heightfield (`ETC_REPEAT`) or not

examples_tests/13.BRDF_Explorer/createDerivMapFromBumpMap.cpp
```diff
@@ -0,0 +1,56 @@
...
+
+video::IVirtualTexture* createDerivMapFromBumpMap(video::IVirtualTexture* _bumpMap, IrrlichtDevice* _device, float _heightFactor)
+{
+    asset::IAssetManager& assetMgr = _device->getAssetManager();
+    asset::IAssetLoader::SAssetLoadParams lparams;
+
+    video::IVideoDriver* driver = _device->getVideoDriver();
+
+    const uint32_t* derivMap_sz = _bumpMap->getSize();
+    video::ITexture* derivMap = driver->createGPUTexture(video::ITexture::ETT_2D, derivMap_sz, std::log2(std::max(derivMap_sz[0], derivMap_sz[1])), asset::EF_R8G8_SNORM);
```
devshgraphicsprogramming: I think you're forgetting a `ceilf` after the `log2`, and a +1 for the numer of mipmaps

See the source of `COpenGLTexture`

examples_tests/13.BRDF_Explorer/createDerivMapFromBumpMap.cpp
```diff
@@ -0,0 +1,56 @@
...
+
+    {
+        video::STextureSamplingParams params;
+        params.UseMipmaps = 0;
+        params.MaxFilter = params.MinFilter = video::ETFT_LINEAR_NO_MIP;
+        params.TextureWrapU = params.TextureWrapV = video::ETC_CLAMP_TO_EDGE;
+        const_cast<video::COpenGLDriver::SAuxContext*>(reinterpret_cast<video::COpenGLDriver*>(driver)->getThreadContext())->setActiveTexture(7, _bumpMap, params);
+    }
+
+    GLuint deriv_map_gen_cs = CComputeShaderManager::getShader("../deriv_map_gen.comp");
```
devshgraphicsprogramming: if this is going to be in engine core, then you can build-in the shader

examples_tests/13.BRDF_Explorer/createDerivMapFromBumpMap.cpp
```diff
@@ -0,0 +1,56 @@
...
+    );
+
+    video::COpenGLExtensionHandler::extGlBindImageTexture(0u, 0u, 0, GL_FALSE, 0, GL_WRITE_ONLY, GL_R8); //unbind image
+    { //unbind texture
+        video::STextureSamplingParams params;
+        const_cast<video::COpenGLDriver::SAuxContext*>(reinterpret_cast<video::COpenGLDriver*>(driver)->getThreadContext())->setActiveTexture(7, nullptr, params);
+    }
+    video::COpenGLExtensionHandler::extGlUseProgram(previousProgram); //rebind previously bound program
+
+    derivMap->regenerateMipMapLevels();
```
devshgraphicsprogramming: acceptable

examples_tests/13.BRDF_Explorer/createDerivMapFromBumpMap.cpp
```diff
@@ -0,0 +1,56 @@
...
+    video::COpenGLExtensionHandler::extGlDispatchCompute((derivMap_sz[0]+15u) / 16u, (derivMap_sz[1]+15u) / 16u, 1u);
+    video::COpenGLExtensionHandler::extGlMemoryBarrier(
+        GL_TEXTURE_FETCH_BARRIER_BIT |
+        GL_SHADER_IMAGE_ACCESS_BARRIER_BIT |
+        GL_PIXEL_BUFFER_BARRIER_BIT |
+        GL_TEXTURE_UPDATE_BARRIER_BIT |
+        GL_FRAMEBUFFER_BARRIER_BIT
+    );
+
+    video::COpenGLExtensionHandler::extGlBindImageTexture(0u, 0u, 0, GL_FALSE, 0, GL_WRITE_ONLY, GL_R8); //unbind image
```
devshgraphicsprogramming: you can use the multibind with `nullptr` to unbind

examples_tests/13.BRDF_Explorer/deriv_map_gen.comp
```diff
@@ -0,0 +1,45 @@
...
+		ivec2 offset = ivec2(mix(ivec2(1, 0), ivec2(-1, 0), LC_IDX.x == 0));
+		smem[getAddr(LC_IDX+offset)] = texelFetch(bumpMapSampler, G_IDX+offset, 0).x * uHeightScaleFactor;
+	}
+	if (LC_IDX.y == 0 || LC_IDX.y == 15)
+	{
+		ivec2 offset = ivec2(mix(ivec2(0, 1), ivec2(0, -1), LC_IDX.y == 0));
+		smem[getAddr(LC_IDX+offset)] = texelFetch(bumpMapSampler, G_IDX+offset, 0).x * uHeightScaleFactor;
+	}
+	
+	barrier(); // AFAIK execution barrier implies memory barrier
```
devshgraphicsprogramming: does it?

examples_tests/13.BRDF_Explorer/BRDFExplorerApp.cpp
```diff
@@ -351,12 +357,20 @@ vec3 calculateSurfaceNormal(in vec3 position, in vec2 uv, in vec3 normal, in vec
 void main() {
     vec3 N = normalize(Normal);
     const vec2 texCoords = vec2(TexCoords.x, 1.0-TexCoords.y);
+    vec3 dp1 = dFdx( WorldPos );
+    vec3 dp2 = dFdy( WorldPos );
+    vec2 duv1 = dFdx( TexCoords );
+    vec2 duv2 = dFdy( TexCoords );
+    vec3 dp2perp = cross( dp2, N );
+    vec3 dp1perp = cross( N, dp1 );
+    vec3 T = normalize(dp2perp * duv1.x + dp1perp * duv2.x);
```
devshgraphicsprogramming: don't you already have TBN for derivative map?

examples_tests/13.BRDF_Explorer/BRDFExplorerApp.cpp
```diff
@@ -351,12 +357,20 @@ vec3 calculateSurfaceNormal(in vec3 position, in vec2 uv, in vec3 normal, in vec
 void main() {
     vec3 N = normalize(Normal);
     const vec2 texCoords = vec2(TexCoords.x, 1.0-TexCoords.y);
+    vec3 dp1 = dFdx( WorldPos );
+    vec3 dp2 = dFdy( WorldPos );
+    vec2 duv1 = dFdx( TexCoords );
+    vec2 duv2 = dFdy( TexCoords );
```
devshgraphicsprogramming: those 4 are redundant computation

examples_tests/13.BRDF_Explorer/BRDFExplorerApp.cpp
```diff
@@ -351,12 +357,20 @@ vec3 calculateSurfaceNormal(in vec3 position, in vec2 uv, in vec3 normal, in vec
...
+    vec3 T = normalize(dp2perp * duv1.x + dp1perp * duv2.x);
+    vec3 B = normalize(dp2perp * duv1.y + dp1perp * duv2.y);
 )"
     +
     (_params.derivMapIsPresent ?
 R"(
-    vec2 derivMapSz = vec2(textureSize(uDerivativeMap));
-    vec2 h_gradient = texture(uDerivativeMap, texCoords).xy*0.5*max(derivMapSz.x, derivMapSz.y);
+    vec2 derivMapSz = vec2(textureSize(uDerivativeMap, 0));
+    vec2 h_gradient = texture(uDerivativeMap, texCoords).xy*0.5*uHeightScaleFactor*max(derivMapSz.x, derivMapSz.y);
```
devshgraphicsprogramming: what is the 0.5 here for?

examples_tests/13.BRDF_Explorer/BRDFExplorerApp.cpp
```diff
@@ -374,19 +388,33 @@ R"(
...
         const float LplusV_lenSq = 2.0 + 2.0*LdotV;
         const float LplusV_rcpLen = inversesqrt(LplusV_lenSq);
         const float NdotH = max((NdotL + NdotV) * LplusV_rcpLen, 0.0);
         const float VdotH = max(LplusV_rcpLen + LplusV_rcpLen*LdotV, 0.0);
+        */
+        const float NdotH = max(dot(N, H), 0.0);
+        const float VdotH = max(dot(V, H), 0.0);
         NdotV = max(NdotV, 0.0);
+
+        const float TdotV = max(dot(T, V), 0.0);
```
devshgraphicsprogramming: why is max(,0.0) here?

examples_tests/13.BRDF_Explorer/BRDFExplorerApp.cpp
```diff
@@ -433,23 +461,23 @@ R"(
...
 
-    float ndf = GGXTrowbridgeReitz(a2, NdotH);
-    float geom = GGXSmith_wo_numerator(a2, NdotL, NdotV); // TODO: Correlated Smith!
+    float ndf = GGXBurleyAnisotropic(at, ab, TdotH, BdotH, NdotH);//GGXTrowbridgeReitz(a2, NdotH);
+    float geom = GGXSmithHeightCorrelated_aniso_wo_numerator(at, ab, TdotL, TdotV, BdotL, BdotV, NdotL, NdotV);//GGXSmithHeightCorrelated_wo_numerator(a2, NdotL, NdotV);
 
-    // Note: (4.0*NdotV*NdotL) denominator is cancelled by GGXSmith's numerator, thus the use of GGXSmith_wo_numerator()
-    return ndf*geom*out_fresnel;
+    // Note: (2.0*NdotV*NdotL) denominator is cancelled by GGXSmith's numerator, thus the use of GGXSmithHeightCorrelated_wo_numerator()
+    return 0.5*ndf*geom*out_fresnel;
```
devshgraphicsprogramming: Put the 0.5 in `GGXSmithHeightCorrelated_wo_numerator`

examples_tests/13.BRDF_Explorer/BRDFExplorerApp.cpp
```diff
@@ -374,19 +388,33 @@ R"(
...
         // dots with H identities taken from Earl Hammon's PBR Diffuse Lighting GDC17 lecture
+        /*
         const float LplusV_lenSq = 2.0 + 2.0*LdotV;
         const float LplusV_rcpLen = inversesqrt(LplusV_lenSq);
         const float NdotH = max((NdotL + NdotV) * LplusV_rcpLen, 0.0);
         const float VdotH = max(LplusV_rcpLen + LplusV_rcpLen*LdotV, 0.0);
+        */
+        const float NdotH = max(dot(N, H), 0.0);
+        const float VdotH = max(dot(V, H), 0.0);
         NdotV = max(NdotV, 0.0);
```
devshgraphicsprogramming: why clamp?

examples_tests/13.BRDF_Explorer/CBRDFBuiltinIncludeLoader.h
```diff
@@ -75,6 +83,38 @@ float GGXSmith_wo_numerator(in float a2, in float NdotL, in float NdotV)
...
+float GGXSmithHeightCorrelated(in float a2, in float NdotL, in float NdotV)
+{
+    float denom = NdotV*sqrt(a2 + (1.0 - a2)*NdotL*NdotL) + NdotL*sqrt(a2 + (1.0 - a2)*NdotV*NdotV);
+    return 2.0*NdotL*NdotV / denom;
+}
+
+float GGXSmithHeightCorrelated_wo_numerator(in float a2, in float NdotL, in float NdotV)
+{
+    float denom = NdotV*sqrt(a2 + (1.0 - a2)*NdotL*NdotL) + NdotL*sqrt(a2 + (1.0 - a2)*NdotV*NdotV);
+    return 1.0 / denom;
```
devshgraphicsprogramming: 0.5

examples_tests/13.BRDF_Explorer/CBRDFBuiltinIncludeLoader.h
```diff
@@ -75,6 +83,38 @@ float GGXSmith_wo_numerator(in float a2, in float NdotL, in float NdotV)
...
+float GGXSmithHeightCorrelated_approx(in float a, in float NdotL, in float NdotV)
+{
+    float num = 2.0*NdotL*NdotV;
+    return num / mix(num, NdotL+NdotV, a);
+}
+
+// Note a, not a2!
+float GGXSmithHeightCorrelated_approx_wo_numerator(in float a, in float NdotL, in float NdotV)
+{
+    return 1.0 / mix(2.0*NdotL*NdotV, NdotL+NdotV, a);
```
devshgraphicsprogramming: 0.5

examples_tests/13.BRDF_Explorer/CBRDFBuiltinIncludeLoader.h
```diff
@@ -75,6 +83,38 @@ float GGXSmith_wo_numerator(in float a2, in float NdotL, in float NdotV)
...
+}
+
+float GGXSmithHeightCorrelated_aniso_wo_numerator(in float at, in float ab, in float TdotL, in float TdotV, in float BdotL, in float BdotV, in float NdotL, in float NdotV)
+{
+    float Vterm = NdotL * length(vec3(at*TdotV, ab*BdotV, NdotV));
+    float Lterm = NdotV * length(vec3(at*TdotL, ab*BdotL, NdotL));
+    return 1.0 / (Vterm + Lterm);
```
devshgraphicsprogramming: 0.5

examples_tests/13.BRDF_Explorer/BRDFExplorerApp.cpp
```diff
@@ -312,7 +318,7 @@ IncludeHandler->getIncludeStandard("irr/builtin/glsl/brdf/specular/fresnel/fresn
 +
 R"(
 float diffuse(in float a2, in vec3 N, in vec3 L, in vec3 V, in float NdotL, in float NdotV);
-vec3 specular(in float a2, in float NdotL, in float NdotV, in float NdotH, in float VdotH, in mat2x3 ior, in float metallic, out vec3 out_fresnel);
+vec3 specular(in float a2, in float at, in float ab, in float NdotL, in float NdotV, in float NdotH, in float VdotH, in float TdotV, in float TdotL, in float BdotV, in float BdotL, in float TdotH, in float BdotH, in mat2x3 ior, in float metallic, out vec3 out_fresnel);
```
devshgraphicsprogramming: There's absolutely no need for a `VdotH` since `NdotH == VdotH`

examples_tests/13.BRDF_Explorer/BRDFExplorerApp.cpp
```diff
@@ -374,19 +388,33 @@ R"(
 		// there are better identities to get all of these
 		const vec3 V = normalize(uEyePos - WorldPos);
 		const vec3 L = normalize(relLightPos);
```
devshgraphicsprogramming: optimize this, you already calculate the inverseqsrt of `relLightPosLen2` above!

examples_tests/13.BRDF_Explorer/BRDFExplorerApp.cpp
```diff
@@ -374,19 +388,33 @@ R"(
...
 
         // dots with H identities taken from Earl Hammon's PBR Diffuse Lighting GDC17 lecture
+        /*
         const float LplusV_lenSq = 2.0 + 2.0*LdotV;
         const float LplusV_rcpLen = inversesqrt(LplusV_lenSq);
         const float NdotH = max((NdotL + NdotV) * LplusV_rcpLen, 0.0);
         const float VdotH = max(LplusV_rcpLen + LplusV_rcpLen*LdotV, 0.0);
+        */
+        const float NdotH = max(dot(N, H), 0.0);
+        const float VdotH = max(dot(V, H), 0.0);
```
devshgraphicsprogramming: no `max` necessary, its impossible to have a view vector at an obtuse angle to the half vector!

Assuming normalized V and L

H = (V+L)/Positive_Divisor
VdotH = (1+VdotL)/Positive_Divisor
VdotL >= -1.0

imply VdotH >=0.0

examples_tests/13.BRDF_Explorer/CBRDFBuiltinIncludeLoader.h
```diff
@@ -48,6 +48,14 @@ float GGXTrowbridgeReitz(in float a2, in float NdotH)
     return a2 / (3.14159265359 * denom*denom);
 }
 
+float GGXBurleyAnisotropic(float at, float ab, float TdotH, float BdotH, float NdotH) {
```
devshgraphicsprogramming: This is correct

examples_tests/13.BRDF_Explorer/BRDFExplorerApp.cpp
```diff
@@ -433,23 +461,23 @@ R"(
...
 
-    // Note: (4.0*NdotV*NdotL) denominator is cancelled by GGXSmith's numerator, thus the use of GGXSmith_wo_numerator()
-    return ndf*geom*out_fresnel;
+    // Note: (2.0*NdotV*NdotL) denominator is cancelled by GGXSmith's numerator, thus the use of GGXSmithHeightCorrelated_wo_numerator()
+    return 0.5*ndf*geom*out_fresnel;
 }
 
 vec3 Fresnel_combined(in mat2x3 ior, in float cosTheta, in float metallic) {
-    bvec3 is_inf = isinf(ior[0]);
+    bvec3 is_inf = isinf(ior[0]*ior[0] + ior[1]*ior[1]);
```
devshgraphicsprogramming: don't use isinf, use `<FLT_MAX` and swap the order of arguments in the `mix`

examples_tests/13.BRDF_Explorer/CBRDFBuiltinIncludeLoader.h
devshgraphicsprogramming: Although I would pass `a2` and `anisotropy` (`uAnisotropy`) to this function.

This way you could optimize the computation like this
```cpp
float antiAniso = 1.0-anisotropy;
float atab = a2*antiAniso; //old a2
float anisoTdotH = antiAniso*TdotH;
float anisoNdotH = antiAniso*NdotH;
float w2 = antiAniso/(BdotH*BdotH+anisoTdotH*anisoTdotH+anisoNdotH*anisoNdotH*a2);
return w2*w2*atab / PI;
```

Double check this, but it would be nice because no sqrt needed

examples_tests/13.BRDF_Explorer/BRDFExplorerApp.cpp
```diff
@@ -433,23 +461,23 @@ R"(
         else source += "\treturn 0.0;";
         source += "\n}\n";
         source += 
-R"(vec3 specular(in float a2, in float NdotL, in float NdotV, in float NdotH, in float VdotH, in mat2x3 ior, in float metallic, out vec3 out_fresnel) {
+R"(vec3 specular(in float a2, in float at, in float ab, in float NdotL, in float NdotV, in float NdotH, in float VdotH, in float TdotV, in float TdotL, in float BdotV, in float BdotL, in float TdotH, in float BdotH, in mat2x3 ior, in float metallic, out vec3 out_fresnel) {
 	//assert(NdotL>FLT_MIN);
     out_fresnel = Fresnel_combined(ior, VdotH, metallic);
     if (NdotV<FLT_MIN)
         return vec3(0.0);
     if (a2<FLT_MIN)
```
devshgraphicsprogramming: it needs to check for `a2*(1.0-uAnisotropy)`

examples_tests/13.BRDF_Explorer/BRDFExplorerApp.cpp
```diff
@@ -374,19 +388,33 @@ R"(
...
+        const float TdotV = max(dot(T, V), 0.0);
+        const float TdotL = max(dot(T, L), 0.0);
+        const float BdotV = max(dot(B, V), 0.0);
+        const float BdotL = max(dot(B, L), 0.0);
+        const float TdotH = max(dot(T, H), 0.0);
+        const float BdotH = max(dot(B, H), 0.0);
 		// identity comment end (but also do you need to clamp all of them?)
 
 		const float a2 = getRoughness(texCoords);
+        const float at = sqrt(a2);
```
devshgraphicsprogramming, line 418: see my comment about the NDF

examples_tests/13.BRDF_Explorer/BRDFExplorerApp.cpp
devshgraphicsprogramming: that happens in your derivative mapping code

examples_tests/13.BRDF_Explorer/BRDFExplorerApp.cpp
```diff
@@ -351,6 +357,14 @@ vec3 calculateSurfaceNormal(in vec3 position, in vec2 uv, in vec3 normal, in vec
...
     vec3 N = normalize(Normal);
     const vec2 texCoords = vec2(TexCoords.x, 1.0-TexCoords.y);
+    vec3 dp1 = dFdx( WorldPos );
+    vec3 dp2 = dFdy( WorldPos );
+    vec2 duv1 = dFdx( TexCoords );
+    vec2 duv2 = dFdy( TexCoords );
+    vec3 dp2perp = cross( dp2, N );
+    vec3 dp1perp = cross( N, dp1 );
+    vec3 T = normalize(dp2perp * duv1.x + dp1perp * duv2.x);
+    vec3 B = normalize(dp2perp * duv1.y + dp1perp * duv2.y);
```
devshgraphicsprogramming: Because of texture warping on curved surfaces you can bet on TBN not being an orthogonal frame.
This will fuck up all the lighting computation that needs to know about tangents.

So maybe make `vec3 B = cross(N,T);` (don't normalize, its redundant)

examples_tests/13.BRDF_Explorer/CBRDFBuiltinIncludeLoader.h
```diff
@@ -61,26 +69,101 @@ float _GGXSmith_G1_(in float a2, in float NdotX)
...
+   vec3 Rp = Rs * (t3 - t4) / (t3 + t4);
+
+   return 0.5 * (Rp + Rs);
+}
+float Fresnel_dielectric(in float Eta, in float CosTheta)
+{
+   float SinTheta2 = 1.0 - CosTheta * CosTheta;
+
+   float t0 = sqrt(1.0 - (SinTheta2 / (Eta * Eta)));
+   float t1 = Eta * t0;
```
devshgraphicsprogramming, line 159: We need Fresnel Diffuse Multiscatter Constant Polynomial Approximation in terms of `t1`

examples_tests/13.BRDF_Explorer/BRDFExplorerApp.cpp
devshgraphicsprogramming: why still clamp `NdotV` ?

examples_tests/13.BRDF_Explorer/BRDFExplorerApp.cpp
```diff
@@ -398,18 +413,18 @@ R"(
         const float LplusV_lenSq = 2.0 + 2.0*LdotV;
```
devshgraphicsprogramming, line 404: are you going to use those identities?

examples_tests/13.BRDF_Explorer/BRDFExplorerApp.cpp
```diff
@@ -340,54 +340,69 @@ float applyChainRule(in vec2 h_gradient, in vec2 dUVd_)
...
+	vec2 dUVdx = dUVd_.xy;
+	vec2 dUVdy = dUVd_.zw;
  
     float dhdx = applyChainRule(h_gradient, dUVdx);
     float dhdy = applyChainRule(h_gradient, dUVdy);
  
     return perturbNormal(normal, dpdx, dpdy, dhdx, dhdy);
 }
 
+float calcDiffuseCorrectionFactor_(float n)
```
devshgraphicsprogramming: you know you could just do this whole function in `vec3` then if someone wants the `float` version then they just put `.x` at the end ;)

also call it `calcDiffuseFresnelCorrectionFactor`

examples_tests/13.BRDF_Explorer/BRDFExplorerApp.cpp
```diff
@@ -340,54 +340,69 @@ float applyChainRule(in vec2 h_gradient, in vec2 dUVd_)
 }
 
 // Calculate the surface normal using the uv-space gradient (dhdu, dhdv)
-vec3 calculateSurfaceNormal(in vec3 position, in vec2 uv, in vec3 normal, in vec2 h_gradient)
+vec3 calculateSurfaceNormal(in vec3 normal, in vec2 h_gradient, in vec3 dpdx, in vec3 dpdy, in vec4 dUVd_)
```
devshgraphicsprogramming: why did you feel the need to pack the texture coord derivatives in a vec4 ?

examples_tests/13.BRDF_Explorer/CBRDFBuiltinIncludeLoader.h
devshgraphicsprogramming: @Crisspl wanna try ?

examples_tests/13.BRDF_Explorer/BRDFExplorerApp.cpp
devshgraphicsprogramming: @Crisspl ?

examples_tests/13.BRDF_Explorer/deriv_map_gen.comp
devshgraphicsprogramming: I think its best to put the shared memory barrier in, I've read the Vulkan Memory Model Appendix and I think execution synchronisation and availability+visibility of writes are two separate things (although visibility/availability requires execution synch)

examples_tests/13.BRDF_Explorer/createDerivMapFromBumpMap.cpp
devshgraphicsprogramming: @Crisspl ?

examples_tests/13.BRDF_Explorer/createDerivMapFromBumpMap.cpp
devshgraphicsprogramming: @Crisspl ?

devshgraphicsprogramming:
@Crisspl you should merge my master branch into yours to get latest CMake+submodule fixes.

examples_tests/13.BRDF_Explorer/BRDFExplorerApp.cpp
Crisspl: they were in use until i needed TdotH or BdotH (so now i need H and not only N,V dots with H)

examples_tests/13.BRDF_Explorer/BRDFExplorerApp.cpp
Crisspl: i cant because theres branch and i have no guarantee all of them will be on the same side of 1

examples_tests/13.BRDF_Explorer/BRDFExplorerApp.cpp
Crisspl: less parameters

devshgraphicsprogramming: its confusing, also it won't help anything.

examples_tests/13.BRDF_Explorer/BRDFExplorerApp.cpp
```diff
@@ -579,8 +579,15 @@ BRDFExplorerApp::BRDFExplorerApp(IrrlichtDevice* device, irr::scene::ICameraScen
         ShaderManager(new CShaderManager(Driver->getGPUProgrammingServices(), device->getIncludeHandler(), GUIState, LightAnimData, Camera)),
         DerivativeMapManager(new CDerivativeMapManager(device))
 {
+    {//default mesh
     auto cpumesh = AssetManager.getGeometryCreator()->createSphereMesh(10.f, 64u, 64u);
```
devshgraphicsprogramming: ok, but put cerberus mesh+textures into media repo

examples_tests/13.BRDF_Explorer/BRDFExplorerApp.cpp
devshgraphicsprogramming: for such short functions you should aim for branchless (every branch generates at least 3 instructions + each branch could be executed, a mask generates 1), also you have eta2 laying around (by necessity you alwasy have it, so use it).

Yes DIV is hundreds of cycles, but I'm hoping for a reciprocal instruction to be generated.

Here's an implementation (check it)
```cpp
float calcDiffuseCorrectionFactor(in float n, in float n2)
{
   //assert(n*n==n2);
   bvec3 TIR = n<1.0;
   float invdenum = mix(1.0,1.0/(n2*n2*(554.33 - 380.7*n)),TIR);
   float num = n*mix(0.1921156102251088,n*298.25 - 261.38*n2 + 138.43,TIR);
   num += mix(0.8078843897748912,-1.67,TIR);
   return
}
```

examples_tests/13.BRDF_Explorer/BRDFExplorerApp.cpp
devshgraphicsprogramming: any vector, lets call it `K` can have `KdotH` calculated from identities

dot(K,H) = dot(K,L+V)/length(L+V) = (dot(K,L)+dot(K,V))*LplusV_rcpLen = (KdotL+KdotV)*LplusV_rcpLen 

You actually happen to have all TdotV,TdotL,BdotB,BdotL needed

devshgraphicsprogramming:
@Crisspl There's still about 8 unresolved conversations remaining in the PR.

examples_tests/13.BRDF_Explorer/CBRDFBuiltinIncludeLoader.h
devshgraphicsprogramming: @Crisspl ?

include/irr/asset/IAssetManager.h
```diff
@@ -171,6 +171,7 @@ namespace asset
             std::string filename = _file ? _file->getFileName().c_str() : _supposedFilename;
             io::IReadFile* file = _override->getLoadFile(_file, filename, ctx, _hierarchyLevel);
             filename = file ? file->getFileName().c_str() : _supposedFilename;
+            // TODO what should happen here (after `filename = ...;`)? Open new file if previously it was nullptr?
```
devshgraphicsprogramming: did our discord convo resolve this question for you or not?

examples_tests/13.BRDF_Explorer/createDerivMapFromBumpMap.cpp
devshgraphicsprogramming: @Crisspl ?

examples_tests/13.BRDF_Explorer/createDerivMapFromBumpMap.cpp
devshgraphicsprogramming: @Crisspl ?

examples_tests/13.BRDF_Explorer/createDerivMapFromBumpMap.cpp
devshgraphicsprogramming: @Crisspl ?

examples_tests/13.BRDF_Explorer/createDerivMapFromBumpMap.cpp
devshgraphicsprogramming: @Crisspl ?

examples_tests/13.BRDF_Explorer/deriv_map_gen.comp
devshgraphicsprogramming: @Crisspl ?

examples_tests/13.BRDF_Explorer/BRDFExplorerApp.cpp
devshgraphicsprogramming: @Crisspl  ?

Crisspl: not there already, resolved

examples_tests/13.BRDF_Explorer/createDerivMapFromBumpMap.cpp
Crisspl: what do you mean by build-in? If it's gonna be in engine core then i think it must be in its own class so that shader program can be deleted in the class's destructor

examples_tests/13.BRDF_Explorer/deriv_map_gen.comp
Crisspl: i remember that while you were reading vulkan spec some long time ago you found out that exec barrier implies memory barrier, you tried it out and it was working and saving quite nice % of performance. So i take it it's not true, at lest in opengl?

include/irr/asset/IAssetManager.h
Crisspl: resolved

examples_tests/13.BRDF_Explorer/createDerivMapFromBumpMap.cpp
Crisspl: resloved

examples_tests/13.BRDF_Explorer/createDerivMapFromBumpMap.cpp
Crisspl: resolved

examples_tests/13.BRDF_Explorer/createDerivMapFromBumpMap.cpp
Crisspl: resolved

examples_tests/13.BRDF_Explorer/createDerivMapFromBumpMap.cpp
Crisspl: resolved

examples_tests/13.BRDF_Explorer/deriv_map_gen.comp
Crisspl: resolved
but give me an answer for the question :)

examples_tests/13.BRDF_Explorer/CBRDFBuiltinIncludeLoader.h
Crisspl: left TODO
i think your code is missing `a2` factor and then it's ok

Crisspl: but i'll check again tomorrow to be sure...

devshgraphicsprogramming:
Cool but you didn't include `CDerivativeMapper.h` in any other include so user won't be able to use it by just including `irrlicht.h`

src/irr/video/CDerivativeMapCreator.cpp
```diff
@@ -0,0 +1,155 @@
...
+
+CDerivativeMapCreator::CDerivativeMapCreator(video::IVideoDriver* _driver) : m_driver(_driver)
+{
+    m_deriv_map_gen_cs = this->createComputeShader(DERIV_MAP_FROM_BUMP_MAP_CS_SRC);
+}
+
+video::IVirtualTexture* CDerivativeMapCreator::createDerivMapFromBumpMap(video::IVirtualTexture* _bumpMap, float _heightFactor, bool _texWrapRepeat) const
+{
+    const uint32_t* derivMap_sz = _bumpMap->getSize();
+    video::ITexture* derivMap = m_driver->createGPUTexture(video::ITexture::ETT_2D, derivMap_sz, 1u+uint32_t(std::floor(std::log2(float(core::max_(derivMap_sz[0], derivMap_sz[1]))))), asset::EF_R8G8_SNORM);
```
devshgraphicsprogramming, line 122: how does this resolve 
https://github.com/buildaworldnet/IrrlichtBAW/pull/333#discussion_r298715523 
?

examples_tests/13.BRDF_Explorer/deriv_map_gen.comp
devshgraphicsprogramming: noooo.... I read in the Vulkan spec that exec barrier implies only exec barrier and memory barrier is needed to ensure transactions are visible and available.

~I did however read that `coherent` qualified variables do not need `memoryBarrier` because they are well, coherent.~

devshgraphicsprogramming: https://www.khronos.org/opengl/wiki/Compute_Shader#Shared_memory_coherency

> The usual set of memory barriers is available to compute shaders, but they also have access to memoryBarrierShared(); this barrier is specifically for shared variable ordering. groupMemoryBarrier() acts like memoryBarrier(), ordering memory writes for all kinds of variables, but it only orders read/writes for the current work group.

> While all invocations within a work group are said to execute "in parallel", that doesn't mean that you can assume that all of them are executing in lock-step. If you need to ensure that an invocation has written to some variable so that you can read it, you need to synchronize execution with the invocations, not just issue a memory barrier (you still need the memory barrier though).

devshgraphicsprogramming: Basically
`memoryBarrierShared` < `groupMemoryBarrier` < `memoryBarrier`

Also 
> After ensuring ordering, the other element that is needed for visibility is special GLSL syntax. The image or buffer variable being written to and read from must be qualified by the coherent qualifier. Note that both the writer and reader shaders must qualify their variables properly; otherwise, nothing is guaranteed.

> Qualifying the variable with coherent alone is not enough however. You also need to use a memory barrier;

So basically you need both `coherent` and a memory-barrier.

Unless you are doing *atomic* operations *only* on *coherent* qualified variables, then no memory barrier is necessary.

examples_tests/13.BRDF_Explorer/CBRDFBuiltinIncludeLoader.h
devshgraphicsprogramming: doesn't `float atab = a2*antiAniso; //old a2` do the job? my `atab` is your `a2`

examples_tests/13.BRDF_Explorer/BRDFExplorerApp.cpp
```diff
@@ -548,21 +548,22 @@ class CDerivativeMapManager
...
-    IrrlichtDevice* Device;
+    irr::video::CDerivativeMapCreator* DerivMapCreator;
 
 public:
-    CDerivativeMapManager(IrrlichtDevice* _device) : Device{_device} {}
+    CDerivativeMapManager(IrrlichtDevice* _device) : DerivMapCreator(new irr::video::CDerivativeMapCreator(_device->getVideoDriver())) {}
     ~CDerivativeMapManager() {
         for (auto& t : DerivMaps)
             t.second->drop();
+        DerivMapCreator->drop();
```
devshgraphicsprogramming: this sort of thing makes  me think we should create a default one in the `IVideoDriver` and just do a getter so there are not 10s of the same thing instantiated.

source/Irrlicht/COpenGLExtensionHandler.cpp
```diff
@@ -977,7 +977,8 @@ void COpenGLExtensionHandler::initExtensions(bool stencilBuffer)
 
     num=0;
     glGetIntegerv(GL_MAX_COMBINED_TEXTURE_IMAGE_UNITS, &num);
-	MaxTextureUnits = core::min_(static_cast<uint8_t>(num), static_cast<uint8_t>(MATERIAL_MAX_TEXTURES));
+    //can we get rid of that? I don't really see reason for this clamp at this level of abstraction
```
devshgraphicsprogramming, line 980: SMaterial only has `MATERIAL_MAX_TEXTURES` slots for now, once we solve #129 #130 we can remove this limit.

examples_tests/13.BRDF_Explorer/CBRDFBuiltinIncludeLoader.h
Crisspl: no, i meant something different.
But seems like it's ok, it doesn't need anything

src/irr/video/CDerivativeMapCreator.cpp
```diff
@@ -0,0 +1,155 @@
...
+
+    video::COpenGLDriver* gldriver = static_cast<video::COpenGLDriver*>(m_driver);
+
+    {
+        video::STextureSamplingParams params;
+        params.UseMipmaps = 0;
+        params.MaxFilter = params.MinFilter = video::ETFT_LINEAR_NO_MIP;
+        params.TextureWrapU = params.TextureWrapV = (_texWrapRepeat ? video::ETC_REPEAT : video::ETC_CLAMP_TO_EDGE);
+        //TODO do it without SAuxContext, and on different tex unit
```
Crisspl: to do it without SAuxContext, i'd have to create sampler object (in order not to modify texture's state) with pure GL calls, etc, etc.. I have to do it without SAuxContext because i have to bind it to tex unit untouchable by the engine (>=8) and SAuxContext doesnt allow it. @devshgraphicsprogramming  any ideas?

src/irr/video/CDerivativeMapCreator.cpp
Crisspl: why do you want `ceil` and not `floor`? + i've just taken the code from `COpenGL2DTexture`

src/irr/video/CDerivativeMapCreator.cpp
devshgraphicsprogramming: why do you have to bind it to a tex unit untouchable by the engine?

src/irr/video/CDerivativeMapCreator.cpp
devshgraphicsprogramming: ah yeah its 1+floor() so ok

PR #340, opened 2019-08-03
Mitsuba loader extension

Przemog1:


examples_tests/18.MitsubaLoader/main.cpp
```diff
@@ -0,0 +1,83 @@
...
+	
+	if (device == 0)
+		return 1; // could not create selected driver.
+
+	video::IVideoDriver* driver = device->getVideoDriver();
+	io::IFileSystem* fs = device->getFileSystem();
+	asset::IAssetManager& am = device->getAssetManager();
+
+	//temporary solution ofc
+	ext::MitsubaLoader::CMitsubaLoader* loader = new ext::MitsubaLoader::CMitsubaLoader(device);
```
devshgraphicsprogramming: register the loader with asset manager to comply with req. 0 from #340

examples_tests/18.MitsubaLoader/main.cpp
```diff
@@ -0,0 +1,83 @@
...
+
+#ifdef MITSUBA_LOADER_TESTS
+	std::string filePath = "C:\\IrrlichtBAW\\IrrlichtBAW\\ext\\MitsubaLoader\\testScene.xml";
+#else
+	pfd::message("Choose file to load", "Choose mitsuba XML file to load. \nIf you cancel or choosen file fails to load bathroom will be loaded.", pfd::choice::ok);
+	pfd::open_file file("Choose XML file", "", { "XML files (.xml)", "*.xml" });
+	std::string filePath = file.result().empty() ? "C:\\IrrlichtBAW\\/IrrlichtBAW\\examples_tests\\media\\mitsuba\\bathroom\\scene.xml" : file.result()[0];
+#endif
+
+	io::IReadFile* xmlFile = fs->createAndOpenFile(filePath.c_str());
```
devshgraphicsprogramming: if you register loader with asset manager, you can just load by path

ext/MitsubaLoader/CElementMatrix.h
```diff
@@ -0,0 +1,44 @@
...
+	virtual std::string getLogName() const override { return "matrix"; };
+
+	const core::matrix4SIMD getMatrix() const { return matrix; }
+
+private:
+	core::matrix4SIMD matrix;
+	const CElementMatrix::Type type;
+
+private:
+	std::pair<bool, core::matrix4SIMD> getMatrixFromString(std::string _str) const;
```
devshgraphicsprogramming: looks like a candidate for a static function ?

ext/MitsubaLoader/CElementMatrix.h
```diff
@@ -0,0 +1,44 @@
...
+	virtual bool processAttributes(const char** _args) override;
+	virtual bool onEndTag(asset::IAssetManager& _assetManager, IElement* _parent) override;
+	virtual IElement::Type getType() const override { return IElement::Type::MATRIX;  };
+	virtual std::string getLogName() const override { return "matrix"; };
+
+	const core::matrix4SIMD getMatrix() const { return matrix; }
+
+private:
+	core::matrix4SIMD matrix;
+	const CElementMatrix::Type type;
```
devshgraphicsprogramming: why do you need to track the type?

ext/MitsubaLoader/CElementMatrix.cpp
```diff
@@ -0,0 +1,83 @@
...
+		float f = std::numeric_limits<float>::quiet_NaN();
+		ss >> f;
+		std::cout << f << std::endl;
+
+		matrixData[i] = f;
+
+		if (isnan(f))
+		{
+			_IRR_DEBUG_BREAK_IF(true);
+			ParserLog::mitsubaLoaderError("Invalid matrix specified.");
```
devshgraphicsprogramming: `ParserLog::mitsubaLoaderError` what's this?

ext/MitsubaLoader/CElementMatrix.cpp
```diff
@@ -0,0 +1,83 @@
...
+				matrix = result.second;
+				return result.first;
+			}
+			else
+			{
+				ParserLog::wrongAttribute(_atts[i], getLogName());
+				false;
+			}
+		}
+	case CElementMatrix::Type::TRANSLATION:
```
devshgraphicsprogramming: explain?

ext/MitsubaLoader/CElementMatrix.h
```diff
@@ -0,0 +1,44 @@
...
+		ROTATION,
+		SCALE
+	};
+
+public:
+	CElementMatrix(CElementMatrix::Type _type)
+		:type(_type) {};
+
+
+	virtual bool processAttributes(const char** _args) override;
```
devshgraphicsprogramming: override + final

ext/MitsubaLoader/CElementMatrix.cpp
```diff
@@ -0,0 +1,83 @@
...
+namespace irr { namespace ext { namespace MitsubaLoader {
+
+
+bool CElementMatrix::processAttributes(const char** _atts)
+{
+
+	switch (type)
+	{
+	case CElementMatrix::Type::ARBITRARY:
+		for (int i = 0; _atts[i]; i += 2)
```
devshgraphicsprogramming: shouldn't it always be `_atts[2]==nullptr` and `_atts[0] && _atts[1]` ?

ext/MitsubaLoader/CElementFactory.cpp
```diff
@@ -0,0 +1,151 @@
...
+	ParserLog::mitsubaLoaderError("There is no type attribute for shape element. \n");
+	return nullptr;
+}
+
+IElement* CElementFactory::parseSimpleElement(const char* _el, const char** _atts, IElement::Type type)
+{
+	switch (type)
+	{
+	case IElement::Type::FLOAT:
+		return new CElementFloat();
```
devshgraphicsprogramming: ok if you're going to use `new` that heavily just spawn a context for the current loading session (like in BAW file loader) and give that context a linear allocator to use to do placement-new

ext/MitsubaLoader/CElementFactory.cpp
```diff
@@ -0,0 +1,151 @@
...
+				return nullptr;
+			}
+		}
+	}
+
+	ParserLog::mitsubaLoaderError("There is no type attribute for shape element. \n");
+	return nullptr;
+}
+
+IElement* CElementFactory::parseSimpleElement(const char* _el, const char** _atts, IElement::Type type)
```
devshgraphicsprogramming: or do you seriously need to implement the simple element stuff to dynamically create `IElement`s ?

ext/MitsubaLoader/CElementMatrix.h
Przemog1: I dont think I will need that anywhere else

devshgraphicsprogramming: if it does not need the internal state of an object (does not need `this`) if should always be a static.

Bonus point, you can have a private static.

ext/MitsubaLoader/CElementMatrix.h
Przemog1: to make downcasting easier, I do a lot of downcasting, especially in processChildData functions.

Just like in this example:
https://github.com/Przemog1/IrrlichtBAW/blob/mmitsuba_loader/ext/MitsubaLoader/CElementShapeOBJ.cpp#L59

devshgraphicsprogramming: but matrix is always 4x4 or 3x4 (with implicit 0,0,0,1 last row). This is matrix Type not element type!

ext/MitsubaLoader/CElementMatrix.cpp
Przemog1: it just prints "Mitsuba loader error:\nInvalid .xml file structure: '" followed by its only parameter `errorMessage`:

https://github.com/Przemog1/IrrlichtBAW/blob/mmitsuba_loader/ext/MitsubaLoader/ParserUtil.cpp#L29

I think it needs better name and also I will use there `Printer::log` instead of `Printer::print`

ext/MitsubaLoader/CElementShapeCube.h
```diff
@@ -0,0 +1,39 @@
...
+	CElementShapeCube();
+
+	virtual bool processAttributes(const char** _args) override;
+	virtual bool onEndTag(asset::IAssetManager& _assetManager, IElement* _parent) override;
+	virtual IElement::Type getType() const override { return IElement::Type::SHAPE_CUBE;  };
+	virtual std::string getLogName() const override { return "shape cube"; };
+	virtual bool processChildData(IElement* child) override;
+
+private:
+	core::matrix4SIMD transform;
```
devshgraphicsprogramming: so if you have `CElementMatrix` why do you keep a separate transform here?

ext/MitsubaLoader/CElementShapeCube.cpp
```diff
@@ -0,0 +1,55 @@
...
+}
+
+bool CElementShapeCube::onEndTag(asset::IAssetManager& _assetManager, IElement* _parent)
+{
+	asset::ICPUMesh* cubeMesh = _assetManager.getGeometryCreator()->createCubeMesh(core::vector3df(2.0f, 2.0f, 2.0f));
+
+	if (!cubeMesh)
+		return false;
+
+	//transform cubeMesh by this->transformMatrix
```
devshgraphicsprogramming: ok I see this as todo

ext/MitsubaLoader/CElementShapeOBJ.cpp
```diff
@@ -0,0 +1,79 @@
...
+
+#include "ParserUtil.h"
+#include "CSimpleElement.h"
+
+namespace irr { namespace ext { namespace MitsubaLoader {
+
+
+CElementShapeOBJ::~CElementShapeOBJ()
+{
+	if (mesh)
```
devshgraphicsprogramming: use smart pointer

ext/MitsubaLoader/CElementShapeOBJ.cpp
```diff
@@ -0,0 +1,79 @@
...
+bool CElementShapeOBJ::onEndTag(asset::IAssetManager& _assetManager, IElement* _parent)
+{
+	if (!fileName.size())
+	{
+		_IRR_DEBUG_BREAK_IF(true);
+		//print error (name of the .obj file is not set)
+		return false;
+	}
+
+	//getAssetInHierarchy?
```
devshgraphicsprogramming: You're gonna be wanting to be using a variant of this one
https://github.com/buildaworldnet/IrrlichtBAW/blob/2bf1c6a3e3518068c12433d560c23c2e9075b0c0/include/irr/asset/IAssetManager.h#L215

hierarchy is as follows:
- 0 for root level asset (an ICPUMesh for now, we have no scene object)
- 1 for whatever ICPUMesh contains, so meshbuffers

So the hierarchy would be 0u, 1u in the future if we make an ICPUScene object or something like that.

ext/MitsubaLoader/CElementShapeOBJ.h
```diff
@@ -0,0 +1,50 @@
+﻿#ifndef __C_ELEMENT_SHAPE_OBJ_H_INCLUDED__
+#define __C_ELEMENT_SHAPE_OBJ_H_INCLUDED__
+
+#include "IElement.h"
```
devshgraphicsprogramming: your include paths on extension headers are all wrong...

See RULE nr. 5 in the readme
https://github.com/buildaworldnet/IrrlichtBAW/tree/master/ext

ext/MitsubaLoader/CElementMatrix.cpp
Przemog1: if .xml file structure is correct, then _atts[0] is always "value" and _atts[1] is a string consisting of 16 floats

ext/MitsubaLoader/CMitsubaLoader.cpp
```diff
@@ -0,0 +1,107 @@
...
+ - handle 'version' attribute
+
+ - make sure default values for <float .. /> <boolean .. /> etc. are correct
+ - test atof function (CSimpleElement.h)
+ 
+ - how should loader treat upper/lower case letters in xml?
+
+ - clean up memory dynamically allocated by ParserData::scene
+
+ - resolve header include conflict (IElement.h needs to include ParserUtil.h (only for ParserLog), and ParserUtil.h needs to include IElement.h)
```
devshgraphicsprogramming: ParserUtil.cpp less inline functions

ext/MitsubaLoader/CElementMatrix.cpp
Przemog1: it is not implemented yet

Przemog1: sorry, I should leave some comment about that

ext/MitsubaLoader/CMitsubaScene.cpp
```diff
@@ -0,0 +1,83 @@
...
+bool CMitsubaScene::appendMesh(const asset::ICPUMesh* _mesh)
+{
+	if (!_mesh)
+	{
+		_IRR_DEBUG_BREAK_IF(true);
+		return false;
+	}
+
+
+	//is it possible that shapeMesh->getMeshBufferCount() > 1 ?
```
devshgraphicsprogramming: YES

ext/MitsubaLoader/CElementMatrix.cpp
devshgraphicsprogramming: but still, explain?

devshgraphicsprogramming:
Remember about this "Respect and use IAssetLoaderOverride during the loading process (v. important)"

I'm gonna be really anal about that.

ext/MitsubaLoader/CElementShapeOBJ.h
Crisspl: where this rule come from actually?

devshgraphicsprogramming: the justification is that the extension should not fuck-up your build system because you didn't add its directory to the search path.

Imagine that your project includes extension files, but extension files include their peers with relative paths, you're gonna have a fuck-up when two extensions have the same public include filenames, unless full paths relative to repo's `./include/` are specified

ext/MitsubaLoader/CElementMatrix.cpp
devshgraphicsprogramming: there's a thing called valiation, someone might be trying to feed the loader a bad .XML and the loader should not crash because of that

ext/MitsubaLoader/CElementShapeOBJ.cpp
```diff
@@ -38,7 +38,6 @@ bool CElementShapeOBJ::onEndTag(asset::IAssetManager& _assetManager, IElement* _
 		return false;
 	}
 
-	//getAssetInHierarchy?
```
devshgraphicsprogramming: what did we agree on in the end?

devshgraphicsprogramming: you should be using the overload!

ext/MitsubaLoader/CElementColor.cpp
```diff
@@ -0,0 +1,90 @@
...
+	return _parent->processChildData(this);
+}
+
+std::pair<bool, core::vector3df_SIMD> CElementColor::retriveColorFromValueAttribute(std::string value)
+{
+	std::replace(value.begin(), value.end(), ',', ' ');
+
+	float r = std::numeric_limits<float>::quiet_NaN();
+	float g = std::numeric_limits<float>::quiet_NaN();
+	float b = std::numeric_limits<float>::quiet_NaN();
```
devshgraphicsprogramming: what about alpha?

ext/MitsubaLoader/CElementMatrix.h
devshgraphicsprogramming: explain?

ext/MitsubaLoader/CMitsubaLoader.cpp
```diff
@@ -20,15 +20,36 @@ namespace irr { namespace ext { namespace MitsubaLoader {
...
 
  - clean up memory dynamically allocated by ParserData::scene
 
  - resolve header include conflict (IElement.h needs to include ParserUtil.h (only for ParserLog), and ParserUtil.h needs to include IElement.h)
 
+ - check, if transform name is indeed "toWorld"
+
+ - create an issue about isALoadableFileFormat in CBAWMeshFileLoader
+
+ - disable asset cacheing (use ECF_DUPLICATE_TOP_LEVEL)
```
devshgraphicsprogramming: why?

ext/MitsubaLoader/CMitsubaScene.cpp
```diff
@@ -72,9 +69,39 @@ bool CMitsubaScene::appendMesh(const asset::ICPUMesh* _mesh)
 	}
 
 
-	//is it possible that shapeMesh->getMeshBufferCount() > 1 ?
+	_mesh->grab();
+
 	for (int i = 0; i < _mesh->getMeshBufferCount(); i++)
-		mesh->addMeshBuffer(_mesh->getMeshBuffer(i));
+	{
+		//here vertex positions and normals are premultiplied with world transfrom matrix
```
devshgraphicsprogramming: no need if going to provide metadata

ext/MitsubaLoader/CMitsubaScene.cpp
```diff
@@ -72,9 +69,39 @@ bool CMitsubaScene::appendMesh(const asset::ICPUMesh* _mesh)
 	}
 
 
-	//is it possible that shapeMesh->getMeshBufferCount() > 1 ?
+	_mesh->grab();
```
devshgraphicsprogramming: use smart pointer?

devshgraphicsprogramming:
@Przemog1 You have an astronomical amount of duplicate code, use templates, CRTP or inhertance.

Przemog1:
>@Przemog1 You have an astronomical amount of duplicate code,

yeah, I had to redo a lot of things because of it

ext/MitsubaLoader/CMitsubaLoader.cpp
Przemog1: needed that for "baking" mesh transforms, but it is irrelevant now

include/irr/asset/IAsset.h
```diff
@@ -6,155 +6,166 @@
 #include "irr/core/Types.h"
 #include "irr/core/refctd_dynamic_array.h"
 
-namespace irr { namespace asset
```
devshgraphicsprogramming, line 7: what's happened here?

include/irr/asset/IAsset.h
```diff
@@ -6,155 +6,166 @@
...
+			virtual E_TYPE getAssetType() const = 0;
+			//! 
+			inline bool isADummyObjectForCache() const { return isDummyObjectForCacheAliasing; }
+		};
+
+		class SAssetBundle
+		{
+			using contents_container_t = core::refctd_dynamic_array<core::smart_refctd_ptr<IAsset>>;
+		public:
+			SAssetBundle(std::initializer_list<core::smart_refctd_ptr<IAsset>> _contents = {}) : m_contents(new contents_container_t(_contents), core::dont_grab)
```
devshgraphicsprogramming: @Crisspl and @Przemog1 you know that there's a `core::make_smart_refctd_ptr` so you don't need to use the nasty `dont_grab` constructor

include/irr/asset/IAsset.h
Przemog1: it is probably because I copied and pasted entire code from my "backup file"

ext/MitsubaLoader/CMitsubaScene.cpp
```diff
@@ -49,58 +48,17 @@ bool CMitsubaScene::processChildData(IElement* _child)
 			return false;
 		}
 
-		const asset::ICPUMesh* shapeMesh = shape->getMesh();
-		return appendMesh(shapeMesh, shape->getTransformMatrix());
+		const core::smart_refctd_ptr<asset::ICPUMesh> shapeMesh = shape->getMesh();
```
devshgraphicsprogramming, line 51: what are you doing here?

ext/MitsubaLoader/Shape.h
```diff
@@ -12,7 +12,7 @@ class CShape : public IElement
 	CShape()
 		: mesh(nullptr) {};
 
-	inline asset::ICPUMesh* getMesh() { return mesh; }
+	inline core::smart_refctd_ptr<asset::ICPUMesh> getMesh() { return mesh; }
```
devshgraphicsprogramming, line 15: the convention here is to return the raw pointer, we only want to return a `smart_refctd_ptr` if we are sure that the callee wants to hold onto the reference for a longer time.

ext/MitsubaLoader/CMitsubaScene.cpp
Przemog1: here i retrive mesh from shape element

Przemog1: then this retrived mesh is added to the array which contain all loaded meshes

devshgraphicsprogramming: I meant ref-counting-wise?

PR #341, opened 2019-08-03
TSSAA

devshgraphicsprogramming:
need to sort out the branches before I review


examples_tests/35.TAA/main.cpp
```diff
@@ -311,6 +311,8 @@ int main()
         else
         {
             prevVP = currVP;
+            // TODO how to force driver to update view and projection matrices without smgr->drawAll()?
+            // without this fix, here i'm getting previous frame's VP to currVP (and so prevVP is 2 frames in the past)
```
devshgraphicsprogramming: get the transforms from the camera, and not the driver... also refactor into an extension like I told you on discord, then there is absolutely no problem with history tracking

examples_tests/35.TAA/main.cpp
```diff
@@ -294,17 +289,40 @@ int main()
...
+        }
+        else
+        {
+            prevVP = currVP;
+            currVP = driver->getTransform(video::EPTS_PROJ_VIEW);
+#if USE_JITTER
+            core::matrix4SIMD jitter;
+            float jitterOffset[3]{ 0.f,0.f,0.f };
+            memcpy(jitterOffset, &JITTER_OFFSETS[g_FrameNum%JITTER_OFFSET_CNT].X, 8);
+            jitter.setTranslation(jitterOffset);
```
devshgraphicsprogramming, line 297: apply Jitter onto the camera proj matrix before it `render`s or `drawAll` is called, that will make all matrices consistent (all combos of proj matrix) also right now you're not rendering with the jitter!

Secondly you've hardcoded the `JITTER_OFFSETS` to have the window resolution inside, it would be far better to have them [0,1)^2 pixel footprint relative sobol or halton samples (or standard Vulkan MSAA pattern sequence) that are later expanded to viewport resolution (not render target resolution, but active viewport resolution).

Furthermore your offsets are miscalculated because NDC is in [-1,1]^2 so your jitters need to be in range [-1/RES_X,1/RES_X]*[-1/RES_Y,1/RES_Y], so approximately 2x too small and biased to one side

Lastly, the matrix concatenation is wrong, there will be a W divide at the end, and that will make the vanishing point at (0,0) NDC, when in-fact for a jittered view, the vanishing point should be at `jitter_offset`.
https://computergraphics.stackexchange.com/questions/6021/changing-the-projection-reference-point-in-open-gl
Which means that you also need to modify column 3 to add a `jitter` proportional factor for the viewspace Z component, so the vanishing line/point stays the same in world coordinates.

examples_tests/35.TAA/main.cpp
```diff
@@ -294,17 +289,40 @@ int main()
...
+            currVP = core::matrix4SIMD::concatenateBFollowedByA(jitter, currVP);
+#endif//USE_JITTER
+        }
+        velBufCb->setPrevVP(prevVP);
+        velBufCb->setCurrentVP(currVP);
+
         // VELOCITY BUF PASS
```
devshgraphicsprogramming: for the love of god precompute your FBO attachment combos + use Multiple Render Targets to output velocity + color in one pass!

examples_tests/35.TAA/main.cpp
```diff
@@ -333,6 +351,7 @@ int main()
         // AA PASS
         if (g_FrameNum > 0u) //first frame is without AA
         {
+            TAAcb->setJitterOffset(JITTER_OFFSETS[g_FrameNum%JITTER_OFFSET_CNT]);
```
devshgraphicsprogramming, line 314: does your TAA shader also know the previous frame's jitter offset so it can properly look things up in history buffer?

examples_tests/35.TAA/quad.vert
```diff
@@ -0,0 +1,12 @@
+#version 430 core
```
devshgraphicsprogramming, line 1: A. We have a full screen triangle extension to use
B. We shouldn't be even using that anymore because it should be done in compute shaders

examples_tests/35.TAA/taa.frag
```diff
@@ -0,0 +1,61 @@
...
+layout (binding = 3) uniform sampler2D uVelocityBuf;
+
+in vec2 TexCoords;
+
+layout (location = 0) out vec4 outScreen;  //what will be visible on screen
+layout (location = 1) out vec4 outHistory; //history for next frame
+
+float luminanceFromRGB(in vec3 color)
+{
+	return dot(color, vec3(0.299, 0.587, 0.114));
```
devshgraphicsprogramming: a dot is just a 1x3 and 3x1 matrix transform, if first matrix is 3x3 then you can have RGB to YUV

Just remember to convert Tonemapped sRGB -> linear RGB -> linear YUV then do the clipping/clamping in linear YUV

examples_tests/35.TAA/taa.frag
```diff
@@ -0,0 +1,61 @@
...
+	vec4 clip = vec4(2.0*vec3(TexCoords, depth) - 1.0, 1.0);
+	vec4 world = invVP*clip;
+	world = world/world.w;
+	
+	vec4 q_cs = uPrevVP*world;
+	vec2 q_uv = 0.5*(q_cs.xy/q_cs.w) + 0.5;
+	vec2 vel = TexCoords - q_uv;
+*/
+	// it should sample nearest (depth) in 3x3 neighbourhood
+	vec2 vel = textureLod(uVelocityBuf, TexCoords, 0.0).xy;
```
devshgraphicsprogramming: when you use compute, change that to texelFetch

examples_tests/35.TAA/taa.frag
```diff
@@ -0,0 +1,61 @@
...
+	
+	vec2 q_uv = TexCoords - vel;
+	
+	vec3 current = textureLod(uCurrentBuf, TexCoords, 0.0).rgb;
+	vec3 history = textureLod(uHistoryBuf, q_uv, 0.0).rgb;
+	// AABB clamping, better try (variance?) clipping later (TODO)
```
devshgraphicsprogramming: ok but compute the AABB/Variance Ellipsoid in YUV

Btw, you can compute a 3x3 covariance matrix, that basically gives an ellipsoid that you can scale up and down and clip against

examples_tests/35.TAA/taa.frag
```diff
@@ -0,0 +1,61 @@
...
+	vec3 neigh0 = textureLodOffset(uCurrentBuf, TexCoords, 0.0, ivec2(1, 0)).rgb;
```
devshgraphicsprogramming: use `textureGather()` you'll get all 2x2 neighbourhood at once (one channel at a time)

examples_tests/35.TAA/taa.frag
```diff
@@ -0,0 +1,61 @@
...
+	vec3 neigh1 = textureLodOffset(uCurrentBuf, TexCoords, 0.0, ivec2(0, 1)).rgb;
+	vec3 neigh2 = textureLodOffset(uCurrentBuf, TexCoords, 0.0, ivec2(-1, 0)).rgb;
+	vec3 neigh3 = textureLodOffset(uCurrentBuf, TexCoords, 0.0, ivec2(0, -1)).rgb;
+
+	vec3 aabb_min = min(current, min(neigh0, min(neigh1, min(neigh2, neigh3))));
+	vec3 aabb_max = max(current, max(neigh0, max(neigh1, max(neigh2, neigh3))));
+
+	history = clamp(history, aabb_min, aabb_max);
```
devshgraphicsprogramming: clamp will shift hue, clip instead.... clip against elipsoid to be even nicer

devshgraphicsprogramming: You can clip a line against an ellipsoid quite easily in the special case of a line originating at 0 and an ellipsoid centered at 0 as well. When doing ellipsoid fitting with covariance matrices, you'll have to center the ellipsoid on the mean/average value of your sample set.

You just need to find the 3x3 matrix that transforms a unit sphere into your ellipsoid, then find the inverse of that matrix, you transform the direction vector of the line (doesn't matter if normalized) by the inverse matrix, that moves it into the unit sphere space, then if you normalize the transformed vector you clip it!

Now transform the normalized vector by the matrix that scales a unit sphere into your ellipsoid, and you will get the clipped vector.

Actually the matrix that makes your unit sphere the elipsoid is the covariance matrix
https://towardsdatascience.com/interesting-properties-and-use-cases-of-the-covariance-matrix-80d72939c7ab?gi=1d570671e2a6
That's actually how we visuallize the contours of a mixture of gaussians ;)

You can look at how the Ellipsoid Collider in our simple raycast collision code does it. But anyway this is what happens
```cpp
 // deviationScale is how many variances away do you want the ellipsoid, same as for gaussian, 1 <= 68%, 2 <= 95%, 3 <= 97.5%
mat3 scaledCovariance = deviationScale*covariance;

vec3 lineDir = history-averageColorOfNeighbourhood;
vec3 unitClipped = inverse(scaledCovariance)*lineDir;
float unitClippedLen = length(unitClipped);
if (unitClippedLen>1.0)
   unitClipped /= unitClippedLen;
vec3 lineDirClipped = scaledCovariance*unitClipped;
history = lineDirClipped+averageColorOfNeighbourhood;
```

Crisspl: don't review outdated sources, it makes no sense

devshgraphicsprogramming: If you factor it all out then really what is happening is
```c++
 // deviationScale is how many variances away do you want the ellipsoid, same as for gaussian, 1 <= 68%, 2 <= 95%, 3 <= 97.5%
vec3 lineDir = history-averageColorOfNeighbourhood;
float normFactor = deviationScale/length(inverse(covariance)*lineDir); // maybe there's a way to optimize the inverse and length
if (normFactor<1.0)
   history = lineDir*normFactor+averageColorOfNeighbourhood;
```

examples_tests/35.TAA/taa.frag
```diff
@@ -0,0 +1,61 @@
...
+	
+	vec3 current_unjittered = textureLod(uCurrentBuf, TexCoords-uJitter, 0.0).rgb;
+	
+	vec3 color = mix(current_unjittered, history, 0.7);
```
devshgraphicsprogramming: I think INSIDE used a factor of 0.9 or something

examples_tests/35.TAA/taa.frag
```diff
@@ -0,0 +1,61 @@
...
+	
+	outHistory = vec4(color, 1.0);
+	
+	// TODO fallback to motion blur (for fast moving fragments) and this will actually go to screen
```
devshgraphicsprogramming: can you do both TSSAA and Motion Blur combined at the same time?

examples_tests/35.TAA/velbuf.vert
```diff
@@ -0,0 +1,22 @@
...
+layout (location = 0) in vec3 vPos;
+
+layout (location = 0) uniform mat4 MVP;
+layout (location = 1) uniform mat4 prevMVP;
+
+out vec2 vVelocity;
+
+void main()
+{
+	vec4 p_cs = MVP*vec4(vPos, 1.0);
```
devshgraphicsprogramming: define a FAST_TRANSFORM macro for future use `Matrix[0]*vec.x+Matrix[1]*vec.x+Matrix[2]*vec.x+Matrix[3]`

examples_tests/35.TAA/velbuf.vert
```diff
@@ -0,0 +1,22 @@
...
+	gl_Position = p_cs;
+	
+	// convert to ndc
+	p_cs /= p_cs.w;
```
devshgraphicsprogramming: nope, Z-divide must happen per pixel (move it to pixel shader and pass `xyw` along).

examples_tests/35.TAA/velbuf.vert
```diff
@@ -0,0 +1,22 @@
...
+	vec4 q_cs = prevMVP*vec4(vPos, 1.0);
+	q_cs /= q_cs.w;
+	
+	// convert to screen space and calculate velocity
+	vVelocity = 0.5*(p_cs.xy - q_cs.xy); // 0.5*(p-q) is same as (0.5*p + 0.5) - (0.5*q + 0.5)
```
devshgraphicsprogramming: just beware of NDC Y direction and Texture Coordinate Y direction (we use Vulkan convention)

examples_tests/35.TAA/main.cpp
```diff
@@ -281,9 +281,16 @@ int main()
...
 
     video::SGPUMaterial TAAMaterial;
+    video::STextureSamplingParams sparams;
+    sparams.UseMipmaps = 0u;
+    sparams.MinFilter = sparams.MaxFilter = video::ETFT_NEAREST_NO_MIP;
+    sparams.TextureWrapU = sparams.TextureWrapV = video::ETC_MIRROR_CLAMP_TO_EDGE;
     TAAMaterial.setTexture(1u, colorBuf);
+    TAAMaterial.TextureLayer[1].SamplingParams = sparams;
     TAAMaterial.setTexture(2u, historyBuffers[1]);
+    TAAMaterial.TextureLayer[2].SamplingParams = sparams;
```
devshgraphicsprogramming, line 263: waaait, why no filtering when sampling history buffer?

examples_tests/35.TAA/taa.frag
```diff
@@ -19,6 +19,36 @@ float luminanceFromRGB(in vec3 color)
...
+vec3 constrainHistSample(in vec3 hist, in vec3 aabb_min, in vec3 aabb_max)
+{
+	if (aabb_min==aabb_max)
+		return aabb_min; //return immediately if all neighbourhood samples were same color
+	// scale needed to transform unit sphere into ellipsoid insribed into the aabb
+	vec3 scale = 0.5*(aabb_max-aabb_min);
+	vec3 inv_scale = vec3(1.0)/scale;
+	vec3 center = 0.5*(aabb_min+aabb_max);
+	
+	// translate into ellipsoid's space and scale hist sample by inverse scale
```
devshgraphicsprogramming, line 54: for proper ellipsoid calculate 3x3 covariance matrix (use YUV as XYZ for usual Machine Learning/Statistics algorithm)

you can probably do some funky simple unrolled algo since you'll only have 4 samples.

More here:
https://github.com/buildaworldnet/IrrlichtBAW/pull/341/commits/d71102e72144c3d383eb5cc9aa0087d313dca22c#r310388184

examples_tests/35.TAA/taa.frag
```diff
@@ -48,10 +77,11 @@ void main()
 	vec3 aabb_min = min(current, min(neigh0, min(neigh1, min(neigh2, neigh3))));
 	vec3 aabb_max = max(current, max(neigh0, max(neigh1, max(neigh2, neigh3))));
 
-	history = clamp(history, aabb_min, aabb_max);
+	history = constrainHistSample(history, aabb_min, aabb_max);
 	
 	vec3 current_unjittered = textureLod(uCurrentBuf, TexCoords-uJitter, 0.0).rgb;
 	
+	// TODO use luma to vary interpolation factor as playdeadgames does
```
devshgraphicsprogramming: nice idea.

examples_tests/35.TAA/velbuf.vert
devshgraphicsprogramming: You will have epic arithmetic, divide by 0 and other errors on interpolation for triangles close to the near plane... this is really needed per pixel

examples_tests/35.TAA/taa.frag
```diff
@@ -14,6 +14,8 @@ in vec2 TexCoords;
 layout (location = 0) out vec4 outScreen;  //what will be visible on screen
 layout (location = 1) out vec4 outHistory; //history for next frame
 
+#define SCREEN_SIZE vec2(1280.0, 720.0)
```
devshgraphicsprogramming, line 17: plz plz plz don't hardcode

examples_tests/35.TAA/taa.frag
```diff
@@ -109,11 +114,41 @@ void main()
...
+	weight = 1.0 - weight;
+	weight = weight*weight;
+	
+	float lerpFactor = mix(0.5, 0.9, weight);
+	
+	vec3 color = mix(current_unjittered, history, lerpFactor);
 	
 	outHistory = vec4(color, 1.0);
 	
-	// TODO fallback to motion blur (for fast moving fragments) and this will actually go to screen
```
devshgraphicsprogramming: can you do motion blur almost always?

devshgraphicsprogramming:
@Crisspl you do realize that the velocity buffer is practically almost not needed at all?

Unless ofc you plan to blend velocities for some weird obscure reason (particles).

Since we can very easily work out velocity
```c++
uniform mat4 matrixDiff = oldModelViewProj*inverseCurrentModelViewProj;
// Since (p.v.m) . (P.V.M)^-1 = p.v.m M^-1.V^-1.P^-1 , i.f.f. object does not move, the old model matrix cancels out with the new model matrix, so can use below for prototype
//uniform mat4 matrixDiff = oldViewProj*inverseCurrentViewProj;

vec4 currentProjPos = NDC.xyz*NDC.w; // get back gl_Position, same way that you do for deferred renderers.
vec4 oldProjPos = matrixDiff*currentProjPos;
vec2 velocityNDC = NDC.xy-oldProjPos.xy/oldProjPos.w;
```

devshgraphicsprogramming:
For getting NDC position or View Space Position from non-linear ZBuffer plus texcoords you can check out my previous work here
https://github.com/ConfettiFX/The-Forge/blob/master/Examples_3/Unit_Tests/src/09_LightShadowPlayground/Shaders/Vulkan/sdfSphere.frag#L106

examples_tests/35.TAA/main.cpp
devshgraphicsprogramming: @Crisspl PAge 11 http://twvideo01.ubm-us.net/o1/vault/gdc2016/Presentations/Pedersen_LasseJonFuglsang_TemporalReprojectionAntiAliasing.pdf

> use projection offset to shear frustum

You don't have the shear that maintains the same vanishing point

examples_tests/35.TAA/taa.frag
devshgraphicsprogramming: Also page 17 from Tiago Sousa had a funky idea, but that requires velocity info for both current and previous frame (so current matrix, 1 frame behind, and 2 frames behind).
https://www.slideshare.net/TiagoAlexSousa/graphics-gems-from-cryengine-3-siggraph-2013

And obviously to get v_p, you'd need to reproject pixel to previous frame, and then reproject again, then the diff between the reprojections would be v_p

devshgraphicsprogramming:
When we encounter skinned meshes or moving meshes, we could implement what is already required in Deferred+, that is give every object and triangle a Unique ID.

Then given the world space (or view space) 3D fragment postion, work out the triangle barycentric coordinates, then use data about the previous frame (old transformed skinning vertices by compute shader, old MVP matrices, etc.) to compute the 3 triangle vertex `gl_Position`s and reproject like that ;)

examples_tests/35.TAA/main.cpp
```diff
@@ -277,14 +277,16 @@ int main()
     auto tmpColorBuf = driver->createGPUTexture(video::ITexture::ETT_2D, WINDOWS_SIZE_array, 1u, asset::EF_R8G8B8_UNORM);
     auto depthBuf = driver->createGPUTexture(video::ITexture::ETT_2D, WINDOWS_SIZE_array, 1u, asset::EF_D16_UNORM);
 
+    //TODO maybe better to use separate fbo for each pass? (wouldn't need to reattach every pass every frame)
```
devshgraphicsprogramming: YES

examples_tests/35.TAA/taa.frag
devshgraphicsprogramming: Actually luma weight is used by UE4 + INSIDE did their research and Sousa velocity weight didn't work for them that well... but you can try it once you have the basic TAA working to see whats best.

examples_tests/35.TAA/taa.frag
```diff
@@ -61,8 +90,8 @@ void main()
 	vec2 q_uv = 0.5*(q_cs.xy/q_cs.w) + 0.5;
 	vec2 vel = TexCoords - q_uv;
 */
-	// it should sample nearest (depth) in 3x3 neighbourhood
-	vec2 vel = textureLod(uVelocityBuf, TexCoords, 0.0).xy;
+	// TODO test later wether closestFragmentOffsetIn3x3() is worth it
+	vec2 vel = textureLodOffset(uVelocityBuf, TexCoords, 0.0, closestFragmentOffsetIn3x3(TexCoords)).xy;
```
devshgraphicsprogramming: use `texelFetch`,  offset parameter in `*Offset` in GLSL has to always be a compile-time constant.

examples_tests/35.TAA/taa.frag
devshgraphicsprogramming: refactored to make sure clipping does not project (make vector longer)

examples_tests/35.TAA/taa.frag
devshgraphicsprogramming: also these fetches don't make sense, you're getting a cross around the jittered sample right now
```
0 x 0
x x x
0 x 0
```

do 3 `textureGather`s around the unjittered `TexCoords-uJitter` position, then what you're doing will at least have some meaning/semantic (the gathered samples will be from within output pixel's footprint).

An even higher level of advancement would be to use `texelFetch` and Compute Shader 17x17 shared memory with 16x16 workgroups, then you could get 2x2 or even 4x4 (but then 19x19 smem) neighbourhoods for the covariance effortlessly.

examples_tests/35.TAA/taa.frag
```diff
@@ -0,0 +1,154 @@
...
+	vec3 neigh1 = textureLodOffset(uCurrentBuf, TexCoords, 0.0, ivec2(0, 1)).rgb;
+	vec3 neigh2 = textureLodOffset(uCurrentBuf, TexCoords, 0.0, ivec2(-1, 0)).rgb;
+	vec3 neigh3 = textureLodOffset(uCurrentBuf, TexCoords, 0.0, ivec2(0, -1)).rgb;
+
+	vec3 aabb_min = min(current, min(neigh0, min(neigh1, min(neigh2, neigh3))));
+	vec3 aabb_max = max(current, max(neigh0, max(neigh1, max(neigh2, neigh3))));
+
+	history = constrainHistSample(history, aabb_min, aabb_max);
+	
+	vec3 current_unjittered = textureLod(uCurrentBuf, TexCoords-uJitter, 0.0).rgb;
```
devshgraphicsprogramming: see Brian Karis' 2013 TemporalAA presentation where he mentions Burely07 that Box filter does not behave well with moving samples, so for this tap (and this tap only) a pseudo-small-kernel-gaussian 2 wide pixel support reconstruction filter (save bandwidth with shared memory or approximating gaussian with 4=2x2 `textureLodOffset`).

examples_tests/35.TAA/taa.frag
```diff
@@ -0,0 +1,154 @@
...
+	
+	float lum = luminanceFromRGB(current_unjittered);
```
devshgraphicsprogramming: Luminance is just Y in YUV rep, but you need to find out if they want HDR pre-tonemap, post-tonemap linear or gamma luminance in this weighing scheme.

examples_tests/35.TAA/taa.frag
```diff
@@ -0,0 +1,154 @@
...
+	world = world/world.w;
+	
+	vec4 q_cs = uPrevVP*world;
+	vec2 q_uv = 0.5*(q_cs.xy/q_cs.w) + 0.5;
+	vec2 vel = TexCoords - q_uv;
+*/
+	// TODO test later whether closestFragmentOffsetIn3x3() is worth it
+	vec2 vel = textureLodOffset(uVelocityBuf, TexCoords, 0.0, closestFragmentOffsetIn3x3(TexCoords)).xy;
+	
+	vec2 q_uv = TexCoords - vel;
```
devshgraphicsprogramming, line 133: aaand what do you do if `q_uv` is outside the screen?

examples_tests/35.TAA/taa.frag
```diff
@@ -0,0 +1,154 @@
...
+	
+	vec3 current = textureLod(uCurrentBuf, TexCoords, 0.0).rgb;
+	vec3 history = textureLod(uHistoryBuf, q_uv, 0.0).rgb;
```
devshgraphicsprogramming: See https://github.com/buildaworldnet/IrrlichtBAW/pull/341/files#r310394150

examples_tests/35.TAA/taa.frag
```diff
@@ -0,0 +1,154 @@
...
+	vec4 q_cs = uPrevVP*world;
+	vec2 q_uv = 0.5*(q_cs.xy/q_cs.w) + 0.5;
+	vec2 vel = TexCoords - q_uv;
+*/
+	// TODO test later whether closestFragmentOffsetIn3x3() is worth it
+	vec2 vel = textureLodOffset(uVelocityBuf, TexCoords, 0.0, closestFragmentOffsetIn3x3(TexCoords)).xy;
+	
+	vec2 q_uv = TexCoords - vel;
+	
+	vec3 current = textureLod(uCurrentBuf, TexCoords, 0.0).rgb;
```
devshgraphicsprogramming: if you "fix" your gather for AABB/Covariance samples, then you don't need this tap

examples_tests/35.TAA/taa.frag
```diff
@@ -61,8 +90,8 @@ void main()
 	vec2 q_uv = 0.5*(q_cs.xy/q_cs.w) + 0.5;
 	vec2 vel = TexCoords - q_uv;
 */
-	// it should sample nearest (depth) in 3x3 neighbourhood
-	vec2 vel = textureLod(uVelocityBuf, TexCoords, 0.0).xy;
+	// TODO test later wether closestFragmentOffsetIn3x3() is worth it
```
devshgraphicsprogramming: TOTALLY WORTH IT - see Karis TemporalAA pptx slide 27

However you can still dilate velocity if you have no velocity texture (which is what we should be striving towards).

Just when you run your 16x16 WG Compute Shader, compute Velocity for 18x18 tile (1 pixel border) in shared memory, then dilate (run your filter that's Z dependant) over the innermost 16x16 samples. Basically you'll be making a Tiled Renderer in software XD.

examples_tests/35.TAA/taa.frag
devshgraphicsprogramming: @Crisspl corrected my code listings

examples_tests/35.TAA/velbuf.vert
devshgraphicsprogramming: @Crisspl if you just nuked the Velocity Buffer and went with Deferred+ Ready idea from here
https://github.com/buildaworldnet/IrrlichtBAW/pull/341#issuecomment-518009418

then all of the above would go away and be a non-issue

examples_tests/35.TAA/main.cpp
devshgraphicsprogramming: Look here at `a`,`b`,`c`
https://github.com/playdeadgames/temporal/blob/master/Assets/Scripts/Extensions.cs#L74

Its a shear

examples_tests/35.TAA/taa.frag
devshgraphicsprogramming: I mean like have this extension be a combined TSSAA + Motion Blur

examples_tests/35.TAA/taa.frag
devshgraphicsprogramming: Ok Uncharted and Lumberyard car game
https://aws.amazon.com/blogs/gametech/anti-ghosting-with-temporal-anti-aliasing/

Used stencil buffer to augment luma weight and color clipping (which should be done in YUV or YCoCg), basically if stencil fails then no history sample to be taken (weight = 0.0).

something similar could be done not with the stencil buffer, but with ObjectID buffer, in the far future.

examples_tests/35.TAA/taa.frag
devshgraphicsprogramming: Filmic Temporal Supersample AA presentation mentions using a Bicubic fast appriximation (catmul-rom 5 tap bilinear thing)

examples_tests/35.TAA/taa.frag
devshgraphicsprogramming: Filmic Temporal SuperSample AA mentions depth-rejection of certain samples from the neighbourhood for a tighter bound and less ghosting (so you'd be using 1-5 samples, instead of 5 all the time for your old method, and 1-16 in my new method).

examples_tests/35.TAA/taa.frag
Crisspl: clamp to edge

examples_tests/35.TAA/main.cpp
Crisspl: now its `ETFT_LINEAR_NO_MIP` (thats why you shouldnt review outdated code)

examples_tests/35.TAA/main.cpp
devshgraphicsprogramming: Also lots of impl details here
http://advances.realtimerendering.com/s2016/s16_Ke.pptx

examples_tests/35.TAA/taa.frag
devshgraphicsprogramming: I think correct solution would be to never take a history sample into account because its simply not valid

examples_tests/35.TAA/quad.vert
Crisspl: A. it uses full screen triangle
B. yea, i'll rewrite it as CS when it goes as irrbaw extension

examples_tests/35.TAA/taa.frag
Crisspl: hm why?

devshgraphicsprogramming:
@Crisspl last thing, try and work out the average distance between your jitter samples (obvsiouly repeat them a in a 3x3 grid) to find out the Mip-Map bias to set on all textures (its an OpenGL setting, you can extend `STextureLayer`)

I'd expect this, where N is numer of jitter sample positions

N=1 -> avg_distance = 1.0 pixel
N=2 -> avg_distance = 0.707 pixel
etc.

Your MipMap bias needs to be `log2(avg_distance)`

This is because now textures will get smoother and blurrier, so TSSAA can provide better quality AA in terms of texture filtering.

examples_tests/35.TAA/main.cpp
```diff
@@ -0,0 +1,398 @@
...
+
+private:
+};
+
+// GLOBAL CONSTANTS
+static const dimension2d<uint32_t> WINDOW_SIZE(1280, 720);
+static const uint32_t WINDOWS_SIZE_array[]{ 1280, 720 };
+static const core::vector2df WINDOW_SIZE_vec2df(1280.f, 720.f);
+constexpr uint32_t JITTER_OFFSET_CNT = 8u;
+static const core::vector2df JITTER_OFFSETS[JITTER_OFFSET_CNT]{
```
devshgraphicsprogramming, line 45: ok even according to INSIDE and Ke XU
http://advances.realtimerendering.com/s2016/s16_Ke.pptx

That sequence needs to be Halton or Sobol, probably Halton best.

examples_tests/35.TAA/taa.frag
Crisspl: >(the gathered samples will be from within output pixel's footprint)

i dont understand. What is output pixel's footprint?

devshgraphicsprogramming: Output pixel `x` and its center + its footprint corners `f`.
```
f000f
00000
00x00
00000
f000f
```
Jittered pixel center `j` (where sampling, rasterizing and shading occurs)
```
f000f
0000j
00x00
00000
f000f
```
If you sample the jittered texture at `j-Jitter` (which is `j-(j-x)==x`) then you're in the 2x2 patch between jittered samples, `textureGather` will get you 4 samples from jittered texture around the unjittered position.

examples_tests/35.TAA/taa.frag
devshgraphicsprogramming: Compute has better performance+scheduling than fullscreen tri, the only time you want to use FS quad is when you want to leverage fast culls like stencil for some weird operation.

Not to mention that with Vulkan if you use Compute you can do Async Compute.

Finally texelFetch because compute gives you integer grid coordinates, so there's no need for that FP32 inaccuracy?

examples_tests/35.TAA/taa.frag
devshgraphicsprogramming: Also for reference
https://youtu.be/WzpLWzGvFK4?t=600

examples_tests/35.TAA/taa.frag
devshgraphicsprogramming: Also if you're going to use depth-rejection or object ID to tighten the neighbourhood bounds, then there needs to be a parameter to turn that off (for scenes with translucent media)

examples_tests/35.TAA/taa.frag
devshgraphicsprogramming: Page 26 of
https://www.dropbox.com/sh/dmye840y307lbpx/AAAQpC0MxMbuOsjm6XmTPgFJa?preview=msalvi_temporal_supersampling.pptx

Shows variance clipping, which is a more rudimentary approach.

If you have full covariance matrix (9 floats) instead of variance (2x3 floats) then you can clip against oriented ellipsoid instead of Axis Aligned (unoriented) bounding box.

Now you have a chance for a little bit of original research and contribution to science ;)

examples_tests/35.TAA/taa.frag
devshgraphicsprogramming: For transparents, page 46 https://www.dropbox.com/sh/dmye840y307lbpx/AAAQpC0MxMbuOsjm6XmTPgFJa?preview=msalvi_temporal_supersampling.pptx

suggests that we pick the motion vector that is essentially "the most visible" in the neighbourhood (which I guess wouldn't work as well anymore if all objects are equally opaque, but one could make a heuristic so the motion vector of most-contributing blend is picked for motion blur, and then those vectors are dilated with the whole min-depth filter).

devshgraphicsprogramming: look at pages 50-53 where the Variance Clipping (with 7x7 gather, not 3x3) makes Stochastic Transparency work.

Crisspl:
some screenshots
no camera movement:

taa_d 2019-08-05 18-42-51-49
slight camera movements:
taa_d 2019-08-05 18-43-33-70
fast camera movements:
taa_d 2019-08-05 18-44-06-54

as bmp (i had to convert to jpg for github to allow uploading it): https://www.dropbox.com/sh/e7rejuykgmevy5e/AAAKpTWYovRlpKcjOFtAcdj4a?dl=0

devshgraphicsprogramming:
@Crisspl Oh I read in Ke XU's presentation that the history buffer needs to be FP16, which makes sense because usually we have banding and loose a lot of info to the extent of needing dithering (I actually inject noise before writing out to 8bit textures/framebuffer in Build A World). So basically output post-gamma non-linear (lin2srgb) colors to a linear half-float history buffer.

Also since you will be using the compute shader, you can finally use RGB9E5 shared exponent non-renderable format for the history buffer (make an r32uint image view, use as read+write storage image and pack/unpack manually) which saves us 2x space!

To reduce banding and allow HDR displays (ITU 2020) your input could be FP16, R11G11B10F or RGB9E5, even though its post-resolve/post-gamma. If you get SRGB8 texture as input remember not to sample it as SRGB8 (it will linearize, you don't want to work in linear space), create a RGB(A)8_UNORM view for it.

Finally you could add dithering (nice noise) on the non-history output, after the motion blur if any.

In the future we will roll Tonemapping -> Color Correction -> OETF -> TSSAA & Motion Blur & Output Dithering into one shader

devshgraphicsprogramming:
Those screenshots don't look like AA, there's some blur but the straight edges of the box against the background don't look like 8x TSSAA

Crisspl:
what actually means "8x" in "8x TSSAA"?

devshgraphicsprogramming:
8 jitter positions.

examples_tests/35.TAA/main.cpp
```diff
@@ -300,7 +275,7 @@ int main()
 		driver->beginScene(false, false, video::SColor(255,255,255,255) );
 
         //force view and projection matrices update
-        camera->render();
+        camera->render();//this sets view matrix to all nans in 1st frame...
```
devshgraphicsprogramming, line 278: you also need `OnAnimate()` that actually calculates the matrices from mouse and keyboard

`render()` only sets them in the driver

examples_tests/35.TAA/mesh.frag
```diff
@@ -1,12 +1,18 @@
-#version 330 core
+#version 430 core
 
 in vec4 Color; //per vertex output color, will be interpolated across the triangle
 in vec3 Normal;
 in vec3 lightDir;
+in vec3 currPos_cs;
```
devshgraphicsprogramming, line 6: this one you dont need, you can calc `currPos_ndc` from `gl_FragCoord`

devshgraphicsprogramming: but still, best not to output a velocity buffer at all ;)

examples_tests/35.TAA/taa.frag
```diff
@@ -16,9 +16,21 @@ layout (location = 1) out vec4 outHistory; //history for next frame
 
 #define SCREEN_SIZE vec2(1280.0, 720.0)
 
-float luminanceFromRGB(in vec3 color)
+vec3 RGB_YCoCg(vec3 c)
```
devshgraphicsprogramming, line 19: I would have preferred YUV

devshgraphicsprogramming:
we'll redo it at some point

PR #372, opened 2019-09-20
New Mitsuba Loader

devshgraphicsprogramming:


devshgraphicsprogramming:
Need to support names on Elements that are not tied to elements (there's both an `id` and `name` attribute).

PR #373, opened 2019-09-22
Document important headers, first approach

devshgraphicsprogramming:
I have to wait for criss, without him I won't be able
to document some functionality of those files due to
the lack of knowledge about what's going on there

### DIFF
--- a/include/irr/asset/IAsset.h
+++ b/include/irr/asset/IAsset.h
@@ -11,13 +11,20 @@ namespace asset
 
 class IAssetManager;
 
+//! A class managing Asset's extra metadata context
+/**
+	Sometimes there may be nedeed attaching some metadata by a Loader
+	into Asset structure - that's why the class is defined.
+	Pay attention that it hasn't been done exactly yet, engine doesn't provide
+	metadata injecting.
+*/
 class IAssetMetadata : public core::IReferenceCounted
 {
 protected:
     virtual ~IAssetMetadata() = default;
 
 public:
-    //! this could actually be reworked to something more usable
+    //! This could actually be reworked to something more usable
     virtual const char* getLoaderName() = 0;
 };
 
@@ -131,22 +138,36 @@ private:
     void setMetadata(core::smart_refctd_ptr<IAssetMetadata>&& _metadata) { m_metadata = std::move(_metadata); }
 
 protected:
-    bool isDummyObjectForCacheAliasing;
+    bool isDummyObjectForCacheAliasing; //!< A bool for setting whether Asset is in dummy state. @see convertToDummyObject()
+
     //! To be implemented by base classes, dummies must retain references to other assets
     //! but cleans up all other resources which are not assets.
+	/**
+		Converting to a GPU object is equivalent of signing an Asset as dummy object.
+		If an Asset is being converted to a GPU object, its resources are no longer needed in RAM memory,
+		so everything it has allocated becomes deleted, but the Asset itself remains untouched, 
+		because it's needed as a key by some funtions that find GPU objects.
+		It involves all CPU objects (Assets).
+
+		So an Asset becomes GPU object and deletes some resources in RAM memory.
+	*/
     virtual void convertToDummyObject() = 0;
+
     //! Checks if the object is either not dummy or dummy but in some cache for a purpose
     inline bool isInValidState() { return !isDummyObjectForCacheAliasing /* || !isCached TODO*/; }
+
     //! Pure virtual destructor to ensure no instantiation
     virtual ~IAsset() = 0;
 
 public:
-    //! To be implemented by derived classes
+    //! To be implemented by derived classes. Returns a type of an Asset
     virtual E_TYPE getAssetType() const = 0;
-    //! 
+
+    //! Returning isDummyObjectForCacheAliasing, specifies whether Asset in dummy state
     inline bool isADummyObjectForCache() const { return isDummyObjectForCacheAliasing; }
 };
 
+//! A class storing Assets with the same type
 class SAssetBundle
 {
 	inline bool allSameTypeAndNotNull()
@@ -182,7 +203,7 @@ public:
 
 	//! Returning a type associated with current stored Assets
 	/**
-		An Asset type is specified in E_TYPE enum
+		An Asset type is specified in E_TYPE enum.
 		@see E_TYPE
 	*/
     inline IAsset::E_TYPE getAssetType() const { return m_contents->front()->getAssetType(); }
@@ -195,11 +216,13 @@ public:
 
     //! Whether this asset bundle is in a cache and should be removed from cache to destroy
     inline bool isInAResourceCache() const { return m_isCached; }
+
     //! Only valid if isInAResourceCache() returns true
     std::string getCacheKey() const { return m_cacheKey; }
 
 	//! Getting size of a collection of Assets stored by m_contents
     size_t getSize() const { return m_contents->size(); }
+
 	//! Checking if a collection of Assets stored by m_contents is empty
     bool isEmpty() const { return getSize()==0ull; }
 
--- a/include/irr/asset/IAsset.h
+++ b/include/irr/asset/IAsset.h
@@ -21,12 +21,14 @@ public:
     virtual const char* getLoaderName() = 0;
 };
 
-//! An interface class representing any kind of cpu-objects with capability of being cached
+//! An abstract data type class representing an interface of any kind of cpu-objects with capability of being cached
 /**
-	Actually it can be anything like cpu-side meshes scenes, texture data and material pipelines, 
+	Actually an Asset is a class deriving from it that can be anything like cpu-side meshes scenes, texture data and material pipelines, 
 	but must be serializable into/from .baw file format, unless it comes from an extension (irr::ext), 
 	so forth cached. There are different asset types you can find at IAsset::E_TYPE. 
 	IAsset doesn't provide direct instantiation (virtual destructor), much like IReferenceCounted.
+
+	Asset type's naming-convention is ICPU_x, where _x is a name of an Asset, eg. ICPUBuffer, ICPUMesh e.t.c
 
 	Every asset must be loaded by a particular class derived from IAssetLoader.
 	
@@ -42,9 +44,16 @@ public:
 
 	/**
 		Values of E_TYPE represents an Asset type.
-
-		Entire enum as type represents a register with bit flags,
-		which represent an available Asset type.
+		
+		Types are provided, so known is that the type you're casting to is right, eg.
+		If there is an Asset represeting ICPUBuffer, after calling a function returning
+		a type, it should return you a type associated with it, so
+		
+		\code{.cpp}
+		IAsset* asset;
+		if(asset->getType() == ET_BUFFER)
+			reinterpret_cast<ICPUBuffer*>(asset) // it is safe
+		\endcode
 
 		@see IAsset
 
@@ -62,10 +71,10 @@ public:
 		ET_SCENE represents
 		ET_IMPLEMENTATION_SPECIFIC_METADATA represents a special value used for things like terminating lists of this enum
 
-		Pay attention that a register is limited to 32 bits.
+		Pay attention that an Asset type represents one single bit, so there is a limit to 64 bits.
 
 	*/
-
+	
     enum E_TYPE : uint64_t
     {
         //! asset::ICPUBuffer
@@ -96,8 +105,9 @@ public:
         ET_IMPLEMENTATION_SPECIFIC_METADATA = 1u<<31u
         //! Reserved special value used for things like terminating lists of this enum
     };
-    constexpr static size_t ET_STANDARD_TYPES_COUNT = 12u; //!< The variable shows valid amount of available Asset types in E_TYPE bit register
+    constexpr static size_t ET_STANDARD_TYPES_COUNT = 12u; //!< The variable shows valid amount of available Asset types in E_TYPE enum
 
+	//! Returns a representaion of an Asset type in decimal system
     static uint32_t typeFlagToIndex(E_TYPE _type)
     {
         uint32_t type = (uint32_t)_type;
@@ -170,9 +180,9 @@ public:
         assert(allSameTypeAndNotNull());
     }
 
-	//! Returns an Asset type associated with the Asset
+	//! Returning a type associated with current stored Assets
 	/**
-		An Asset type is specified in bit register 
+		An Asset type is specified in E_TYPE enum
 		@see E_TYPE
 	*/
     inline IAsset::E_TYPE getAssetType() const { return m_contents->front()->getAssetType(); }
@@ -188,18 +198,18 @@ public:
     //! Only valid if isInAResourceCache() returns true
     std::string getCacheKey() const { return m_cacheKey; }
 
-	//! Getting size of an Asset stored by m_contents
+	//! Getting size of a collection of Assets stored by m_contents
     size_t getSize() const { return m_contents->size(); }
-	//! Checking if an Asset stored by m_contents is empty
+	//! Checking if a collection of Assets stored by m_contents is empty
     bool isEmpty() const { return getSize()==0ull; }
 
-	//! Overloaded operator checking if both Assets\b are\b the same objects in memory
+	//! Overloaded operator checking if both collections of Assets\b are\b the same arrays in memory
     bool operator==(const SAssetBundle& _other) const
     {
         return _other.m_contents == m_contents;
     }
 
-	//! Overloaded operator checking if both Assets\b aren't\b the same objects in memory
+	//! Overloaded operator checking if both collections of Assets\b aren't\b the same arrays in memory
     bool operator!=(const SAssetBundle& _other) const
     {
         return !((*this) != _other);
--- a/include/irr/asset/IAsset.h
+++ b/include/irr/asset/IAsset.h
@@ -33,6 +33,14 @@ protected:
 
 public:
     //! This could actually be reworked to something more usable
+	/*
+		To implement by user. Returns a Loader name that may attach some metadata into Asset structure.
+
+		@see IAssetMetadata
+
+		Due to external and custom Asset Loaders static_cast cannot be protected with a type enum comparision, 
+		so a string is provided.
+	*/
     virtual const char* getLoaderName() = 0;
 };
 
--- a/include/irr/asset/IAsset.h
+++ b/include/irr/asset/IAsset.h
@@ -100,37 +100,29 @@ public:
 	
     enum E_TYPE : uint64_t
     {
-        //! asset::ICPUBuffer
-        ET_BUFFER = 1u<<0u,
-        //! asset::CImageData - maybe rename to asset::CSubImageData
-        ET_SUB_IMAGE = 1u<<1u,
-        //! asset::ICPUTexture
-        ET_IMAGE = 1u<<2u,
-        //! asset::ICPUMeshBuffer
-        ET_SUB_MESH = 1u<<3u,
-        //! asset::ICPUMesh
-        ET_MESH = 1u<<4u,
-        //! asset::ICPUSkeleton - to be done by splitting CFinalBoneHierarchy
-        ET_SKELETON = 1u<<5u,
-        //! asset::ICPUKeyframeAnimation - from CFinalBoneHierarchy
-        ET_KEYFRAME_ANIMATION = 1u<<6u,
-        //! asset::ICPUShader
-        ET_SHADER = 1u<<7u,
-        //! asset::ICPUSpecializedShader
-        ET_SPECIALIZED_SHADER = 1u<<8,
-        //! asset::ICPUMeshDataFormatDesc
-        ET_MESH_DATA_DESCRIPTOR = 1u<<8,
-        //! reserved, to implement later
-        ET_GRAPHICS_PIPELINE = 1u<<9u,
-        //! reserved, to implement later
-        ET_SCENE = 1u<<10u,
-        //! lights, etc.
-        ET_IMPLEMENTATION_SPECIFIC_METADATA = 1u<<31u
+        ET_BUFFER = 1u<<0u,								//!< asset::ICPUBuffer
+        ET_SUB_IMAGE = 1u<<1u,						    //!< asset::CImageData - maybe rename to asset::CSubImageData
+        ET_IMAGE = 1u<<2u,								//!< asset::ICPUTexture
+        ET_SUB_MESH = 1u<<3u,							//!< asset::ICPUMeshBuffer
+        ET_MESH = 1u<<4u,								//!< asset::ICPUMesh
+        ET_SKELETON = 1u<<5u,							//!< asset::ICPUSkeleton - to be done by splitting CFinalBoneHierarchy
+        ET_KEYFRAME_ANIMATION = 1u<<6u,					//!< asset::ICPUKeyframeAnimation - from CFinalBoneHierarchy
+        ET_SHADER = 1u<<7u,								//!< asset::ICPUShader
+        ET_SPECIALIZED_SHADER = 1u<<8,					//!< asset::ICPUSpecializedShader
+        ET_MESH_DATA_DESCRIPTOR = 1u<<8,				//!< asset::ICPUMeshDataFormatDesc
+        ET_GRAPHICS_PIPELINE = 1u<<9u,					//!< reserved, to implement later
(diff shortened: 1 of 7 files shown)


include/irr/asset/IAsset.h
```diff
@@ -21,9 +21,51 @@ class IAssetMetadata : public core::IReferenceCounted
...
 };
 
+//! An interface class representing any kind of cpu-objects with capability of being cached
+/**
+	Actually it can be anything like cpu-side meshes scenes, texture data and material pipelines, 
+	but must be serializable into/from .baw file format, unless it comes from an extension (irr::ext), 
+	so forth cached. There are different asset types you can find at IAsset::E_TYPE. 
+	IAsset doesn't provide direct instantiation (virtual destructor), much like IReferenceCounted.
+
+	Every asset must be loaded by a particular class derived from IAssetLoader.
```
devshgraphicsprogramming, line 64: Not true, IAsset is just an ADT, so its not an object in itself (cannot be instantiated). But concrete classes (such as ICPUBuffer, ICPUMeshBuffer, etc.) derived from `IAsset` can be instantiated via `new` or `core::make_smart_refctd_ptr`

include/irr/asset/IAsset.h
```diff
@@ -21,9 +21,51 @@ class IAssetMetadata : public core::IReferenceCounted
...
+		ET_SKELETON represents
+		ET_KEYFRAME_ANIMATION represents
+		ET_SHADER represents
+		ET_SPECIALIZED_SHADER represents
+		ET_MESH_DATA_DESCRIPTOR represents
+		ET_GRAPHICS_PIPELINE represents
+		ET_SCENE represents
+		ET_IMPLEMENTATION_SPECIFIC_METADATA represents a special value used for things like terminating lists of this enum
+
+		Pay attention that a register is limited to 32 bits.
```
devshgraphicsprogramming: not a register, wrong word, and its a `uint64_t` just by the way

include/irr/asset/IAsset.h
```diff
@@ -21,9 +21,51 @@ class IAssetMetadata : public core::IReferenceCounted
...
+
 class IAsset : virtual public core::IReferenceCounted
 {
 public:
+
+	/**
+		Values of E_TYPE represents an Asset type.
+
+		Entire enum as type represents a register with bit flags,
+		which represent an available Asset type.
```
devshgraphicsprogramming: its provided so you know whether the type you're `static_cast`ing to is right

include/irr/asset/IAsset.h
```diff
@@ -21,9 +21,51 @@ class IAssetMetadata : public core::IReferenceCounted
...
+
+		@see IAsset
+
+		ET_BUFFER represents asset::ICPUBuffer
```
devshgraphicsprogramming, line 91: if `IAsset* asset;` and `asset->getType()==ET_BUFFER` then you can `static_cast<ICPUBuffer*>(asset)` safely

include/irr/asset/IAsset.h
```diff
@@ -128,8 +170,14 @@ class SAssetBundle
         assert(allSameTypeAndNotNull());
     }
 
+	//! Returns an Asset type associated with the Asset
```
devshgraphicsprogramming: Bundle can only hold an array of assets that are the same type

include/irr/asset/IAsset.h
```diff
@@ -140,13 +188,18 @@ class SAssetBundle
     //! Only valid if isInAResourceCache() returns true
     std::string getCacheKey() const { return m_cacheKey; }
 
+	//! Getting size of an Asset stored by m_contents
```
devshgraphicsprogramming: its an SAssetBundle not an Asset

include/irr/asset/IAsset.h
```diff
@@ -140,13 +188,18 @@ class SAssetBundle
...
     size_t getSize() const { return m_contents->size(); }
+	//! Checking if an Asset stored by m_contents is empty
     bool isEmpty() const { return getSize()==0ull; }
 
+	//! Overloaded operator checking if both Assets\b are\b the same objects in memory
```
devshgraphicsprogramming: Bundles are indentical if array pointers match, not if contents match (makes dealing with them a lot faster)

include/irr/asset/IAssetLoader.h
```diff
@@ -7,9 +7,47 @@
 namespace irr { namespace asset
 {
 
+//! A class that defines rules during Asset-loading process
+/**
+	Every Asset must be loaded by a particular class derived from IAssetLoader.
```
devshgraphicsprogramming, line 12: Not true, the particular assets can be instantiated and filled out by hand. Loaders just automate this process and can create assets from "files" (or any other resource interfacing as a file).

include/irr/asset/IAssetLoader.h
```diff
@@ -7,9 +7,47 @@
...
+	@see IAsset
+	@see IAssetManager
+	@see IAssetWriter
+*/
+
 class IAssetLoader : public virtual core::IReferenceCounted
 {
 public:
+
+	//! Caching and resource duplication flags
```
devshgraphicsprogramming, line 50: its a rather weird bitfield, we'd have to explain how the loading process works.

Basically a mesh can reference a submesh, a submesh can reference a material, a material can reference a texture, etc.

If you look at them as recursive functions, the `LEVEL` is basically how deep down the stack you are.
This flag is a bitfield with 2 bits per level, and the enums provie are just some useful constants. But different combinations are valid as well.

Example
```cpp
static_cast<E_CACHING_FLAGS>(ECF_DONT_CACHE_TOP_LEVEL>>4ull) // anything on level 2 (top is 0) will not get cached
```

include/irr/asset/IAssetLoader.h
```diff
@@ -7,9 +7,47 @@
...
+
 class IAssetLoader : public virtual core::IReferenceCounted
 {
 public:
+
+	//! Caching and resource duplication flags 
+	/**
+		They have an impact on loading an Asset.
+
+		E_CACHING_FLAGS::ECF_CACHE_EVERYTHING means that //TODO
```
devshgraphicsprogramming: default, if you can find the asset in the previously cached asset, then you will return it and if you cant you will load it and add it to the cache before returning

include/irr/asset/IAssetManager.h
```diff
@@ -31,6 +31,18 @@ namespace asset
     std::function<void(SAssetBundle&)> makeAssetGreetFunc(const IAssetManager* const _mgr);
     std::function<void(SAssetBundle&)> makeAssetDisposeFunc(const IAssetManager* const _mgr);
 
+	//! Class responsible for handling loading of assets from file system or other resources
+	/**
+		It provides the default methods for converting CPU assets to GPU objects, methods
```
devshgraphicsprogramming: Nope it doesn't CPU 2 GPU conversion is in the `video::` namespace

include/irr/asset/IAssetManager.h
```diff
@@ -31,6 +31,18 @@ namespace asset
     std::function<void(SAssetBundle&)> makeAssetGreetFunc(const IAssetManager* const _mgr);
     std::function<void(SAssetBundle&)> makeAssetDisposeFunc(const IAssetManager* const _mgr);
 
+	//! Class responsible for handling loading of assets from file system or other resources
+	/**
+		It provides the default methods for converting CPU assets to GPU objects, methods 
+		for exporting CPU assets, as well as synthetic asset creators like the IGeometryCreator.
+		There is a loading, writing and creation functionality that is thread-safe.
```
devshgraphicsprogramming: Thread safety still has one issue, if you start loading the same asset at the exact same time (they will both check cache and not find anything), you may end up with two copies in the cache.

include/irr/asset/IAsset.h
AnastaZIuk: actually shouldn't it be reinterpret_cast, should it?

Crisspl: nah, never use reinterpret_cast unless it's an absolute must

devshgraphicsprogramming: `static_cast` works beautifully because it can calculate the relative offset to the pointer accounting for the difference between the start of storage of a base class and a derived class.

Example
```cpp
class A {uint32_t a;};
class B {uint32_t b;};

class C : public A, public B {};

B* base;
C* derived = static_cast<C*>(base);
```


if you print both `base` and `derived` as `%p` you will see that the pointer to derived has been rolled back by 4 bytes.


`reinterpret_cast` is only for casting void pointers really.

include/irr/asset/IAsset.h
devshgraphicsprogramming: still need a word about that you can create them outside of loaders

include/irr/asset/IAssetLoader.h
```diff
@@ -142,10 +155,8 @@ class IAssetLoader : public virtual core::IReferenceCounted
         }
 
         //! Called before loading a file
```
devshgraphicsprogramming: Need a description

AnastaZIuk: I will cover it soon

include/irr/asset/IAsset.h
AnastaZIuk: gonna cover it soon

include/irr/asset/IAsset.h
```diff
@@ -11,13 +11,20 @@ namespace asset
 
 class IAssetManager;
 
+//! A class managing Asset's extra metadata context
+/**
+	Sometimes there may be nedeed attaching some metadata by a Loader
```
devshgraphicsprogramming, line 16: clarify

include/irr/asset/IAsset.h
```diff
@@ -11,13 +11,20 @@ namespace asset
...
+	metadata injecting.
+*/
 class IAssetMetadata : public core::IReferenceCounted
 {
 protected:
     virtual ~IAssetMetadata() = default;
 
 public:
-    //! this could actually be reworked to something more usable
+    //! This could actually be reworked to something more usable
```
devshgraphicsprogramming, line 35: because of external and custom asset loaders we cannot protect a static_cast with a type enum comparison.. this is why we provide a string.

include/irr/asset/IAsset.h
AnastaZIuk: I think it's clear, isn't it?
@Crisspl

include/irr/asset/IAsset.h
```diff
@@ -131,22 +138,36 @@ class IAsset : virtual public core::IReferenceCounted
     void setMetadata(core::smart_refctd_ptr<IAssetMetadata>&& _metadata) { m_metadata = std::move(_metadata); }
 
 protected:
-    bool isDummyObjectForCacheAliasing;
+    bool isDummyObjectForCacheAliasing; //!< A bool for setting whether Asset is in dummy state. @see convertToDummyObject()
+
     //! To be implemented by base classes, dummies must retain references to other assets
     //! but cleans up all other resources which are not assets.
+	/**
+		Converting to a GPU object is equivalent of signing an Asset as dummy object.
```
devshgraphicsprogramming: Not necessairly true.

It may be desireable to have a cache of GPU objects associated with the CPU objects, while at the same time being able to get rid of the RAM overhead for storage of such a CPU object.

This function will get rid of as much memory as possible (including references to CPU objects it needs), the only thing that stays is just the pointer and the memory allocated for that pointer (whole class).

After this operation we can treat the asset contents as invalid garbage.

The default CPU to GPU object converter provided by the engine does this.

include/irr/asset/IAssetManager.h
```diff
@@ -165,8 +165,13 @@ namespace asset
 			}
 		}
 
+		//TODO change name
         //! _supposedFilename is filename as it was, not touched by loader override with _override->getLoadFilename()
-        //TODO change name
+		/**
+			Attempting of getting an Asset. Valid loaders for certein extension are searched. If none of those can't
```
devshgraphicsprogramming: "Attempts to fetch an Asset."

"certein" is a typo -> certain

include/irr/asset/IAssetManager.h
```diff
@@ -165,8 +165,13 @@ namespace asset
...
+			deal with file with certein extension adjusted to found loaders, then the tryout is performed, and the funtion
+			iterates through all available loaders counting on that one of them might work.
```
devshgraphicsprogramming: rephrase

include/irr/asset/ICPUBuffer.h
```diff
@@ -18,6 +18,14 @@ namespace irr
 namespace asset
 {
 
+//! One of CPU class-object representing an Asset
+/**
+	An Asset is a CPU object which needs to be loaded for caching it to RAM memory,
```
devshgraphicsprogramming: This Asset

devshgraphicsprogramming: doesn't need to be converted, there are some functions in the engine which take ICPUBuffer as a nice way to represent a plain byte array.

Its used for storage of large arrays, so that storage can be decoupled from other objects such as meshbuffers, images, animations and shader source/bytecode.

include/irr/core/memory/dynamic_array.h
```diff
@@ -27,6 +27,17 @@ namespace impl
 	};
 }
 
+//! Class for array type, that allocates memory one time dynamically for specified constant amount of objects
+/**
+	An array is allocated dynamically, so not on stack.
+	There is no case you can change the size of such an array in.
```
devshgraphicsprogramming: "no case in which"

include/irr/core/memory/dynamic_array.h
```diff
@@ -27,6 +27,17 @@ namespace impl
...
+	The adventage of this class is core::dynamic_array has constant storage size,
+	so only one allocation is performed once (member and data storage on single allocation),
+	instead of unnecessary 2 allocations that might appear, if there were not such a class
```
devshgraphicsprogramming: compared to an `std::vector<>`

include/irr/core/memory/dynamic_array.h
```diff
@@ -27,6 +27,17 @@ namespace impl
...
+	handling it.
+
```
devshgraphicsprogramming, line 46: and compared to `std::array` its size can be determined at runtime.

As a consequence `sizeof(dynamic_array<T,allocator>)` is completely meaningless, and it can only be allocated on the heap and is non-copyable.

include/irr/core/memory/refctd_dynamic_array.h
```diff
@@ -10,6 +10,20 @@ namespace irr
...
+//! Class for array type, that allocates memory one time dynamically for specified constant amount of objects
+/**
+	An array is allocated dynamically, so not on stack.
+	There is no case you can change the size of such an array in.
+	The adventage of this class is core::dynamic_array has constant storage size,
+	so only one allocation is performed once (member and data storage on single allocation),
+	instead of unnecessary 2 allocations that might appear, if there were not such a class
+	handling it.
+
+	Pay attention it additionally derives from IReferenceCounted, so all adventages and behaviours of IReferenceCounted are provided indeed.
```
devshgraphicsprogramming: the purpose of this class is to compensate for the non-copyability of the base class and allow "pass by reference" (shared contents) without memory leaks and going out of scope.

include/irr/asset/IAsset.h
AnastaZIuk: That's why I described that Assets are needed as a key, so they cannot be removed entirely

devshgraphicsprogramming: yes, but this can be read very wrong... this class method has nothing to do with CPU to GPU object conversion (its needed by the default, but nothing to do with that)

Its only for getting rid of memory and keeping the pointer alive

include/irr/asset/IAsset.h
devshgraphicsprogramming: the english/grammar is not clear here

devshgraphicsprogramming: Meta = above, higher

This is for attaching extra data to an asset that the loader retrieved, which are not ubiquitously representable by the engine. These could be but not limited to (note that categories overlap):
- Global Data about the file or scene
- IDs, names
- Default View/Projection
- Complex Animations and Hierarchies
- Physics Simulation Data
- AI data
- Lighting
- Extra material metadata

There is total flexibility here, its expected that each loader has its own base metadata class implementing the IAssetMetadata interface, with its own type enum that other loader's metadata classes derive from the base.

There's no attempt to standardise the metadata, as loaders are so diverse.

include/irr/asset/IAsset.h
AnastaZIuk: well, I will move it

include/irr/asset/IAssetLoader.h
```diff
@@ -155,6 +180,16 @@ class IAssetLoader : public virtual core::IReferenceCounted
...
+		/**
+			\param inOutFilename is a path to file Asset will be loaded from.
```
devshgraphicsprogramming: the filename argument goes in, the actual resolved path comes out (the override decides how to resolve a local path or even a URL into a "proper" filename).

Hierarchy level is more like a stack counter, we expect that that asset loading will get called recursively, example mesh needs material, material needs texture, etc. so `getLoadFilename` could be called separately for each dependent resource from deeper recursions in the loading stack.

include/irr/asset/IAsset.h
```diff
@@ -100,37 +100,29 @@ class IAsset : virtual public core::IReferenceCounted
...
+        ET_SCENE = 1u<<10u,								//!< reserved, to implement later
+        ET_IMPLEMENTATION_SPECIFIC_METADATA = 1u<<31u	//!< lights, etc.
         //! Reserved special value used for things like terminating lists of this enum
     };
     constexpr static size_t ET_STANDARD_TYPES_COUNT = 12u; //!< The variable shows valid amount of available Asset types in E_TYPE enum
 
 	//! Returns a representaion of an Asset type in decimal system
+	/**
+		Each value is returned from the range 0 to (ET_STANDARD_TYPES_COUNT - 1) to provide
+		easy way to handle arrays operations. For instance ET_SUB_IMAGE returns 1 and ET_MESH
```
devshgraphicsprogramming: replace "arrays operations" to array indexing

include/irr/asset/IAsset.h
```diff
@@ -141,9 +133,18 @@ class IAsset : virtual public core::IReferenceCounted
 
     IAsset() : isDummyObjectForCacheAliasing{false}, m_metadata{nullptr} {}
 
+	//! Returns whole size associated with an Asset and its data
+	/**
+		The size is used to determine compression level while writing process 
+		is performed. As you expect, the more size Asset has, the more compression
```
devshgraphicsprogramming: Its actually necessary to reserve the correct amount of memory when writing to file, "conservative" here actually means that the value returned can be greater than memory actually needed.

What you've written about is actually a secondary functionality,
also, the bigger the asset the more likely it is to be compressed with a more expensive (slower) algorithm (replace the badly formed sentence ending).

include/irr/asset/IAssetWriter.h
```diff
@@ -20,17 +20,10 @@ namespace irr { namespace asset
 */
 enum E_WRITER_FLAGS : uint32_t
```
devshgraphicsprogramming, line 26: its important to note that not every writer supports all these flags and may just silently drop them when writing (example, JPG, PNG do not write encrypted images)

include/irr/asset/IAsset.h
devshgraphicsprogramming: status?

include/irr/asset/IAsset.h
devshgraphicsprogramming: status?

include/irr/asset/IAsset.h
devshgraphicsprogramming: status?

include/irr/asset/IAssetLoader.h
devshgraphicsprogramming: status?

include/irr/asset/IAsset.h
devshgraphicsprogramming: status?

include/irr/asset/IAsset.h
devshgraphicsprogramming: status?

include/irr/asset/IAssetManager.h
devshgraphicsprogramming: status?

include/irr/asset/IAssetManager.h
devshgraphicsprogramming: status?

include/irr/asset/ICPUBuffer.h
devshgraphicsprogramming: status?

include/irr/core/memory/dynamic_array.h
devshgraphicsprogramming: status?

include/irr/core/memory/dynamic_array.h
devshgraphicsprogramming: status?

include/irr/core/memory/dynamic_array.h
devshgraphicsprogramming: status?

include/irr/core/memory/refctd_dynamic_array.h
devshgraphicsprogramming: status?

include/irr/asset/IAssetLoader.h
devshgraphicsprogramming: status?

include/irr/asset/IAsset.h
devshgraphicsprogramming: status?

include/irr/asset/IAsset.h
devshgraphicsprogramming: status?

include/irr/asset/IAssetWriter.h
devshgraphicsprogramming: status?

include/irr/asset/IAsset.h
AnastaZIuk: Solved

include/irr/asset/IAsset.h
AnastaZIuk: Solved

include/irr/asset/IAsset.h
AnastaZIuk: Solved

include/irr/asset/IAssetLoader.h
AnastaZIuk: Solved

include/irr/asset/IAsset.h
AnastaZIuk: Solved

include/irr/asset/IAssetManager.h
AnastaZIuk: Solved

include/irr/asset/IAssetManager.h
AnastaZIuk: Solved

include/irr/asset/ICPUBuffer.h
AnastaZIuk: Solved

include/irr/core/memory/dynamic_array.h
AnastaZIuk: Solved

include/irr/core/memory/dynamic_array.h
AnastaZIuk: Solved

include/irr/core/memory/dynamic_array.h
AnastaZIuk: Solved

include/irr/core/memory/refctd_dynamic_array.h
AnastaZIuk: Solved

include/irr/asset/IAssetLoader.h
AnastaZIuk: Solved

include/irr/asset/IAsset.h
AnastaZIuk: Solved

devshgraphicsprogramming:
resolved everything, merge to other fork

PR #386, opened 2019-10-23
Add loaderFlags and CBufferLoaderBIN

AnastaZIuk:


include/irr/asset/IAssetLoader.h
```diff
@@ -91,13 +109,14 @@ class IAssetLoader : public virtual core::IReferenceCounted
...
+            : decryptionKeyLen(_decryptionKeyLen), decryptionKey(_decryptionKey), cacheFlags(_cacheFlags), loaderFlags(_loaderFlags)
         {
         }
-        size_t decryptionKeyLen;			 //!< The size of decryptionKey
-        const uint8_t* decryptionKey;		 //!< The key it used to decrypt potentially encrypted files
-        const E_CACHING_FLAGS cacheFlags;	 //!< Flags defining rules during loading process
+        size_t decryptionKeyLen;				    	//!< The size of decryptionKey
+        const uint8_t* decryptionKey;				    //!< The key it used to decrypt potentially encrypted files
+        const E_CACHING_FLAGS cacheFlags;		     	//!< Flags defining rules during loading process
+		const E_LOADER_PARAMETER_FLAGS loaderFlags;		//!< Flags having an impact on extraordinary tasks during loading process
```
AnastaZIuk, line 119: @devshgraphicsprogramming shouldn't it actually be a uint64_t, should it? I mean, you could use then || and specify more than one flag at once

devshgraphicsprogramming: nope all good.

include/irr/asset/IAssetLoader.h
```diff
@@ -78,6 +78,24 @@ class IAssetLoader : public virtual core::IReferenceCounted
...
     };
 
+	//! Parameter flags for a loader
+	/**
+		These are extra flags that have an impact on extraordinary tasks while loading.
+
+		E_LOADER_PARAMETER_FLAGS::ELPF_NONE is default and means that there is nothing to perform.
+		E_LOADER_PARAMETER_FLAGS::ELPF_RIGHT_HANDED_MESHES specifies that a mesh will be flipped in such 
+		a way that it'll look correctly in right-handed camera system. If it isn't set, left-handed camera system
+		becomes valid.
```
devshgraphicsprogramming, line 88: If it isn't set, compatibility with left-handed coordinate camera is assumed.

include/irr/asset/IAssetLoader.h
```diff
@@ -91,13 +109,14 @@ class IAssetLoader : public virtual core::IReferenceCounted
 
     struct SAssetLoadParams
     {
-        SAssetLoadParams(const size_t& _decryptionKeyLen = 0u, const uint8_t* _decryptionKey = nullptr, const E_CACHING_FLAGS& _cacheFlags = ECF_CACHE_EVERYTHING)
-            : decryptionKeyLen(_decryptionKeyLen), decryptionKey(_decryptionKey), cacheFlags(_cacheFlags)
+        SAssetLoadParams(const size_t& _decryptionKeyLen = 0u, const uint8_t* _decryptionKey = nullptr, const E_CACHING_FLAGS& _cacheFlags = ECF_CACHE_EVERYTHING, const E_LOADER_PARAMETER_FLAGS _loaderFlags = ELPF_NONE)
```
devshgraphicsprogramming, line 112: const reference, not just a const

src/irr/asset/CBufferLoaderBIN.h
```diff
@@ -0,0 +1,44 @@
...
+			public:
+				virtual bool isALoadableFileFormat(io::IReadFile* _file) const override;
+
+				virtual const char** getAssociatedFileExtensions() const override
+				{
+					static const char* extensions[]{ "bin", nullptr };
+					return extensions;
+				}
+
+				//virtual uint64_t getSupportedAssetTypesBitfield() const override { return asset::IAsset::TODO; } should I add new IAsset as binary?
```
devshgraphicsprogramming, line 29: `ET_BUFFER` it already exists

src/irr/asset/CBufferLoaderBIN.cpp
```diff
@@ -0,0 +1,67 @@
...
+	{
+		asset::SAssetBundle CBufferLoaderBIN::loadAsset(io::IReadFile* _file, const asset::IAssetLoader::SAssetLoadParams& _params, asset::IAssetLoader::IAssetLoaderOverride* _override, uint32_t _hierarchyLevel)
+		{
+			if (!_file)
+				return {};
+
+			SContext ctx;
+
+			ctx.files = _file;
+			ctx.files->grab();
```
devshgraphicsprogramming, line 15: no need to grab, you only grab if you expect to hold onto the object after the function returns.

src/irr/asset/CBufferLoaderBIN.cpp
```diff
@@ -0,0 +1,67 @@
...
+
+			uint16_t index{};
+
+			for (io::IReadFile* file = ctx.files; file; ++file)
```
devshgraphicsprogramming, line 19: wtf? there's only ever one file to read!

src/irr/asset/CBufferLoaderBIN.h
```diff
@@ -0,0 +1,44 @@
...
+
+				virtual asset::SAssetBundle loadAsset(io::IReadFile* _file, const asset::IAssetLoader::SAssetLoadParams& _params, asset::IAssetLoader::IAssetLoaderOverride* _override = nullptr, uint32_t _hierarchyLevel = 0u) override;
+
+			private:
+				struct SContext
+				{
+					io::IReadFile* files;
+					core::vector<core::smart_refctd_ptr<char>> sourceCodeBuffersCache;
```
devshgraphicsprogramming, line 37: what on earth is this?

src/irr/asset/CBufferLoaderBIN.cpp
```diff
@@ -0,0 +1,67 @@
...
+			{
+				ctx.sourceCodeBuffersCache.emplace_back(core::make_smart_refctd_ptr<char>(_IRR_NEW_ARRAY(char, file->getSize())));
+				file->read(ctx.sourceCodeBuffersCache[index++].get(), file->getSize());
+			}
+
+			return SAssetBundle(ctx.sourceCodeBuffersCache); // is it valid actually?
```
devshgraphicsprogramming, line 25: the SAssetBundle should return exactly one `ICPUBuffer`

src/irr/asset/CBufferLoaderBIN.cpp
```diff
@@ -0,0 +1,67 @@
...
+
+		bool CBufferLoaderBIN::isALoadableFileFormat(io::IReadFile* _file) const
+		{
+			std::string filePath(_file->getFileName().c_str());
+			uint16_t beggingOfFileExtIndex{};
+
+			auto getIndexInArrayForFileExtension = [&](auto& index)
+			{
+				for (uint16_t i = filePath.size() - 1; i > 0u; --i)
+					if (filePath[i] == '.')
```
devshgraphicsprogramming, line 36: we already have a function for getting the file extension, see how other loaders use it.

src/irr/asset/CBufferLoaderBIN.cpp
```diff
@@ -0,0 +1,67 @@
...
+			for (uint16_t i = beggingOfFileExtIndex; i < filePath.size(); ++i)
+				fileExtension += filePath[i];
+
+			auto availableExtensions(getAssociatedFileExtensions());
+			for (const char* extension = const_cast<char*>(availableExtensions[0]); extension; ++extension)
+			{
+				auto isFormatValid = [&](const char* _ext)
+				{
+					std::string checkedExtension(_ext);
+					if (checkedExtension == fileExtension)
```
devshgraphicsprogramming, line 56: You do realize that the whole point of `const char** getAssociatedFileExtensions()` is that the `IAssetManager` does all this for you?

The `bool isALoadableFileFormat(io::IReadFile* _file) const` function should only check the ability to load the file based on the file contents (not the extension) such as the presence of a header, etc.

In this case we can obviously see that the function need always `return true` because a .bin file does not have a header.

devshgraphicsprogramming:
I actually wanted these commits to be applied on top of `master` branch and a PR into `master`

PR #389, opened 2019-10-25
Add loader flags, CBufferLoaderBIN and flipping around X axie

AnastaZIuk:


src/irr/asset/CBufferLoaderBIN.h
```diff
@@ -0,0 +1,44 @@
...
+
+#include "irr/asset/IAssetLoader.h"
+#include "irr/asset/ICPUMeshBuffer.h"
+
+namespace irr
+{
+	namespace asset
+	{
+		//! Binaryloader capable of loading source code in binary format
+		class CBufferLoaderBIN : public asset::IAssetLoader
```
devshgraphicsprogramming: I guess you can write `class final` and remove all the `virtual` keywords from the methods

src/irr/asset/CBufferLoaderBIN.cpp
```diff
@@ -0,0 +1,25 @@
...
+	{
+		asset::SAssetBundle CBufferLoaderBIN::loadAsset(io::IReadFile* _file, const asset::IAssetLoader::SAssetLoadParams& _params, asset::IAssetLoader::IAssetLoaderOverride* _override, uint32_t _hierarchyLevel)
+		{
+			if (!_file)
+				return {};
+
+			SContext ctx(_file->getSize());
+			ctx.file = _file;
+
+			ctx.file->read(ctx.sourceCodeBuffer.get(), ctx.file->getSize());
```
devshgraphicsprogramming: this is going to crash badly.

You need `.get()->getPointer()`

src/irr/asset/CSTLMeshFileLoader.cpp
```diff
@@ -75,8 +75,26 @@ asset::SAssetBundle CSTLMeshFileLoader::loadAsset(io::IReadFile* _file, const as
         {
         core::vectorSIMDf n;
 		getNextVector(_file, n, binary);
-        normals.push_back(core::normalize(n));
-        }
+
+		auto performActionBasedOnOrientationSystem = [&](auto performOnRightHanded, auto performOnLeftHanded = [&](void) {})
+		{
+			switch (_params.loaderFlags)
```
devshgraphicsprogramming: nope `if (_params.loaderFlags&E_LOADER_PARAMETER_FLAGS::ELPF_RIGHT_HANDED_MESHES)`

src/irr/asset/CSTLMeshFileLoader.cpp
```diff
@@ -75,8 +75,26 @@ asset::SAssetBundle CSTLMeshFileLoader::loadAsset(io::IReadFile* _file, const as
...
+				}
+				default:
+				{
+					performOnLeftHanded();
+					break;
+				}
+			}
+		};
+
+		performActionBasedOnOrientationSystem([&]() {n.z = -n.z;}, [&]() {});
```
devshgraphicsprogramming: why are you flipping around .Z and not .X ?

src/irr/asset/CSTLMeshFileLoader.cpp
```diff
@@ -94,9 +112,10 @@ asset::SAssetBundle CSTLMeshFileLoader::loadAsset(io::IReadFile* _file, const as
                     return {};
 			}
 			getNextVector(_file, p[i], binary);
+			performActionBasedOnOrientationSystem([&]() {p[i].z = -p[i].z; }, [&]() {});
 		}
-        for (uint32_t i = 0u; i < 3u; ++i) // seems like in STL format vertices are ordered in clockwise manner...
-            positions.push_back(p[2u-i]);
+		for (uint32_t i = 0u; i < 3u; ++i) // seems like in STL format vertices are ordered in clockwise manner...
```
devshgraphicsprogramming, line 108: STL is completely non indexed?

AnastaZIuk: it seems it is non indexed

src/irr/asset/CXMeshFileLoader.cpp
```diff
@@ -225,12 +225,12 @@ core::matrix4x3 getGlobalMatrix_evil(asset::ICPUSkinnedMesh::SJoint* joint)
 }
 
 
-bool CXMeshFileLoader::load(SContext& _ctx, io::IReadFile* file)
+bool CXMeshFileLoader::load(SContext& _ctx, io::IReadFile* file, const asset::IAssetLoader::SAssetLoadParams& _params)
```
AnastaZIuk, line 228: @devshgraphicsprogramming  should _params be a const& member or should it be passed through functions?

devshgraphicsprogramming: and which one do you think will crash if you attempt to call the loader from multiple threads?

src/irr/asset/CXMeshFileLoader.cpp
```diff
@@ -1504,9 +1554,21 @@ bool CXMeshFileLoader::parseDataObjectMeshNormals(SContext& _ctx, SXMesh &mesh)
...
+						mesh.Vertices[mesh.Indices[normalidx++]].Normal.set(normals[polygonfaces[jk + 1]]);
+						mesh.Vertices[mesh.Indices[normalidx++]].Normal.set(normals[polygonfaces[0]]);
+					},
+					[&]()
+					{
+						mesh.Vertices[mesh.Indices[normalidx++]].Normal.set(normals[polygonfaces[0]]);
+						mesh.Vertices[mesh.Indices[normalidx++]].Normal.set(normals[polygonfaces[jk + 1]]);
+						mesh.Vertices[mesh.Indices[normalidx++]].Normal.set(normals[polygonfaces[jk + 2]]);
+					}
+				);
```
devshgraphicsprogramming: where's your flip of the position and the normal around the X axis?

Also you shouldn't be inverting indices here, you'll end up premuting them twice... you'll get the same output as a result when rendering.

AnastaZIuk: oh right, I forgot about normals flip and did premuting twice actually

AnastaZIuk: there is a pos flip, but sure, rest to fix

src/irr/asset/CXMeshFileLoader.cpp
AnastaZIuk: being honest I don't know, I haven't done anything with multithreading before

devshgraphicsprogramming:
how on earth were you able to merge into my master?

src/irr/asset/CXMeshFileLoader.cpp
devshgraphicsprogramming: you're permuting the indices, and then you're permuting how normals and positions are read according to the indices... two reversals of indices == no reversal

src/irr/asset/CXMeshFileLoader.cpp
devshgraphicsprogramming: well if you have a member and the loader is shared between threads, what's gonna happen?

AnastaZIuk: I think that operations on such a member won't be atomic, right? So it could crash

devshgraphicsprogramming: well done, now you know why every loader has a context struct and passes a reference to it as a parameter.... rule nr. 1 of multi-threading, avoid global states.

src/irr/asset/CXMeshFileLoader.cpp
AnastaZIuk: https://github.com/buildaworldnet/IrrlichtBAW/pull/389/commits/8b150696b2c9791cf37751d8b49c05c0b7eff34d

there is only one index mirrowing (last commit i removed double permuting), normals (added in last) and possitions are flipped as well

devshgraphicsprogramming: where are the positions?

AnastaZIuk: here there are my boss! 
https://github.com/buildaworldnet/IrrlichtBAW/pull/389/files#diff-cb7f186202d159b4b05d2bdcc4c37d6eR989 positions
https://github.com/buildaworldnet/IrrlichtBAW/pull/389/files#diff-cb7f186202d159b4b05d2bdcc4c37d6eR1500 normals
https://github.com/buildaworldnet/IrrlichtBAW/pull/389/files#diff-cb7f186202d159b4b05d2bdcc4c37d6eR1031 indicies

Issue #382, opened 2019-10-18
BSDF Importance Sampling Validation


achalpandeyy:
A simple utility to validate that the formulae used for the 3 functions needed for integration using Multiple Importance Sampling are consistent for every observer position (above and below the surface):

1. The BSDF probability density function over a full sphere (known as `eval` in Cycles)
2. The BSDF importance sampling function that transforms uniformly distributed random numbers in `[0,1]^2` into some distribution X (known as `sample` in Cycles)
3. The probability density function of X that will be used for MIS weights

The application will allow the user to open their own text file with GLSL language definitions of the 3 functions (with fixed BSDF parameters to constants) and produce a 2D image mapped to a surface of a sphere produced by splatting `GL_POINTS` using additive blending the intensity of function `3` in positions determined by function `2` and will compare the RMSE against the analytical evaluation of function `1`.

This will not be anything groundbreaking in itself, more like a Unit Test for the functions we will use in the Mitsuba Scene Rendering application.

devshgraphicsprogramming:
Current examples that validate importance sampling via derivatives work just fine.

Issue #384, opened 2019-10-18
How to write good mesh loaders


AnastaZIuk:

devshgraphicsprogramming:
Tagging @Crisspl (subscribe plz)

devshgraphicsprogramming:
# Split Your Loaders

Example, OBJ loader that loads textures from materials specified in an `.mtl` file. You should have 3+ Loaders:
1. OBJ loader that only reads `.obj` produces an `asset::ICPUMesh` necessarily invokes an MTL loader
2. MTL loader that only reads `.mtl` files and produces (or returns) a half-filled `asset::ICPURenderpassIndependentPipeline` (that OBJ can modify slightly for raster and vertex input parameters), naturally it invokes image loaders for needed textures. 
3. Texture Loaders for PNG, JPEG, etc. 

# If something cannot be passed within the returned asset bundle, use metadata to communicate between loaders

Example OBJ loader will probably want to know the material parameters that go with an MTL pipeline, but a pipeline does not include the actual parameters used such as "push constants" or "descriptor sets" only the layouts (generic material definition, not specific).

So the specific material definitions should come in the MTL-specific-metadata-derived-class that OBJ loader could use.

# Share `ICPUDescriptorSetLayout` as much as possible

You can over-declare descriptor set layouts, i.e. a set that has every single resource and more that a shader could use (all the textures, all the buffers for all shader types).

Its completely okay to have a descriptor in a set that is not used by the shader (not the other way round), and its okay not to populate a descriptor set binding that was specified in the layout as long as its not statically used by the shader (dead code).
(@Crisspl you may have to fix up the descriptor set code on OpenGL so it sets up the correct arguments for non-populated descriptor)

This helps the engine do less strenous resource rebinds (see Vulkan pipeline compatibility rules).

# Don't create copious amount of shaders, use Ubershader whenever possible (needs an engine extension, c.f. bottom of post)

Most formats such as STL, PLY, OBJ and even glTF declare a very limited set of materials.

You WILL find that instead of creating a new `ICPUSpecializedShader` for every single new `ICPURenderpassIndependentPipeline`, it would make much more sense to create the few `ICPUSpecializedShaders` and even `ICPUDescriptorSetLayout` as well as pipeline layouts in the constructor of the loader and cache them under a special path (similar to GLSL built-in includes) in the asset manager cache (then remove upon loader destruction).

[...]

PR #390, opened 2019-10-25
BSDFValidator starter files

achalpandeyy:
First steps towards #382 

Currently, there is only a single button in GUI which loads `.vert` and `.frag` files - which would be used by the user to upload glsl function definitions.

These loaded function definitions would then be appended to the shader source code, which already have the function signatures. The function signature of `bsdf_cos_sample` should be present in the vertex shader because it outputs positions, but it is currently in the fragment shader.

I've also encountered a bug where, as soon as I take control of the GUI cursor, the yellow quad on the screen, which was visible earlier, disappears.

It seems like there is a `*.layout` entry in the `.gitignore` file which prevents the XML layout (which also has the extension `.layout`) from being tracked.


devshgraphicsprogramming:
Will review tomorrow

devshgraphicsprogramming:
> It seems like there is a *.layout entry in the .gitignore file which prevents the XML layout (which also has the extension .layout) from being tracked.

this is a remnant of a time before cmake, we used to have codeblocks project files in the repo. Now we can allow `.layout` to be tracked.

examples_tests/28.BSDFValidator/BSDFValidatorApp.h
```diff
@@ -0,0 +1,67 @@
...
+
+private:
+    void EventFunctionDefinitionBrowse(const CEGUI::EventArgs& e);
+    
+    void LoadDefinitions(const std::string& path);
+
+    static constexpr const char* s_FunctionDefinitionFileDialogTitle = "Select Function Definitions";
+    const std::vector<std::string> m_FunctionDefinitionFileDialogFilters =
+    {
+            "Shaders (*.frag *.vert)",
```
devshgraphicsprogramming: you only want to be loading `.glsl` files

examples_tests/28.BSDFValidator/Main.cpp
```diff
@@ -0,0 +1,124 @@
...
+    uint32_t offsets[2] = { video::StreamingTransientDataBufferMT<>::invalid_address,video::StreamingTransientDataBufferMT<>::invalid_address };
+    uint32_t alignments[2] = { sizeof(decltype(vertices[0u])),sizeof(decltype(indices[0u])) };
+    uint32_t sizes[2] = { sizeof(vertices),sizeof(indices) };
+    upStreamBuff->multi_place(2u, (const void* const*)dataToPlace, offsets, sizes, alignments);
+
+    if (upStreamBuff->needsManualFlushOrInvalidate())
+    {
+        auto upStreamMem = upStreamBuff->getBuffer()->getBoundMemory();
+        driver->flushMappedMemoryRanges({ video::IDriverMemoryAllocation::MappedMemoryRange(upStreamMem, offsets[0],sizes[0]),video::IDriverMemoryAllocation::MappedMemoryRange(upStreamMem,offsets[1],sizes[1]) });
+    }
```
devshgraphicsprogramming: try rendering an attribute-less mesh (like in Vulkan examples)

move the `vertices` array to the vertex shader and index it with `gl_VertexID`

All this code is overkill for splatting points accoridng to a BSDF

examples_tests/28.BSDFValidator/Main.cpp
```diff
@@ -0,0 +1,124 @@
...
+            auto buff = core::smart_refctd_ptr<video::IGPUBuffer>(upStreamBuff->getBuffer());
+
+            // Position attribute.
+            desc->setVertexAttrBuffer(core::smart_refctd_ptr(buff), asset::EVAI_ATTR0, asset::EF_R32G32B32_SFLOAT, sizeof(VertexStruct), 0);
+
+            desc->setIndexBuffer(std::move(buff));
+        }
+
+        meshBuffer->setIndexBufferOffset(offsets[1]);
+        meshBuffer->setIndexType(asset::EIT_16BIT);
```
devshgraphicsprogramming: if you set `EIT_UNKNOWN` then you can create a mesh without index buffer, Index-Count becomes vertex count

examples_tests/28.BSDFValidator/Shader.frag
```diff
@@ -0,0 +1,8 @@
+#version 330 core
+
+out vec4 pixelColor;
```
devshgraphicsprogramming: you need `layout(location = 0)` for this

examples_tests/28.BSDFValidator/Shader.frag
```diff
@@ -0,0 +1,8 @@
+#version 330 core
```
devshgraphicsprogramming: we use `#version 430 core`

examples_tests/28.BSDFValidator/Main.cpp
achalpandeyy: As mentioned in the comment above, this was just to test if the custom shaders are working as intended. It will not make it to the final version.

examples_tests/28.BSDFValidator/Main.cpp
achalpandeyy: Same as above.

achalpandeyy:
Thanks for the review.

Do you have a specific GUI in mind?

devshgraphicsprogramming:
> Do you have a specific GUI in mind?

512x512 window.
Drop down list of BSDFs to use:
- Diffuse
- Custom (Load File)

Diffuse BSDF and sample function shouldn't be too hard for you to put together.
When "Custom" gets chosen, then file open dialog pops up asking for a `.glsl` file

achalpandeyy:
How about instead of a drop down let's just default to Diffuse BSDF, then whenever the user presses the `Load Function Definitions` button they'll be prompted to open a `.glsl` file?

If going back to Diffuse BSDF is important we can add a `Reset` button as well.

I feel like a drop down will introduce unnecessary complexity.

devshgraphicsprogramming:
> I feel like a drop down will introduce unnecessary complexity.

Fine.

achalpandeyy:
Function signatures in the Fragment Shader maybe wrong, it is just to build without errors. They'll be updated accordingly.

Next is the reset button.

examples_tests/28.BSDFValidator/BSDFValidatorApp.cpp
```diff
@@ -50,34 +53,125 @@ BSDFValidatorApp::BSDFValidatorApp(IrrlichtDevice& device)
...
+
+    m_VideoDriver->setTransform(video::E4X3TS_WORLD, core::matrix4x3());
+
+    upStreamBuff->multi_free(2u, (uint32_t*)&offsets, (uint32_t*)&sizes, m_VideoDriver->placeFence());
+
+}
+
+BSDFValidatorApp::~BSDFValidatorApp()
+{
+    delete m_ShaderManager;
```
devshgraphicsprogramming: you could use a `smart_refctd_ptr` you know, if only `irr::ShaderManager` inherited from `irr::core::ReferenceCounted`

examples_tests/28.BSDFValidator/BSDFValidatorApp.cpp
```diff
@@ -50,34 +53,125 @@ BSDFValidatorApp::BSDFValidatorApp(IrrlichtDevice& device)
...
+}
+
 void BSDFValidatorApp::EventFunctionDefinitionBrowse(const CEGUI::EventArgs& e)
 {
     const auto p = m_GUI->openFileDialog(s_FunctionDefinitionFileDialogTitle, m_FunctionDefinitionFileDialogFilters);
     if (p.first)
     {
-        LoadDefinitions(p.second);
+        std::string functionDefinitions = LoadDefinitions(p.second);
+        if (!functionDefinitions.empty())
```
devshgraphicsprogramming, line 146: you might want to check for shader compile fail.

examples_tests/28.BSDFValidator/ShaderManager.cpp
```diff
@@ -0,0 +1,60 @@
...
+
+#include "ShaderManager.h"
+
+#include "SMaterial.h"
+#include "IGPUProgrammingServices.h"
+
+#include <iostream>
+#include <string>
+
+namespace irr
```
devshgraphicsprogramming: do not put random example class in my engine's namespaces!

examples_tests/28.BSDFValidator/ShaderManager.h
```diff
@@ -0,0 +1,80 @@
...
+#define _IRR_SHADER_MANAGER_INCLUDED_
+
+#include "SMaterial.h"
+
+#include <string>
+
+// Forward Declarations
+namespace irr { namespace video { class IGPUProgrammingServices; } }
+
+namespace irr
```
devshgraphicsprogramming: do not put random example class in my engine's namespaces!

examples_tests/28.BSDFValidator/ShaderManager.h
```diff
@@ -0,0 +1,80 @@
...
+            gl_Position = vec4(vPosition, 1.0);
+        }
+    )";
+    static constexpr const char* FRAGMENT_SHADER_SOURCE =
+        R"(#version 430 core
+
+        layout (location = 0) out vec4 pixelColor;
+
+        vec3 bsdf_cos_eval();
+        vec3 bsdf_cos_sample();
```
devshgraphicsprogramming: I'm thinking that maybe bsdf_cos_sample should return both the chosen direction and probability (might be less intensive to compute them together) and abandon the `bsdf_cos_sample_probability` function

devshgraphicsprogramming:
I need to spec out the signatures of the functions for the bsdf for you

examples_tests/28.BSDFValidator/ShaderManager.cpp
achalpandeyy: Yeah, this was something I was kind of confused about.
Why do we put `BSDFValidatorApp` in `irr` namespace, then? I observed similar behavior in the BRDF Explorer.

examples_tests/28.BSDFValidator/BSDFValidatorApp.cpp
achalpandeyy: Isn't it already done deep within `COpenGLDriver::addHighLevelShaderMaterial` somewhere in `source/Irrlicht/COpenGLSLMaterialRenderer.cpp` ?

I was getting compile errors when I appended illegal function definitions to the shader. Is it worth doing it again?

examples_tests/28.BSDFValidator/ShaderManager.cpp
devshgraphicsprogramming: because I (unfortunately) wansn't babysitting the person who wrote it.

examples_tests/28.BSDFValidator/BSDFValidatorApp.cpp
devshgraphicsprogramming: if `COpenGLDriver::addHighLevelShaderMaterial` returns <0 then the shader failed to build.

examples_tests/28.BSDFValidator/BSDFValidatorApp.cpp
```diff
@@ -26,25 +26,22 @@ SOFTWARE.
...
 
-#include "ShaderManager.h"
-#include "IrrlichtDevice.h"
 #include "../../ext/CEGUI/ExtCEGUI.h"
+#include "irrlicht.h"
 
 #include <iostream>
 #include <memory>
 
-namespace irr
```
devshgraphicsprogramming: you can still do `using namespace irr;` in C++ files

devshgraphicsprogramming: and `using namespace irr::ext::cegui;`

examples_tests/28.BSDFValidator/Main.cpp
```diff
@@ -18,7 +18,7 @@ int main()
     params.Bits = 24; //may have to set to 32bit for some platforms
     params.ZBufferBits = 24; //we'd like 32bit here
     params.DriverType = video::EDT_OPENGL; //! Only Well functioning driver, software renderer left for sake of 2D image drawing
-    params.WindowSize = core::dimension2d<uint32_t>(1280, 720);
+    params.WindowSize = core::dimension2d<uint32_t>(512, 512);
```
devshgraphicsprogramming, line 21: you will probably want to display the buttons/UI not over the texture, so you need a larger window

achalpandeyy:
Are the function signatures finalized?

devshgraphicsprogramming:
On it

devshgraphicsprogramming:
You can use this as a draft
```glsl
// do not use this struct in SSBO or UBO, its wasteful on memory
struct BSDFBasicParams
{
   vec3 L;
   vec3 N;
   // optional can have undef value (specular only)
   vec3 V;
   // for differentials
   vec3 dPosdx;
   vec3 dPosdy;
};

// do not use this struct in SSBO or UBO, its wasteful on memory
struct BSDFSurfaceParams
{
   BSDFBasicParams basic;
   float NdotL;
   float NdotL_squared;
   float NdotV;
   float NdotV_squared;
   float NdotH;
   float VdotL; // same as LdotV
   float VdotH; // same as LdotH
   float LplusV_rcpLen; // left over for BSDF that want to implement bump mapping
};

// do not use this struct in SSBO or UBO, its wasteful on memory
struct BSDFAnisotropicParams
{
   BSDFIsotropicParams isotropic;
   vec3 T;
   float TdotH;
   vec3 B;
   float BdotH;
   vec3 TdotL;
   vec3 BdotL;
   vec3 TdotV;
   vec3 BdotV;
};
```

devshgraphicsprogramming:
BSDF function signatures to be used in your program
```glsl
// Spectrum can be exchanged to a float for monochrome
#define Spectrum vec3

//! This is the function that evaluates the BSDF for specific view and observer direction
Spectrum bsdf_cos_eval(in BSDFIsotropicParams);

//! generates a incoming light sample position given a random number in [0,1]^2 + returns a probability of the sample being generated (ideally distributed exactly according to bsdf_cos_eval)
BSDFSample bsdf_cos_gen_sample(in vec2 _sample);

//! returns bsdf_cos_eval/Probability_of(bsdf_cos_gen_sample) but without numerical issues
Spectrum bsdf_cos_sample_eval(out DirAndDifferential L, in ViewSurfaceInteraction V, in vec2 _sample)
```

devshgraphicsprogramming:
Ok @achalpandeyy I stopped editing around the BSDF struct definitions

achalpandeyy:
The user is going to fill in the struct details in the loaded function definitions, right?
So that means, things like `VdotL`, `NdotL` etc are not necessary, because different BSDFs will have different formulae and they might not require `VdotL` (for example) specifically for their evaluation.

achalpandeyy:
There is no `BSDFSample` struct anymore !

devshgraphicsprogramming:
> The user is going to fill in the struct details in the loaded function definitions, right?

No, but I'm making those functions for you.

>There is no BSDFSample struct anymore !

Its there
```glsl
// do not use this struct in SSBO or UBO, its wasteful on memory
struct BSDFSample
{
   DirAndDifferential L;  // incoming direction, normalized
   float probability; // for a single sample (don't care about number drawn)
};
```

> So that means, things like VdotL, NdotL etc are not necessary, because different BSDFs will have different formulae and they might not require VdotL (for example) specifically for their evaluation.

Then the compiler will optimize them away ;)

achalpandeyy:
> No, but I'm making those functions for you.

So, they are not meant to be included in the fragment shader source code?
I mean the structs.

devshgraphicsprogramming:
> So, they are not meant to be included in the fragment shader source code?

Not the user's code, you will prepend all that.

achalpandeyy:
Right.
So, the user is expected to write function definitions in terms of these structs?

devshgraphicsprogramming:
Yep.

devshgraphicsprogramming:
@achalpandeyy getter functions 
https://github.com/buildaworldnet/IrrlichtBAW/issues/384#issuecomment-546903906

You can either:

### A
Map each pixel in a cubemap to an incoming (light) direction.
Then produce a number of different cubemaps (cubemap array) for varying outgoing (view) direction.

### B
Assign one axis of a 2D image to the angle between the view and normal (`NdotV`) the other to the light and normal (`NdotL`).
Then produce a number of different textures (2D texture array) for varying `VdotL` values.

I believe Disney calls it BRDF slices, (you'd just have to extend the axes from [0,1] to [-1,1] for the NdotB and NdotL)

For anisotropic BSDF you can map your texture array layers to varying `VdotL` AND `TdotV`.
It will take you a bit of trigonometry to derive `TdotL` from the 4 other angles.

### Notes

If you have anisotropic BSDF then you need to come up with a way to turn a [0,array_count] sequence into different samples of 2D values.

devshgraphicsprogramming:
The official BSDF signatures are now here
https://github.com/buildaworldnet/IrrlichtBAW/issues/384

achalpandeyy:
Unfortunately, I ran out of time to work on this. I will not be able to make any significant progress for at least a month. If its urgent, please feel free to take it from here.

devshgraphicsprogramming:
> Unfortunately, I ran out of time to work on this. I will not be able to make any significant progress for at least a month. If its urgent, please feel free to take it from here.

No problem.

devshgraphicsprogramming:
You should move this work onto the `shader_pipeline` branch from my repository, since you're like 10-15% done, it would make sense to use the new API.

PR #398, opened 2019-11-01
Make sure flipping work, debug and fix CX and CPLY loaders

AnastaZIuk:
`COBJ` and `CSTL` loaders have worked good since now with new flipping dependency, but

`CPLY` crashes, but I have made it doesn't crash and now renders primitives
Even though Chair.ply I have downloaded doesn't appear on screen, it's weird because it seems to me everything needed is already set
`CX` crashes because of dereferencing desc member that is nullptr, it will be fixed soon

I have taken last working commits from your master and added my flipiping one more time


src/irr/asset/CXMeshFileLoader.h
```diff
@@ -193,11 +195,11 @@ class CXMeshFileLoader : public asset::IAssetLoader
 
 	bool parseDataObjectMaterial(SContext& _ctx, video::SCPUMaterial& material);
 
-	bool parseDataObjectAnimationSet(SContext& _ctx);
+	bool parseDataObjectAnimationSet(SContext& _ctx, orientationFunction performActionBasedOnOrientationSystem);
```
devshgraphicsprogramming: use a reference, the functional object is very expensive to pass around by copy

In all honesty I'd prefer a bunch of nice staateless static methods to a lambda

AnastaZIuk: yup, it will be fixed

devshgraphicsprogramming:
Status according to conversation with @AnastaZIuk :
- Memory Leak in PLY or some sort of corruption leading to crash
- BAW loader does not respect the flag yet but works.

src/irr/asset/CXMeshFileLoader.h
AnastaZIuk: @devshgraphicsprogramming  well, how should it look like?
do you mean a static method expecting two lambdas for handling orientation?

AnastaZIuk:
> Status according to conversation with @AnastaZIuk :
> 
> * Memory Leak in PLY or some sort of corruption leading to crash
> * BAW loader does not respect the flag yet but works.

There is BAW Loader left

src/irr/asset/CXMeshFileLoader.h
AnastaZIuk: @devshgraphicsprogramming  I mean, if there should be such a method, it should have access to `_params`. Should I create a pointer to it as a member? How would you like it to look like?

AnastaZIuk: @devshgraphicsprogramming how should I apply flipping on BAW Loader exactly?

devshgraphicsprogramming: are there still questions?

AnastaZIuk:
@devshgraphicsprogramming  I think it can be **merged**
everything has been tested in significant detail

the plan is to do BAW later, after writer flags
here are loaders that work great with flipping dependecy

src/irr/asset/CXMeshFileLoader.h
AnastaZIuk: it's been valid still
> how should I apply flipping on BAW Loader exactly?

PR #399, opened 2019-11-07
Get rid of a fuckton of warnings about loosing precision

AnastaZIuk:
The goal is to get rid of warnings and provide appropriate clamping.

### DIFF
--- a/include/irr/asset/format/decodePixels.h
+++ b/include/irr/asset/format/decodePixels.h
@@ -629,7 +629,7 @@ namespace irr { namespace video // why is this in the video namespace!?
     inline void decodePixels<asset::EF_R16_USCALED, double>(const void* _pix[4], double* _output, uint32_t _blockX, uint32_t _blockY)
     {
         const uint16_t& pix = reinterpret_cast<const uint16_t*>(_pix[0])[0];
-        _output[0] = static_cast<double>(((pix >> 0) & 0xffffULL));
+		_output[0] = ((pix >> 0) & 0xffffULL);
     }
 
     template<>
@@ -1035,7 +1035,7 @@ namespace irr { namespace video // why is this in the video namespace!?
         {
             const uint64_t& pix = reinterpret_cast<const uint64_t*>(_pix)[0];
             for (uint32_t i = 0u; i < chCnt; ++i)
-                _output[i] = core::Float16Compressor::decompress(static_cast<uint16_t>(pix >> i*16));
+				_output[i] = core::Float16Compressor::decompress(pix >> i * 16);
         }
     }
 	    template<>
@@ -1157,20 +1157,20 @@ namespace irr { namespace video // why is this in the video namespace!?
                     struct { uint64_t r, g, b, a; };
                     uint64_t c[4];
                 };
-			} p[4]{};
+			} p[4];
 
             uint16_t r0, g0, b0, r1, g1, b1;
 
             const void* input = &col.c0;
             decodePixels<asset::EF_B5G6R5_UNORM_PACK16, uint64_t>(&input, p[0].c, 0u, 0u);
-            r0 = static_cast<uint16_t>(p[0].r);
-            g0 = static_cast<uint16_t>(p[0].g);
-            b0 = static_cast<uint16_t>(p[0].b);
+			r0 = p[0].r;
+			g0 = p[0].g;
+			b0 = p[0].b;
             input = &col.c1;
             decodePixels<asset::EF_B5G6R5_UNORM_PACK16, uint64_t>(&input, p[1].c, 0u, 0u);
-            r1 = static_cast<uint16_t>(p[1].r);
-            g1 = static_cast<uint16_t>(p[1].g);
-            b1 = static_cast<uint16_t>(p[1].b);
+			r1 = p[1].r;
+			g1 = p[1].g;
+			b1 = p[1].b;
             if (col.c0 > col.c1)
             {
                 p[2].r = (2 * r0 + 1 * r1) / 3;
@@ -1197,7 +1197,7 @@ namespace irr { namespace video // why is this in the video namespace!?
             const uint32_t idx = 4u*_y + _x;
             const uint32_t cw = 3u & (col.lut >> (2u * idx));
             for (uint32_t i = 0u; i < (_alpha ? 4u : 3u); ++i)
-                _output[i] = static_cast<T>(p[cw].c[i]);
+				_output[i] = p[cw].c[i];
         }
         template<typename T>
         inline void decodeBC2(const void* _pix, T* _output, uint32_t _x, uint32_t _y)
@@ -1297,7 +1297,7 @@ namespace irr { namespace video // why is this in the video namespace!?
             SRGB2lin<double>(s);
             T* lin = _srgb;
             for (uint32_t i = 0; i < 3u; ++i)
-                lin[i] = static_cast<T>(core::clamp(s[i] * 255., 0.4, 255.4));
+				lin[i] = s[i] * 255.;
         }
 
         template<typename T>
@@ -1307,7 +1307,7 @@ namespace irr { namespace video // why is this in the video namespace!?
             lin2SRGB<double>(s);
             T* srgb = _lin;
             for (uint32_t i = 0; i < 3u; ++i)
-                srgb[i] = static_cast<T>(core::clamp(s[i] * 255., 0.4, 255.4));
+				srgb[i] = s[i] * 255.;
         }
     }
 
--- a/include/irr/asset/format/decodePixels.h
+++ b/include/irr/asset/format/decodePixels.h
@@ -629,7 +629,7 @@ namespace irr { namespace video // why is this in the video namespace!?
     inline void decodePixels<asset::EF_R16_USCALED, double>(const void* _pix[4], double* _output, uint32_t _blockX, uint32_t _blockY)
     {
         const uint16_t& pix = reinterpret_cast<const uint16_t*>(_pix[0])[0];
-        _output[0] = ((pix >> 0) & 0xffffULL);
+        _output[0] = static_cast<double>(((pix >> 0) & 0xffffULL));
     }
 
     template<>
@@ -1035,7 +1035,7 @@ namespace irr { namespace video // why is this in the video namespace!?
         {
             const uint64_t& pix = reinterpret_cast<const uint64_t*>(_pix)[0];
             for (uint32_t i = 0u; i < chCnt; ++i)
-                _output[i] = core::Float16Compressor::decompress(pix >> i*16);
+                _output[i] = core::Float16Compressor::decompress(static_cast<uint16_t>(pix >> i*16));
         }
     }
 	    template<>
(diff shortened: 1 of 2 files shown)


include/irr/asset/format/encodePixels.h
```diff
@@ -499,87 +499,87 @@ namespace irr { namespace video
     {
         int8_t* pix = reinterpret_cast<int8_t*>(_pix);
         for (uint32_t i = 0u; i < 3u; ++i)
-            pix[i] = _input[i] * 127.;
+            pix[i] = static_cast<int8_t>(_input[i] * 127.);
```
devshgraphicsprogramming: this bad-boy needs a `core::clamp`

include/irr/asset/format/encodePixels.h
```diff
@@ -499,87 +499,87 @@ namespace irr { namespace video
...
     }
 	
     template<>
     inline void encodePixels<asset::EF_R8G8B8_USCALED, double>(void* _pix, const double* _input)
     {
         uint8_t* pix = reinterpret_cast<uint8_t*>(_pix);
         for (uint32_t i = 0u; i < 3u; ++i)
-            pix[i] = _input[i];
+            pix[i] = static_cast<int8_t>(_input[i]);
```
devshgraphicsprogramming: this bad-boy needs a `core::clamp`

include/irr/asset/format/encodePixels.h
```diff
@@ -499,87 +499,87 @@ namespace irr { namespace video
...
     }
 	
     template<>
     inline void encodePixels<asset::EF_R8G8B8_SSCALED, double>(void* _pix, const double* _input)
     {
         int8_t* pix = reinterpret_cast<int8_t*>(_pix);
         for (uint32_t i = 0u; i < 3u; ++i)
-            pix[i] = _input[i];
+            pix[i] = static_cast<int8_t>(_input[i]);
```
devshgraphicsprogramming: this bad-boy needs a `core::clamp`

devshgraphicsprogramming:
generally speaking most of the encodes need a `core::clamp`

include/irr/asset/format/encodePixels.h
AnastaZIuk: Actuallly I don't get it technically. I mean, won't there be a maximum value of `int8_t` if I casted `double` to `int8_t` assuming my double variable is around 100k? Won't it be clamped automaticaly even though?

Crisspl: Nah, it's UB. See [c++14 standard](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2014/n4296.pdf), paragraph 4.9.1.
However on gcc indeed there is a clamp http://coliru.stacked-crooked.com/a/9f6f5ec4b0ded22c
But ofc you cannot rely on behavior of single compiler since standard says that anything really can happen (just as well it could be wrap-around did in some tricky way or some random shit)

AnastaZIuk: > Nah, it's UB. See [c++14 standard](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2014/n4296.pdf), paragraph 4.9.1.
> However on gcc indeed there is a clamp http://coliru.stacked-crooked.com/a/9f6f5ec4b0ded22c
> But ofc you cannot rely on behavior of single compiler since standard says that anything really can happen (just as well it could be wrap-around did in some tricky way or some random shit)

Thank you so much, I'm gonna check it!

include/irr/asset/format/encodePixels.h
```diff
@@ -499,23 +499,23 @@ namespace irr { namespace video
...
+            pix[i] = static_cast<int8_t>(core::clamp(_input[i], 0.4, 255.4));
     }
 	
     template<>
     inline void encodePixels<asset::EF_R8G8B8_SSCALED, double>(void* _pix, const double* _input)
     {
         int8_t* pix = reinterpret_cast<int8_t*>(_pix);
         for (uint32_t i = 0u; i < 3u; ++i)
-            pix[i] = static_cast<int8_t>(_input[i]);
+            pix[i] = static_cast<int8_t>(core::clamp(_input[i], 0.4, 255.4));
```
devshgraphicsprogramming: wrong range... maybe make a template intermediate function calling `core::clamp` that will deduce the limits from integer type?

include/irr/asset/format/decodePixels.h
```diff
@@ -1297,7 +1297,7 @@ namespace irr { namespace video // why is this in the video namespace!?
             SRGB2lin<double>(s);
             T* lin = _srgb;
             for (uint32_t i = 0; i < 3u; ++i)
-                lin[i] = static_cast<T>(s[i] * 255.);
+                lin[i] = static_cast<T>(core::clamp(s[i] * 255., 0.4, 255.4));
```
devshgraphicsprogramming: don't touch these functions in `decode`

devshgraphicsprogramming:
> 
> 
> > 0.4 is an epsilon for clamping, isn't it good actually?
> 
> i guess it should be like
> 
> ```
> core::clamp(_input[i]+singf64(_input[i])*0.4, -32768.0, 32767.0)
> ```
> 
> where signf64() is some function returning -1.0 or 1.0 depending on sign of input
> 

Actually its the clamp that should have the epsillon, its because lots of NPoT numbers are not exactly representable in IEE754, and floats round down.

Should be on the clamp, because it affects the clamp (make sure that i.e. 2^32-1 doesn't truncate lower)

But you are correct that floats truncate towards zero, so the clamp minimums less than 0.0 should have -0.4 added not 0.4 

> ```
> double signf64(double _x)
> {
>     constexpr double retval[2] { 1.0, -1.0 }; 
>     const uint64_t& x = reinterprest_cast<uint64_t&>(_x);
>     return retval[(x>>63)&1ull];
> }
> ```
> 
> > yeah, but there won't be a symmetry, so shader could get broken if you did it as you have written
> > there would not be a 0.0 in shader, but because of shift there would be arround +/- 1/128
> 
> i think u're talking about normalized formats where indeed AFAIK -128 should map to -1 (example for 8bit snorm), but [s|u]scaled formats are seen in shaders as just "static_cast(intval)"

Yeah "SCALED" formats need a larger minimum in the clamp.
"NORM" stays as is.

AnastaZIuk:
@devshgraphicsprogramming 
> Yeah "SCALED" formats need a larger minimum in the clamp.
> "NORM" stays as is.

https://github.com/buildaworldnet/IrrlichtBAW/pull/399/files#diff-36027e8ca0aae34f9c02456484835698R41
So NORM shouldn't have an epsilon at minimum? I mean, If there is for instance uint8_t NORM, min = 0, max = 255 + epsilon

devshgraphicsprogramming:
@AnastaZIuk NORM and SCALED clamp to the same values (because they are stored in the same integers)

The only difference is that the `double` value gets multiplied by the Max Positive Integer Value before the clamp for NORM and there is no mulitplication for SCALED before the clamp

include/irr/asset/format/encodePixels.h
```diff
@@ -7,9 +7,43 @@
 #include "irr/core/core.h"
 #include "irr/asset/format/EFormat.h"
 
-
 namespace irr { namespace video
 {	
+
+	template<typename T>
+	inline uint64_t getRangeValueOfVariable(bool maxValue = true)
```
devshgraphicsprogramming: don't we have this function already in terms of `E_FORMAT` @Crisspl ?

include/irr/asset/format/encodePixels.h
```diff
@@ -7,9 +7,43 @@
...
+		else if (std::is_same<T, int16_t>::value)
+			return 32767ull * (maxValue ? 1 : -1);
+		else if (std::is_same<T, int32_t>::value)
+			return 2147483647ull * (maxValue ? 1 : -1);
+		else
+			return -1; // handle an error
+	}
+
+	template<typename T, typename U>
+	inline void clampVariableProperly(T& variableToAssignClampingTo, const U& variableToClamp, bool isItNormalizedVariable = true)
```
devshgraphicsprogramming: input is always a `double`

you can take away the boolean argument 
we have a function that is a template of `E_FORMAT` I believe its called `isFormatNormalized<E_FORMAT format>()` or something like that.

include/irr/asset/format/encodePixels.h
```diff
@@ -7,9 +7,43 @@
...
+	{
+		const auto& max = getRangeValueOfVariable<T>(true);
+		const auto& min = getRangeValueOfVariable<T>(false);
+		constexpr float epsilon = 0.4f;
+		const auto& epsilonToAddToMin = (min < 0 ? -epsilon : epsilon);
+
+		if (isItNormalizedVariable)                                              // should NORM has epsilon?
+			variableToAssignClampingTo = static_cast<T>(core::clamp(variableToClamp, min + epsilonToAddToMin, max + epsilon));
+		else
+			variableToAssignClampingTo = static_cast<T>(core::clamp(variableToClamp * static_cast<U>(max), min + epsilonToAddToMin, max + epsilon));
```
devshgraphicsprogramming: its actually the other way round... the NORMALIZED gets a multiply up, the SCALED stays as is

Reason is that normalized store the floating point [0,1] or [-1,1] range in all the ints, SCALED just casts a float to an int.

AnastaZIuk:
I'm done If there is no function @devshgraphicsprogramming mentioned here https://github.com/buildaworldnet/IrrlichtBAW/pull/399#discussion_r347090520 about

include/irr/asset/format/encodePixels.h
```diff
@@ -30,18 +30,18 @@ namespace irr { namespace video
 			return -1; // handle an error
 	}
 
-	template<typename T, typename U>
-	inline void clampVariableProperly(T& variableToAssignClampingTo, const U& variableToClamp, bool isItNormalizedVariable = true)
+	template<typename T>
+	inline void clampVariableProperly(T& variableToAssignClampingTo, const double& variableToClamp, const asset::E_FORMAT &format)
```
devshgraphicsprogramming: make format a template parameter.

include/irr/asset/format/encodePixels.h
```diff
@@ -30,18 +30,18 @@ namespace irr { namespace video
...
 	{
 		const auto& max = getRangeValueOfVariable<T>(true);
```
devshgraphicsprogramming: shouldn't this all be constexpr + constexpr function instead of reference?

include/irr/asset/format/encodePixels.h
devshgraphicsprogramming: ok @AnastaZIuk answer accepted.

include/irr/asset/format/encodePixels.h
AnastaZIuk: Is it a good practice or it goes on performance or you want it to be as you said just simply?

include/irr/asset/format/encodePixels.h
AnastaZIuk: to fix, sure

include/irr/asset/format/encodePixels.h
devshgraphicsprogramming: template parameter basically is guaranteed inlining and duplicate code generation for every value of `format`, a parameter the compiler may still leave some code to deal with a mutable `format`

include/irr/asset/format/encodePixels.h
AnastaZIuk: so is there such?

devshgraphicsprogramming: nope not for this.

include/irr/asset/format/encodePixels.h
```diff
@@ -30,13 +30,13 @@ namespace irr { namespace video
...
 	}
 
-	template<typename T>
-	inline void clampVariableProperly(T& variableToAssignClampingTo, const double& variableToClamp, const asset::E_FORMAT &format)
+	template<asset::E_FORMAT format, typename T>
+	inline void clampVariableProperly(T& variableToAssignClampingTo, const double& variableToClamp)
 	{
-		const auto& max = getRangeValueOfVariable<T>(true);
-		const auto& min = getRangeValueOfVariable<T>(false);
+		constexpr uint64_t& max = getRangeValueOfVariable<T>(true);
```
devshgraphicsprogramming: you don't need `&` on a constexpr, thats actually an error to do so

include/irr/asset/format/encodePixels.h
devshgraphicsprogramming: make it `constexpr inline uint64_t`

AnastaZIuk: sure, to fix

devshgraphicsprogramming:
does it build?

AnastaZIuk:
@devshgraphicsprogramming  section with encode builds, status: to merge

include/irr/asset/format/encodePixels.h
```diff
@@ -7,9 +7,43 @@
...
+			return 	4294967295ull * (maxValue ? 1 : 0);
+
+		else if (std::is_same<T, int8_t>::value)
+			return 127ull * (maxValue ? 1 : -1);
+		else if (std::is_same<T, int16_t>::value)
+			return 32767ull * (maxValue ? 1 : -1);
+		else if (std::is_same<T, int32_t>::value)
+			return 2147483647ull * (maxValue ? 1 : -1);
+		else
+			return -1; // handle an error
```
devshgraphicsprogramming, line 30: you could use a `static_assert` plus `!core::is_any_of` in a future PR

PR #400, opened 2019-11-12
Compute Pipelines and Push Constants

devshgraphicsprogramming:


devshgraphicsprogramming:
@Crisspl are you using an older revision of glew (or one not from @achalpandeyy ) in your branch?

devshgraphicsprogramming:
> 
> 
> this applies to all the `clear` functions
> 
> Seems we need to split `GSB_PIPELINE_AND_RASTER_PARAMETERS` back into two, there's no reason for the shader to be flushed and other stuff just because we're clearing an FBO buffer XD

@Crisspl?

src/irr/asset/CImageLoaderPNG.cpp
```diff
@@ -236,18 +239,32 @@ asset::SAssetBundle CImageLoaderPng::loadAsset(io::IReadFile* _file, const asset
...
 
+    auto texelBuffer = core::make_smart_refctd_ptr<ICPUBuffer>(image->getImageDataSizeInBytes());
+    auto regions = core::make_refctd_dynamic_array<core::smart_refctd_dynamic_array<ICPUImage::SBufferCopy>>(1u);
+    ICPUImage::SBufferCopy& region = regions->front();
+    //region.imageSubresource.aspectMask = ...; //waits for Vulkan
+    region.imageSubresource.mipLevel = 0u;
+    region.imageSubresource.baseArrayLayer = 0u;
+    region.imageSubresource.layerCount = 1u;
+    region.bufferOffset = 0u;
+    region.bufferRowLength = 0u;//??
```
devshgraphicsprogramming: pitch, i think in bytes

src/irr/asset/CImageLoaderPNG.cpp
```diff
@@ -236,18 +239,32 @@ asset::SAssetBundle CImageLoaderPng::loadAsset(io::IReadFile* _file, const asset
...
+    region.bufferImageHeight = 0u; //??
```
devshgraphicsprogramming: see how address is calculated in VK, if this is 0 then its assumed to be tightly packed.

src/irr/asset/CImageLoaderPNG.cpp
devshgraphicsprogramming: actually not in bytes.

But you should aim for it to be image width rounded up to the next "lowest common multiple of 8 and texel block byte size" after "texel block byte size times image width"

BAsically
```cpp
uint32_t findNextAlignedBlock(uint32_t width, uint32_t blockByteSize, uint32_t _alignment=8u)
{
   for (uint32_t addr=width*blockByteSize, n=width; addr<(~0u)-blockSize+1u; width+=blockByteSize,n++)
   {
      if (addr&(_alignment-1u))
         continue;
      return n;
   }
   return 0u;
}
```

src/irr/asset/CImageLoaderPNG.cpp
```diff
@@ -236,18 +239,32 @@ asset::SAssetBundle CImageLoaderPng::loadAsset(io::IReadFile* _file, const asset
...
+    region.imageOffset = { 0u, 0u, 0u };
+    region.imageExtent = image->getCreationParameters().extent;
+
+    image->setBufferAndRegions(std::move(texelBuffer), regions);
+
 	// Fill array of pointers to rows in image data
-	const uint32_t pitch = image->getPitchIncludingAlignment();
-	uint8_t* data = reinterpret_cast<uint8_t*>(image->getData());
+	const uint32_t pitch = texelBuffer->getSize()/Height;
```
devshgraphicsprogramming: `(region.bufferRowLength*blockByteSize)` instead

src/irr/asset/CImageLoaderPNG.cpp
```diff
@@ -236,18 +239,32 @@ asset::SAssetBundle CImageLoaderPng::loadAsset(io::IReadFile* _file, const asset
...
+	uint8_t* data = reinterpret_cast<uint8_t*>(texelBuffer->getPointer());
```
devshgraphicsprogramming, line 281: you've already moved the `texelBuffer` you will crash on nullptr

devshgraphicsprogramming:
@Crisspl looks good.

include/irr/asset/IImage.h
```diff
@@ -401,11 +402,19 @@ class IImage : public IDescriptor
 		//!
 		inline core::vector3du32_SIMD getMipSize(uint32_t level=0u) const
 		{
+            core::vector3d<uint32_t> sz(params.extent.width,
+                params.extent.height,
+                params.extent.depth);
+            core::vector3d<uint32_t> one(1u, 1u, 1u);
+            auto r = core::max(sz, one);
+            return core::vector3du32_SIMD(r.X, r.Y, r.Z);
```
devshgraphicsprogramming, line 410: no this is wrong, you're not taking level into account!

include/vector3d.h
```diff
@@ -70,6 +70,11 @@ namespace core
 			return !operator==(other);
 		}
 
+        bool operator<(const vector3d<T>& other) const
```
devshgraphicsprogramming, line 73: I removed this function for a reason!

src/irr/asset/CImageLoaderPNG.cpp
```diff
@@ -260,17 +261,17 @@ asset::SAssetBundle CImageLoaderPng::loadAsset(io::IReadFile* _file, const asset
     region.imageOffset = { 0u, 0u, 0u };
     region.imageExtent = image->getCreationParameters().extent;
 
-    image->setBufferAndRegions(std::move(texelBuffer), regions);
-
 	// Fill array of pointers to rows in image data
-	const uint32_t pitch = texelBuffer->getSize()/Height;
+	const uint32_t pitch = (region.bufferRowLength*getTexelOrBlockBytesize(image->getCreationParameters().format));
```
devshgraphicsprogramming, line 280: and you `region.bufferRowLength` is still 0

devshgraphicsprogramming: actually its fine.

include/vector3d.h
devshgraphicsprogramming: removed.

Crisspl: I added it temporarily because I needed to perform core::max on this type just for when _mm_cmplt/gt are broken. See commit previous to last one, I replaced vector3simd for vector3d, somewhere in IImage I think

devshgraphicsprogramming: fixing the compile issue now.

devshgraphicsprogramming: changed to a macro in vectorSIMDf, works now

include/irr/asset/IImage.h
devshgraphicsprogramming: fixed it myself.

src/irr/asset/CImageLoaderPNG.cpp
devshgraphicsprogramming: had to change it a bit, realized this wont work

PR #401, opened 2019-11-13
Create WriterFlags and WriterFlipping

AnastaZIuk:


src/irr/asset/CSTLMeshWriter.cpp
```diff
@@ -150,6 +150,42 @@ bool CSTLMeshWriter::writeMeshBinary(io::IWriteFile* file, const asset::ICPUMesh
 	for (uint32_t i=0; i<mesh->getMeshBufferCount(); ++i)
 	{
 		asset::ICPUMeshBuffer* buffer = mesh->getMeshBuffer(i);
+
+		/*
+			pseudo code that has to be applied in binary and ASCII writing
```
AnastaZIuk: @devshgraphicsprogramming  @Crisspl  guys please help me
I have no idea how I can determine which buffer corresponds to which attribute
Is there actaully a way to this? Let's say that I would like to get normals buffer
How can I check it in our engine?

Crisspl: https://github.com/Crisspl/IrrlichtBAW/blob/master/include/irr/asset/ICPUMeshDataFormatDesc.h#L117
also see how getAttribPointer() works https://github.com/Crisspl/IrrlichtBAW/blob/master/include/irr/asset/ICPUMeshBuffer.h#L185

devshgraphicsprogramming: there's a vague convention that some loaders rely on:
- 0 : position (but there's a query for that)
- 1 : UV
- 2 : Color
- 3 : Normal

Not sure if I got 1 and 2 the right way around, see the OBJ loader, its a good reference.

Some formats like `.baw` or `.gltf` don't really give a shit which attribute does what and can save more than the pos+uv+color+normal.

include/irr/asset/IAssetWriter.h
```diff
@@ -24,15 +24,30 @@ enum E_WRITER_FLAGS : uint32_t
...
+	E_LOADER_PARAMETER_FLAGS::ELPF_NONE is default and means that there is nothing to perform.
+	E_LOADER_PARAMETER_FLAGS::ELPF_RIGHT_HANDED_MESHES specifies that a mesh will be flipped, 
+	if current flag responsible for orientation is signed as left handed. If it isn't,
+	no flipping will be performed, because there will be no need for that.
+*/
+
+enum E_WRITER_PARAMETER_FLAGS : uint64_t
+{
+	EWPF_NONE = 0,											//!< default value, it doesn't do anything
+	EWPF_RIGHT_HANDED_MESHES = 0x1,							//!< specifies that a mesh will be flipped in such a way that it'll look correctly in right-handed camera system
```
devshgraphicsprogramming: call it `EWPF_MESH_IS_RIGHT_HANDED`

It does not mean it will be flipped.

Explanation:

If a mesh format is right-handed then no flipping for right handed input mesh, but flipping for left handed input mesh (the default).
Example: Possibly OBJ (I think we flip when we load by default), definitely `.gtlf`

If a mesh format is left-handed then flip for right-handed input but no flipping for left-handed input.
Example: DirectX .X

include/irr/asset/IAssetWriter.h
```diff
@@ -42,6 +57,8 @@ class IAssetWriter : public virtual core::IReferenceCounted
         size_t encryptionKeyLen;
         const uint8_t* encryptionKey;
         const void* userData;
+		const E_WRITER_PARAMETER_FLAGS writerFlags;
+		const irr::asset::IAssetLoader::E_LOADER_PARAMETER_FLAGS loaderFlagsUsedForWriting;
```
devshgraphicsprogramming: what's `loadderFlagsUSedForWriting` doing here?

include/irr/asset/IAssetWriter.h
devshgraphicsprogramming: btw move the enum inside `IAssetWriter`

src/irr/asset/CSTLMeshWriter.cpp
devshgraphicsprogramming: also input to function is const so you can't flip in-place on the mesh

devshgraphicsprogramming:
@AnastaZIuk an overarching idea for .BAW format patching, we will store the EWPF_MESH_IS_RIGHT_HANDED flag (1 bit) in the MeshBufferBlobV2 (we need to bump the version).

Now if that flag is present in the .baw file's meshbuffer blob, then we know that mesh is right-handed, if its not then its left-handed (like it has been now).

If right-handed flag in blob does not match the right-handed loader parameter flag when loading the asset, then flip position and normal.

include/irr/asset/IAssetWriter.h
AnastaZIuk: We have to know the orientation of loaded mesh, I mean the orientation applied to mesh we want to write/save, haven't we?

devshgraphicsprogramming: Your proposed approach makes no sense, for two reasons:
1. Not every mesh to be written is loaded (some are generated, i.e. geometry creator)
2. You will use the `writerFlags` to tell the writer if the meshes being passed in are right handed or not (each writer will decide internally whether to flip or not depending on the convention for the format OR if the format can take the orientation metadata like .baw)

AnastaZIuk:
Well, the result loading, writing and loading is https://imgur.com/a/nnyaEgQ
It seems to me that winding order should be changed to solve it, so simply exchange vertex1 with vertex3, but I think it **shouldn't** be applied actually, something went wrong, didn't it?

the code responsible for loading:

```
asset::IAssetLoader::SAssetLoadParams lparams(0, nullptr, irr::asset::IAssetLoader::ECF_CACHE_EVERYTHING, nullptr, irr::asset::IAssetLoader::ELPF_NONE);
auto cpumesh = core::smart_refctd_ptr_static_cast<asset::ICPUMesh>(*am->getAsset("../../media/extrusionLogo_TEST_fixed.stl", lparams).getContents().first);
	
asset::CBAWMeshWriter::WriteProperties bawprops;
asset::IAssetWriter::SAssetWriteParams wparams(cpumesh.get(), asset::EWF_COMPRESSED, 0.f, 0, nullptr, &bawprops, irr::asset::IAssetWriter::EWPF_WRITE_RIGHT_HANDED);
am->writeAsset("extrusionLogo_TEST_fixed.stl", wparams);

cpumesh = core::smart_refctd_ptr_static_cast<asset::ICPUMesh>(*am->getAsset("extrusionLogo_TEST_fixed.stl", lparams).getContents().first);
	
auto setMaterialTypeOnAllMaterials = [](auto* node, auto newMaterialType)
{
	auto* mesh = node->getMesh();
	for (auto i = 0u; i < mesh->getMeshBufferCount(); i++)
	{
		auto& material = mesh->getMeshBuffer(i)->getMaterial();
		material.MaterialType = newMaterialType;
	}
};

if (cpumesh)
     setMaterialTypeOnAllMaterials(smgr->addMeshSceneNode(std::move(driver->getGPUObjectsFromAssets(&cpumesh.get(), (&cpumesh.get())+1)->operator[](0))),newMaterialType);
```

Loaders assume that every resource  is loaded as input in righhanded, but the default is lefthanded, so flipping is supplied if there is no righthanded flag set while loading. I tell the writer that it has to save it as righthanded and loading it should be same as loading at begging, but winding order error occours

**EDIT**
for sure changing vertex1 with vertex3 is _valid_, because CSTLLoader changes it as well due to winding order :D

**@devshgraphicsprogramming it can be merged, I'm done and everything has been tested thoroughly**

src/irr/asset/CSTLMeshWriter.cpp
AnastaZIuk: Thank you guys, I will remember those hints

include/irr/asset/IAssetWriter.h
```diff
@@ -24,30 +24,32 @@ enum E_WRITER_FLAGS : uint32_t
...
+		E_LOADER_PARAMETER_FLAGS::EWPF_MESH_IS_RIGHT_HANDED specifies that a mesh will be flipped,
+		if current E_LOADER_PARAMETER_FLAGS::EWPF_WRITE_RIGHT_HANDED responsible for orientation is signed as left handed.
+		If it isn't, no flipping will be performed, because there will be no need for that.
+	*/
+
+	enum E_WRITER_PARAMETER_FLAGS : uint64_t
+	{
+		EWPF_NONE = 0,											//!< default value, it doesn't do anything
+		EWPF_MESH_IS_RIGHT_HANDED = 0x1,						//!< specifies the incoming orientation of loaded mesh we want to write						
+		EWPF_WRITE_RIGHT_HANDED = 0x2							//!< specifies the output handness orientation. There will be no flipping applied if incoming orientation matches output orientation - in case where EWPF_MESH_IS_RIGHT_HANDED and EWPF_WRITE_RIGHT_HANDED are both attached
```
devshgraphicsprogramming: this flag shouldn't exist, you should know whether mesh should be written as right handed from the format you're writing to.

devshgraphicsprogramming:
There's also a PLY writer TODO

include/irr/asset/IAssetWriter.h
AnastaZIuk: what do you mean? Don't we want to write as left or right in dependence of user choice?

devshgraphicsprogramming: the mesh format almost always has a convention, if you write out as right handed to .X then all the model viewers in the world will see the model wrong.

AnastaZIuk: okay, so I'll throw out 'EWPF_WRITE_RIGHT_HANDED'

devshgraphicsprogramming: great

AnastaZIuk:
> There's also a PLY writer TODO

@devshgraphicsprogramming done and tested

src/irr/asset/CPLYMeshWriter.cpp
```diff
@@ -208,28 +208,30 @@ bool CPLYMeshWriter::writeAsset(io::IWriteFile* _file, const SAssetWriteParams&
...
 
 	return true;
 }
 
-void CPLYMeshWriter::writeBinary(io::IWriteFile* _file, asset::ICPUMeshBuffer* _mbuf, size_t _vtxCount, size_t _fcCount, asset::E_INDEX_TYPE _idxType, void* const _indices, bool _forceFaces, const bool _vaidToWrite[4]) const
+void CPLYMeshWriter::writeBinary(io::IWriteFile* _file, asset::ICPUMeshBuffer* _mbuf, size_t _vtxCount, size_t _fcCount, asset::E_INDEX_TYPE _idxType, void* const _indices, bool _forceFaces, const bool _vaidToWrite[4], const SAssetWriteParams& _params) const
 {
     const size_t colCpa = asset::getFormatChannelCount(_mbuf->getMeshDataAndFormat()->getAttribFormat(asset::EVAI_ATTR1));
 
+	bool flipVectors = (!(_params.writerFlags & IAssetWriter::EWPF_MESH_IS_RIGHT_HANDED)) ? true : false;
```
devshgraphicsprogramming, line 225: so PLY is right handed format?

src/irr/asset/CSTLMeshWriter.cpp
```diff
@@ -304,24 +288,8 @@ void CSTLMeshWriter::writeFaceText(io::IWriteFile* file,
...
-	else
-	{
-		if (!(_params.writerFlags & IAssetWriter::EWPF_MESH_IS_RIGHT_HANDED))
-		{
-			// do nothing
-		}
-		else
-			flipVectors();
-	}
+	if (!(_params.writerFlags & IAssetWriter::EWPF_MESH_IS_RIGHT_HANDED))
```
devshgraphicsprogramming, line 291: STL is right handed too?

src/irr/asset/CPLYMeshWriter.cpp
AnastaZIuk: as we have spoken, right? Default right, but to make it left there was a flip at beggining
btw google says that it is righthanded :D

src/irr/asset/CSTLMeshWriter.cpp
AnastaZIuk: https://github.com/buildaworldnet/IrrlichtBAW/pull/401#discussion_r347512821

src/irr/asset/CPLYMeshWriter.cpp
devshgraphicsprogramming: ok cool, nice to paste a reference in the comments.

src/irr/asset/CSTLMeshWriter.cpp
devshgraphicsprogramming: so STL AND PLY are right handed?

AnastaZIuk: they are righthanded

devshgraphicsprogramming: cool, never knew that ;)

devshgraphicsprogramming:
Builds, works and is mergable?

AnastaZIuk:
@devshgraphicsprogramming  status: to merge

PR #381, opened 2019-10-18
Add support for loading KTX Textures

achalpandeyy:
Currently only works for 2D textures.

TODO: Multi Layered Textures (Texture Arrays) & Cube Maps.

Includes a KTX texture file for testing purposes.

PR #402, opened 2019-11-28
CGLILoader/Writer

AnastaZIuk:
**Status TODO:**

- handle data pointers and image's settings in loader
- handle saving -> CGLIWriter
- handle format conversion properly


source/Irrlicht/CMakeLists.txt
```diff
@@ -127,6 +127,7 @@ set(IRRLICHT_SRCS_COMMON
 	${IRR_ROOT_PATH}/src/irr/asset/CImageLoaderJPG.cpp
 	${IRR_ROOT_PATH}/src/irr/asset/CImageLoaderPNG.cpp
 	${IRR_ROOT_PATH}/src/irr/asset/CImageLoaderTGA.cpp
+	${IRR_ROOT_PATH}/src/irr/asset/CGLILoader.cpp
```
devshgraphicsprogramming, line 132: removed DDS?

src/irr/asset/CGLILoader.cpp
```diff
@@ -0,0 +1,157 @@
...
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
+*/
+
+#include "CBufferLoaderBIN.h"
```
devshgraphicsprogramming: wrong include?

src/irr/asset/CGLILoader.cpp
```diff
@@ -0,0 +1,157 @@
...
+
+			const char* filename = _file->getFileName().c_str();
+			gli::texture texture = gli::load(filename);
+			if (texture.empty())
+			{
+				os::Printer::log("Failed to load texture at path ", filename, ELL_ERROR);
+				return {};
+			}
+			
+		    gli::gl GL(gli::gl::PROFILE_GL33);
```
devshgraphicsprogramming: whats the difference between this and PROFILE_KTX

src/irr/asset/CGLILoader.cpp
```diff
@@ -0,0 +1,157 @@
...
+			glm::tvec3<GLsizei> const extent(texture.extent()); // it is on 0 level without specyfing it explicitly
+			GLsizei const totalFaces = static_cast<GLsizei>(texture.layers() * texture.faces());
+
+			ICPUImage::SCreationParams imageInfo;
+			imageInfo.format = format;
+			imageInfo.extent.width = extent.x;
+			imageInfo.extent.height = extent.y;
+			imageInfo.extent.depth = extent.z;
+			imageInfo.mipLevels = texture.levels();
+			imageInfo.arrayLayers = texture.layers();
```
devshgraphicsprogramming: total faces

src/irr/asset/CGLILoader.cpp
```diff
@@ -0,0 +1,157 @@
...
+			glTexParameteri(target, GL_TEXTURE_SWIZZLE_R, format.Swizzles[0]);
+			glTexParameteri(target, GL_TEXTURE_SWIZZLE_G, format.Swizzles[1]);
+			glTexParameteri(target, GL_TEXTURE_SWIZZLE_B, format.Swizzles[2]);
+			glTexParameteri(target, GL_TEXTURE_SWIZZLE_A, format.Swizzles[3]);*/
+			
+			glm::tvec3<GLsizei> const extent(texture.extent()); // it is on 0 level without specyfing it explicitly
+			GLsizei const totalFaces = static_cast<GLsizei>(texture.layers() * texture.faces());
+
+			ICPUImage::SCreationParams imageInfo;
+			imageInfo.format = format;
```
devshgraphicsprogramming: you will need to translate the format between gli format and our own E_FORMAT

devshgraphicsprogramming:
AFAIK the GLM submodule may be fucked up and have problems configuring/building when added as a subdirectory in CMake, have to check and if its the case figure out what's wrong.

devshgraphicsprogramming:
btw have a look at 
https://github.com/buildaworldnet/IrrlichtBAW/pull/381/files

source/Irrlicht/CMakeLists.txt
AnastaZIuk: @devshgraphicsprogramming do you want me to delete DDS (and KTX) loader?

src/irr/asset/CGLILoader.cpp
AnastaZIuk: sure, to fix

src/irr/asset/CGLILoader.cpp
AnastaZIuk: so what is a difference between layer and face? I can imagine a face as an index of texture (if it is a cubemap)

src/irr/asset/CGLILoader.cpp
AnastaZIuk: @devshgraphicsprogramming well, so the **format** struct looks like this: 
https://github.com/g-truc/gli/blob/master/gli/gl.hpp#L335

I will translate it, but I'm not sure which format member I should choose for translating
internal format: https://github.com/g-truc/gli/blob/master/gli/gl.hpp#L16
external format: https://github.com/g-truc/gli/blob/master/gli/gl.hpp#L238
or type format: https://github.com/g-truc/gli/blob/master/gli/gl.hpp#L265

src/irr/asset/CGLILoader.cpp
AnastaZIuk: I don't know actually and I can't find anything about it, so I have created an issue https://github.com/g-truc/gli/issues/172

I hope they will clarify us it

src/irr/asset/CGLILoader.cpp
devshgraphicsprogramming: that's a distinction that only GLi seems to make, in Vulkan and modern (DSA) OpenGL, faces of a cubemap are layers 

`arrayLayerIndex%6` gives you the cubemap face
`arrayLayerIndex/6` gives you the cubemap array index

source/Irrlicht/CMakeLists.txt
devshgraphicsprogramming: well yeah, no point in duplicate loaders.

src/irr/asset/CGLILoader.cpp
devshgraphicsprogramming: good.

src/irr/asset/CGLILoader.cpp
AnastaZIuk: @devshgraphicsprogramming what do you think about it?

devshgraphicsprogramming: Thanks to `GL_ARB_texture_storage` we now deal with exclusively internal formats
https://github.com/g-truc/gli/blob/master/gli/gl.hpp#L16

Before OpenGL was a retarded mix of extenal+type format, i.e. `GL_RGBA` + `GL_FLOAT`

Now you just have `GL_RGBA32F`

You should ignore swizzles and target types (well you still need to differentiate 1D, 2D and 3D) for Images and put them in the ImageView creation struct instead, arbitrary swizzles are possible.

src/irr/asset/CGLILoader.cpp
AnastaZIuk: hm, actually it seems to me that our image assumes that the "layer" is always one, but if it is an cubemap it can handle some "layers" considered by our API as faces, right?
so if there were some layers, I have to create pointers of our **ICPUImage** and if target is an cubemap, I can set **arrayLayers** that stores our faces

AnastaZIuk: but in that case where there are loaded layers I have to consider mipmap level if present and create **SCreationParams** for each image, because it won't match anymore

devshgraphicsprogramming: only one ICPUImage is ever produced, "layers" are no different than "depth" and function exactly like "height" pixel addressing.

`((z*height+y)*width+x)*pixelSize`

There's even a function that gets you an offset into a bufer buffer I believe.

ICPUImage consists of declaration of image size+format+etc plus a list of buffer regions that shall be copied into it, see the "fixed" PNG loader, separate mip-map levels are separate image subresources, hence a separate buffer-region to copy. You don't however need 1 buffer per copy region, you can pack everything into one without any problem (that's what offsets are for).

See Vulkan-Tutorial and read documentation of `vkCmdCopyBufferToImage`

include/IrrCompileConfig.h
```diff
@@ -194,22 +194,21 @@ the engine will no longer read .png images. */
...
-#endif
 //! Define _IRR_COMPILE_WITH_TGA_LOADER_ if you want to load .tga files
 #define _IRR_COMPILE_WITH_TGA_LOADER_
 #ifdef NO_IRR_COMPILE_WITH_TGA_LOADER_
 #undef _IRR_COMPILE_WITH_TGA_LOADER_
 #endif
-
+//! Define _IRR_COMPILE_WITH_GLI_LOADER_ if you want to load .ktx, .dds and .kmg files
+// Pay attention:
+// a) DDS doesn't compile on 64-bit currently (not sure if it is true currently)
```
devshgraphicsprogramming, line 204: Why?

include/IrrCompileConfig.h
```diff
@@ -194,22 +194,21 @@ the engine will no longer read .png images. */
...
+// b) anyone enabling it should be aware that S3TC compression algorithm which might be used in that loader which loads DDS files
+// is patented in the US by S3 and they do collect license fees when it's used in applications.
```
devshgraphicsprogramming, line 206: patent expired in 2019

devshgraphicsprogramming: also decompression is patented, not transfer of bytes, sending compressed data stream to OpenGL driver puts the onus of licensing on OpenGL driver implementation not the application using OpenGL
(this is why MESA can't have nice things)

include/IrrCompileConfig.h
```diff
@@ -194,22 +194,21 @@ the engine will no longer read .png images. */
...
+// So if you are unfortunate enough to develop applications for US market and their broken patent system be careful.
+#define _IRR_COMPILE_WITH_GLI_LOADER_
```
devshgraphicsprogramming, line 208: this sort of shit should be a cmake option then go into the buildconfig dynaically configured file.

Look at the first 6 files here
https://github.com/buildaworldnet/IrrlichtBAW/pull/381/files

If you want to you can convert all the other `_IRR_COMPILE_WITH` to cmake options.

src/irr/asset/CGLILoader.cpp
```diff
@@ -46,112 +46,288 @@ namespace irr
...
+			{
+				glm::tvec3<GLsizei> extent(texture.extent(mipmapLevel));
+
+				ICPUImage::SCreationParams imageInfo;
+				imageInfo.format = format;
+				imageInfo.extent.width = extent.x;
+				imageInfo.extent.height = extent.y;
+				imageInfo.extent.depth = extent.z;
+				imageInfo.mipLevels = texture.levels();
+				imageInfo.arrayLayers = texture.faces();
```
devshgraphicsprogramming: arrayLayers = totalFaces = faces()*layers()

src/irr/asset/CGLILoader.cpp
```diff
@@ -46,112 +46,288 @@ namespace irr
...
-							{
-								assert(0);
-								break;
-							}
-						}
+		    const gli::gl glVersion(gli::gl::PROFILE_GL33);
+			const GLenum target = glVersion.translate(texture.target());
+			const E_FORMAT format = getTranslatedGLIFormat(texture, glVersion);
+
+			ICPUImage* rawImagesBundle = _IRR_NEW_ARRAY(ICPUImage, texture.levels());
```
devshgraphicsprogramming: nope, see my other comment.

src/irr/asset/CGLILoader.cpp
```diff
@@ -46,112 +46,288 @@ namespace irr
...
+			return SAssetBundle({ core::smart_refctd_ptr<IAsset>(rawImagesBundle, core::dont_grab) });
 		}
 
 		bool CBufferLoaderBIN::isALoadableFileFormat(io::IReadFile* _file) const
 		{
-			return true; // gli provides a function to load files, but we can check files' signature actally if needed
+			return true; // gli provides a function to load files, but we can check files' signature actually if needed
+		}
+
+		inline E_FORMAT getTranslatedGLIFormat(const gli::texture& texture, const gli::gl& glVersion)
```
devshgraphicsprogramming: i think you'll need to return a `E_FORMAT,ICPUImageView::SCreationParams::Swizzle` pair

src/irr/asset/CGLILoader.h
```diff
@@ -37,6 +37,8 @@ namespace irr
 				io::IReadFile* file;
 				core::smart_refctd_ptr<ICPUBuffer> sourceCodeBuffer;
 			};
+
+			inline E_FORMAT getTranslatedGLIFormat(const gli::texture& texture, const gli::gl& glVersion);
```
devshgraphicsprogramming: static

devshgraphicsprogramming:
You need to view the ICPUImage as the full OpenGL image, all layers and mip-levels included.

ICPUImage gets its pixels filled by a series of buffer to image copies like in vulkan.

src/irr/asset/CGLILoader.h
```diff
@@ -5,30 +5,31 @@
...
+			virtual bool isALoadableFileFormat(io::IReadFile* _file) const override;
+
+			virtual const char** getAssociatedFileExtensions() const override
 			{
 				static const char* extensions[]{ "ktx", "dds", "kmg", nullptr };
 				return extensions;
 			}
 
-			uint64_t getSupportedAssetTypesBitfield() const override { return asset::IAsset::ET_IMAGE; }
+			virtual uint64_t getSupportedAssetTypesBitfield() const override { return asset::IAsset::ET_IMAGE; }
```
devshgraphicsprogramming: image view not image

also you don't need virtual here, dont use virtual here, make the class final

src/irr/asset/CGLILoader.cpp
AnastaZIuk: Sure, but Swizzles are placed in **SComponentMapping** struct

devshgraphicsprogramming: yeah I don't remember all the structs and nested class declarations off by heart ;)

src/irr/asset/CGLILoader.cpp
```diff
@@ -0,0 +1,434 @@
...
+					break;
+				}
+				default:
+				{
+					assert(0);
+					break;
+				}
+			}
+
+			std::pair<core::smart_refctd_dynamic_array<ICPUImage>, core::smart_refctd_dynamic_array<ICPUImageView>> bundle
```
AnastaZIuk, line 110: @devshgraphicsprogramming I think I've got it wrong while reading about copyBuffers
It seems to me that there should be one imageView and images in amount of mipmaps and such images will be filled by copy buffers to finaly create imageView

I need to read it one more time for sure

src/irr/asset/CGLILoader.cpp
```diff
@@ -48,284 +48,385 @@ namespace irr
...
+			{
+				case gli::TARGET_1D:
+				{
+					baseImageType = IImage::ET_1D;
+					imageViewType = ICPUImageView::ET_1D;
+					break;
+				}
+				case gli::TARGET_1D_ARRAY:
+				{
+					imageViewType = ICPUImageView::ET_1D_ARRAY;
```
devshgraphicsprogramming, line 69: baseImageType becomes 1D here IIRC

src/irr/asset/CGLILoader.cpp
```diff
@@ -48,284 +48,385 @@ namespace irr
...
+				}
+				case gli::TARGET_2D:
+				{
+					baseImageType = IImage::ET_2D;
+					imageViewType = ICPUImageView::ET_2D;
+					break;
+				}
+				case gli::TARGET_2D_ARRAY:
+				{
+					imageViewType = ICPUImageView::ET_2D_ARRAY;
```
devshgraphicsprogramming, line 81: baseImageType becomes 2D here IIRC

src/irr/asset/CGLILoader.cpp
```diff
@@ -48,284 +48,385 @@ namespace irr
...
+				}
+				case gli::TARGET_3D:
+				{
+					baseImageType = IImage::ET_3D;
+					imageViewType = ICPUImageView::ET_3D;
+					break;
+				}
+				case gli::TARGET_CUBE:
+				{
+					imageViewType = ICPUImageView::ET_CUBE_MAP;
```
devshgraphicsprogramming, line 93: baseImageType becomes 2D here IIRC

src/irr/asset/CGLILoader.cpp
```diff
@@ -48,284 +48,385 @@ namespace irr
...
+					break;
+				}
+				case gli::TARGET_CUBE_ARRAY:
+				{
+					imageViewType = ICPUImageView::ET_CUBE_MAP_ARRAY;
```
devshgraphicsprogramming, line 99: baseImageType same as non-array version

src/irr/asset/CGLILoader.cpp
```diff
@@ -48,284 +48,385 @@ namespace irr
...
+					break;
+				}
+				case gli::TARGET_COUNT:
```
devshgraphicsprogramming: just collapse this into the default case.

src/irr/asset/CGLILoader.cpp
devshgraphicsprogramming: One Image and one ImageView

There will be one copy range struct per mip-map (but it will point to 1 image and 1 buffer)

src/irr/asset/CGLILoader.cpp
```diff
@@ -48,284 +48,385 @@ namespace irr
...
+
+						ICPUImage::SBufferCopy& region = *(regions->begin() + index);
+						region.imageSubresource.mipLevel = 0u;
+						region.imageSubresource.baseArrayLayer = 0u;
+						region.imageSubresource.layerCount = 1u;
+						region.bufferOffset = 0u;
+						region.bufferRowLength = extent[0] * texelFormatByteSize;
+						region.bufferImageHeight = 0u; //tightly packed
+						region.imageOffset = { 0u, 0u, 0u };
+						region.imageExtent = {extent[0], extent[1], extent[2]};
```
devshgraphicsprogramming: you need to vary `mipLevel` and `bufferOffset` accordingly

Also are you sure GLI will give tightly packed image rows?

devshgraphicsprogramming: Finally `extent` shrinks for each mip level, the ICPUImage or IImage has a utlity function that actually tells you the extent of a given mip-level.

src/irr/asset/CGLILoader.cpp
```diff
@@ -48,284 +48,385 @@ namespace irr
...
+				std::make_pair(gl::SWIZZLE_GREEN, ICPUImageView::SComponentMapping::ES_G),
+				std::make_pair(gl::SWIZZLE_BLUE, ICPUImageView::SComponentMapping::ES_B),
+				std::make_pair(gl::SWIZZLE_ALPHA, ICPUImageView::SComponentMapping::ES_A),
+				std::make_pair(gl::SWIZZLE_ONE, ICPUImageView::SComponentMapping::ES_ONE),
+				std::make_pair(gl::SWIZZLE_ZERO, ICPUImageView::SComponentMapping::ES_ZERO)
+			};
+
+			auto getMappedSwizzle = [&](const gli::gl::swizzle& currentSwizzleToCheck)
+			{
+				for (auto& mappedSwizzle : swizzlesMappingAPI)
```
devshgraphicsprogramming, line 205: this is stupid, `gl::SWIZZLE` is an enum, you can just make an array and just index it by the swizzle

src/irr/asset/CGLILoader.cpp
```diff
@@ -48,284 +48,385 @@ namespace irr
...
-				case gl::INTERNAL_RGB_UNORM: return EF_R8G8B8_UNORM;			//GL_RGB
-				case gl::INTERNAL_BGR_UNORM: return;	//GL_BGR
-				case gl::INTERNAL_RGBA_UNORM: return;		//GL_RGBA
-				case gl::INTERNAL_BGRA_UNORM: return;		//GL_BGRA
-				case gl::INTERNAL_BGRA8_UNORM: return;		//GL_BGRA8_EXT
+				case gl::INTERNAL_RGB_UNORM: return getTranslatedFinalFormat(EF_R8G8B8_UNORM);			//GL_RGB
+				case gl::INTERNAL_BGR_UNORM: return getTranslatedFinalFormat();			//GL_BGR
+				case gl::INTERNAL_RGBA_UNORM: return getTranslatedFinalFormat();		//GL_RGBA
+				case gl::INTERNAL_BGRA_UNORM: return getTranslatedFinalFormat();		//GL_BGRA
+				case gl::INTERNAL_BGRA8_UNORM: return getTranslatedFinalFormat();		//GL_BGRA8_EXT
```
devshgraphicsprogramming, line 227: you'll have to inver the swizzle for some of these

src/irr/asset/CGLIWriter.h
```diff
@@ -25,13 +25,13 @@ namespace irr
 				return extensions;
 			}
 
-			virtual uint64_t getSupportedAssetTypesBitfield() const override { return asset::IAsset::ET_IMAGE; }
+			uint64_t getSupportedAssetTypesBitfield() const override { return asset::IAsset::ET_IMAGE; }
```
devshgraphicsprogramming: IMAGE_VIEW

src/irr/asset/CGLILoader.cpp
AnastaZIuk: sure, but it seems to me it won't match because of that:
https://github.com/g-truc/gli/blob/master/gli/gl.hpp#L314

src/irr/asset/CGLIWriter.h
```diff
@@ -25,13 +25,13 @@ namespace irr
...
 
-			virtual uint32_t getSupportedFlags() override { return asset::EWF_NONE; }
+			uint32_t getSupportedFlags() override { return asset::EWF_NONE; }
 
-			virtual uint32_t getForcedFlags() override { return 0u; }
+			uint32_t getForcedFlags() override { return 0u; }
```
devshgraphicsprogramming: write binary? anything elsE?

src/irr/asset/CGLIWriter.h
AnastaZIuk: I missed it, mistake

src/irr/asset/CGLILoader.cpp
devshgraphicsprogramming: use `core::unordered_map` then

devshgraphicsprogramming: a `static const core::unordered_map`

src/irr/asset/CGLILoader.cpp
AnastaZIuk: > the ICPUImage or IImage has a utlity function that actually tells you the extent of a given mip-level.

well, I get extent relative to mipmapLevel by loaded GLI texture that has a capability of getting an extent for each mip level, but I will remember it

src/irr/asset/CGLILoader.cpp
AnastaZIuk: is it for sure necessary? gli reads it and assigns it properly I think, but I do an extra mapping on our API

src/irr/asset/CGLILoader.cpp
AnastaZIuk: https://github.com/buildaworldnet/IrrlichtBAW/pull/402/files#diff-28a7c4ae6aba8ec71c707eede79b01e9R206 if a swizzle is blue, it will be assigned to red so forth, so it seems to me it is correct

src/irr/asset/CGLILoader.cpp
AnastaZIuk: https://github.com/buildaworldnet/IrrlichtBAW/pull/402/files#diff-28a7c4ae6aba8ec71c707eede79b01e9R206 if swizzle of compoment red that will be assigned to it is blue, it will be extra mapped to our format and then assigned to

devshgraphicsprogramming: I'd assume that if gli reports `INTERNAL_BGRA8_UNORM` then RGBA = BGRA  and then there's an extra swizzle from gli on top of that

meaning that `INTERNAL_BGRA8_UNORM` needs to become `EF_R8G8B8A8_UNORM` + a swizzle that is

RED = BLUE
GREEN = GREEN
BLUE = RED
ALPHA = ALPHA

src/irr/asset/CGLILoader.cpp
devshgraphicsprogramming: what is this last comment about!?

src/irr/asset/CGLILoader.cpp
AnastaZIuk: But that's what my mapping does, it respects it 
Otherwise why does gli read swizzle on 4 size Vector?

devshgraphicsprogramming: well but what happens if you have the BGRA format + a custom swizzle? OR will GLI not give you such?

AnastaZIuk: the default is `RGBA`, without swizzle assumption
standard swizzle vector with size of 4 also assume that it is `RGBA` and it returns enum swizze to each chanel R as [0], G as [1] and so forth I think. In that case, if my format in case statement matches `BGRA` format, we will have to consider that format is got from texture contating loaded parameters, swizzles though. Mapped swizzles (without mapping extraordinary to our API) are in format order therefore, right? 
For instance:

format[4] (where there is r, g, b, a default)
read swizzle on [0] (`r`), GLI returns on it SWIZZLE_BLUE, so I map it and get 'b'  
read swizzle on [1] (`g`), GLI returns on it SWIZZLE_GREEN, so I map it and get 'g'  
read swizzle on [2] (`b`), GLI returns on it SWIZZLE_RED, so I map it and get 'r'  
read swizzle on [3] (`a`), GLI returns on it SWIZZLE_ALPHA, so I map it and get 'a' 

swizzle output -> `BGRA` 
and there is mapping to our API left

> OR will GLI not give you such?

I have to handle it, right

devshgraphicsprogramming: just test and see what assumptions are correct, I know nothing about how GLI works, only our library.

src/irr/asset/CGLILoader.cpp
devshgraphicsprogramming: btw you opened the issue on a fork of the original repo

include/IrrCompileConfig.h
devshgraphicsprogramming: get rid of this useles comment

include/IrrCompileConfig.h
devshgraphicsprogramming: comment needs to go

AnastaZIuk:
close due to moving to https://github.com/buildaworldnet/IrrlichtBAW/pull/411

include/IrrCompileConfig.h
devshgraphicsprogramming: status? @Przemog1

include/IrrCompileConfig.h
devshgraphicsprogramming: status? @Przemog1

src/irr/asset/CGLILoader.cpp
devshgraphicsprogramming: status? @Przemog1

src/irr/asset/CGLILoader.cpp
devshgraphicsprogramming: status? @Przemog1

src/irr/asset/CGLILoader.cpp
devshgraphicsprogramming: status? @Przemog1

src/irr/asset/CGLILoader.cpp
devshgraphicsprogramming: status? @Przemog1

PR #406, opened 2019-12-13
TGA Loader

Przemog1:


src/irr/asset/CImageLoaderTGA.cpp
```diff
@@ -296,7 +296,71 @@ asset::SAssetBundle CImageLoaderTGA::loadAsset(io::IReadFile* _file, const asset
     for (auto& img : images)
         img->drop();
     return SAssetBundle({core::smart_refctd_ptr<IAsset>(tex, core::dont_grab)});
+
+
+
 #else
+	core::vector<asset::ICPUImage*> images;
```
devshgraphicsprogramming: you will only create ONE ICPUImage (previous system was fucked up thats why the API was changed).

The ICPUImage contains all mip-maps (but here you will only have one mip).

src/irr/asset/CImageLoaderTGA.cpp
```diff
@@ -296,7 +296,71 @@ asset::SAssetBundle CImageLoaderTGA::loadAsset(io::IReadFile* _file, const asset
...
+
+	switch (header.ImageType)
+	{
+	case 1: // uncompressed color-mapped image
+	case 2: // uncompressed rgb image
+	case 3: // uncompressed grayscale image
+	case 10:
```
devshgraphicsprogramming: well you're falling through your switch statement, so all non-rgba format images will try to load with wrong format.

src/irr/asset/CImageLoaderTGA.cpp
```diff
@@ -296,7 +296,71 @@ asset::SAssetBundle CImageLoaderTGA::loadAsset(io::IReadFile* _file, const asset
...
+	{
+		const uint64_t imageSize = header.ImageHeight * header.ImageWidth * header.PixelDepth / 8;
+		void* data = new uint8_t[imageSize];
+		core::smart_refctd_ptr<ICPUBuffer> imageBuffer(new ICPUBuffer(imageSize));
+		void* imageBufferDataPointer = imageBuffer->getPointer();
+		auto regions = core::make_refctd_dynamic_array<core::smart_refctd_dynamic_array<ICPUImage::SBufferCopy>>(1u);
```
devshgraphicsprogramming: You can create the ICPU buffer first, and use it instead of `data`

src/irr/asset/CImageLoaderTGA.cpp
Przemog1: I know, but it is for testing purpose, so far I'm trying to load only rgba image

src/irr/asset/CImageLoaderTGA.cpp
```diff
@@ -296,7 +296,71 @@ asset::SAssetBundle CImageLoaderTGA::loadAsset(io::IReadFile* _file, const asset
...
+	switch (header.ImageType)
+	{
+	case 1: // uncompressed color-mapped image
+	case 2: // uncompressed rgb image
+	case 3: // uncompressed grayscale image
+	case 10:
+	{
+		const uint64_t imageSize = header.ImageHeight * header.ImageWidth * header.PixelDepth / 8;
+		void* data = new uint8_t[imageSize];
+		core::smart_refctd_ptr<ICPUBuffer> imageBuffer(new ICPUBuffer(imageSize));
```
devshgraphicsprogramming: this will leak (double grab), use `auto imageBuffer = core::make_smart_refctd_ptr<ICPUBuffer>(imageSize);` instead

src/irr/asset/CImageLoaderTGA.cpp
```diff
@@ -296,7 +296,71 @@ asset::SAssetBundle CImageLoaderTGA::loadAsset(io::IReadFile* _file, const asset
...
+
+		ICPUImage::SCreationParams imgInfo;
+		imgInfo.type = ICPUImage::ET_2D;
+		imgInfo.extent.width = header.ImageWidth;
+		imgInfo.extent.height = header.ImageHeight;
+		imgInfo.extent.depth = 1u;
+		imgInfo.mipLevels = 1u;
+		imgInfo.arrayLayers = 1u;
+		imgInfo.samples = ICPUImage::ESCF_1_BIT;
+		imgInfo.format = asset::EF_R8G8B8A8_SRGB;
```
devshgraphicsprogramming: all of the above are good, you just need to pick the format correctly.

src/irr/asset/CImageLoaderTGA.cpp
```diff
@@ -296,7 +296,71 @@ asset::SAssetBundle CImageLoaderTGA::loadAsset(io::IReadFile* _file, const asset
...
+		imgInfo.flags = static_cast<IImage::E_CREATE_FLAGS>(0u);
+
+		core::smart_refctd_ptr<ICPUImage> image = ICPUImage::create(std::move(imgInfo));
+		image.get()->setBufferAndRegions(std::move(imageBuffer), regions);
```
devshgraphicsprogramming: no need for a `.get()` the `->` operator is already overloaded.

src/irr/asset/CImageLoaderTGA.cpp
devshgraphicsprogramming: ok dude, you need to fill out `regions` with apropriate values, you have uninitialized memory there right now

src/irr/asset/CImageLoaderTGA.cpp
devshgraphicsprogramming: you can delete this variable actually.

src/irr/asset/CImageLoaderTGA.cpp
Przemog1: right

AnastaZIuk:
If you ever had to allocate raw, don't use `new` operator, but use `_IRR_NEW`/`_IRR_NEW_ARRAY` and 
`_IRR_DELETE`/ `_IRR_DELETE_ARRAY` instead https://github.com/buildaworldnet/IrrlichtBAW/pull/406/commits/5b751316cea43adfd251a41140ba965b65e9bd08#diff-c5034cdfa9e2b85a7956da5e84f1d6bbL316

src/irr/asset/CImageLoaderTGA.cpp
```diff
@@ -301,59 +301,135 @@ asset::SAssetBundle CImageLoaderTGA::loadAsset(io::IReadFile* _file, const asset
...
+	{
+		if (header.ImageType != 3)
+		{
+			os::Printer::log("Loading 8-bit non-grayscale is NOT supported.", ELL_ERROR);
+			return {};
+		}
 
-	//it is temporary and for testing purpose
-	case 10:
+		imgInfo.format = asset::EF_R8_SRGB;
```
devshgraphicsprogramming, line 329: this format is not supported widely, this is why I had a function here originally that would convert R8_SRGB to RGB8_SRGB in software.

devshgraphicsprogramming:
Great job!

Let me know when finished.

devshgraphicsprogramming:
How is the status?

src/irr/asset/CImageLoaderTGA.cpp
AnastaZIuk: @devshgraphicsprogramming in that should I calculate average of channels and perform grayscale or should I fill `g` and `b` channels with certain default values like 0?

AnastaZIuk: I see we have someting like `video::convertColor`, but I'm not sure if it'll solve it though

devshgraphicsprogramming: no what  Imean is that you should duplicare the R8_SRGB channel to fill RGB8_SRGB

so its basically is:
```glsl
out.rgb = in.rrr;
```

AnastaZIuk: okay, great

PR #412, opened 2020-01-14
TGA loader

AnastaZIuk:
Status:

- **done**

### DIFF
--- a/include/irr/asset/format/convertColor.h
+++ b/include/irr/asset/format/convertColor.h
@@ -114,7 +114,7 @@ namespace irr { namespace video
             using decT = typename std::conditional<isSignedFormat<sF>(), int64_t, uint64_t>::type;
             using encT = typename std::conditional<isSignedFormat<dF>(), int64_t, uint64_t>::type;
 
-            decT decbuf[4];
+            decT decbuf[4] {0, 0, 0, 0};
             impl::SCallDecode<sF, decT>{}(srcPix, decbuf, _blockX, _blockY);
 			SWIZZLE(decbuf)
             impl::SCallEncode<dF, encT>{}(dstPix, reinterpret_cast<encT*>(decbuf));
@@ -126,7 +126,7 @@ namespace irr { namespace video
             using decT = double;
             using encT = double;
 
-            decT decbuf[4];
+            decT decbuf[4] { 0, 0, 0, 1 };
             impl::SCallDecode<sF, decT>{}(srcPix, decbuf, _blockX, _blockY);
 			SWIZZLE(decbuf)
             impl::SCallEncode<dF, encT>{}(dstPix, decbuf);
@@ -136,7 +136,7 @@ namespace irr { namespace video
             using decT = double;
             using encT = typename std::conditional<isSignedFormat<dF>(), int64_t, uint64_t>::type;
 
-            decT decbuf[4];
+            decT decbuf[4] { 0, 0, 0, 0 };
             impl::SCallDecode<sF, decT>{}(srcPix, decbuf, _blockX, _blockY);
 			SWIZZLE(decbuf)
             encT encbuf[4];
@@ -149,7 +149,7 @@ namespace irr { namespace video
             using decT = typename std::conditional<isSignedFormat<sF>(), int64_t, uint64_t>::type;
             using encT = double;
 
-            decT decbuf[4];
+            decT decbuf[4] { 0, 0, 0, 0 };
             impl::SCallDecode<sF, decT>{}(srcPix, decbuf, _blockX, _blockY);
 			SWIZZLE(decbuf)
             encT encbuf[4];
--- a/include/irr/asset/format/convertColor.h
+++ b/include/irr/asset/format/convertColor.h
@@ -114,7 +114,7 @@ namespace irr { namespace video
             using decT = typename std::conditional<isSignedFormat<sF>(), int64_t, uint64_t>::type;
             using encT = typename std::conditional<isSignedFormat<dF>(), int64_t, uint64_t>::type;
 
-            decT decbuf[4] {0, 0, 0, 0};
+            decT decbuf[4] {0, 0, 0, 1};
             impl::SCallDecode<sF, decT>{}(srcPix, decbuf, _blockX, _blockY);
 			SWIZZLE(decbuf)
             impl::SCallEncode<dF, encT>{}(dstPix, reinterpret_cast<encT*>(decbuf));
@@ -136,7 +136,7 @@ namespace irr { namespace video
             using decT = double;
             using encT = typename std::conditional<isSignedFormat<dF>(), int64_t, uint64_t>::type;
 
-            decT decbuf[4] { 0, 0, 0, 0 };
+            decT decbuf[4] { 0, 0, 0, 1 };
             impl::SCallDecode<sF, decT>{}(srcPix, decbuf, _blockX, _blockY);
 			SWIZZLE(decbuf)
             encT encbuf[4];
@@ -149,7 +149,7 @@ namespace irr { namespace video
             using decT = typename std::conditional<isSignedFormat<sF>(), int64_t, uint64_t>::type;
             using encT = double;
 
-            decT decbuf[4] { 0, 0, 0, 0 };
+            decT decbuf[4] { 0, 0, 0, 1 };
             impl::SCallDecode<sF, decT>{}(srcPix, decbuf, _blockX, _blockY);
 			SWIZZLE(decbuf)
             encT encbuf[4];
--- a/include/irr/asset/format/convertColor.h
+++ b/include/irr/asset/format/convertColor.h
@@ -159,8 +159,17 @@ namespace irr { namespace video
         }
 	#undef SWIZZLE
     }
+	/*
+		It converts a data to desired texel format
+
+		To use it, you have to pass source data with texel format \bsF\b you want to exchange
+		with \bdF\b. srcPix data is an array due to planar formats, but if source data isn't
+		a planar format, you can use \bsrcPix[0]\b.
+		\b_pixOrBlockCnt\b is an amount of texels you want to convert and \b_imgSize\b is a size
+		in texels of an image they belong to.
+	*/
     template<asset::E_FORMAT sF, asset::E_FORMAT dF, class Swizzle = DefaultSwizzle >
-    inline void convertColor(const void* srcPix[4], void* dstPix, size_t _pixOrBlockCnt, core::vector3d<uint32_t>& _imgSize, PolymorphicSwizzle* swizzle = nullptr)
+    inline void convertColor(const void* srcPix[4], void* dstPix, size_t _pixOrBlockCnt, const core::vector3d<uint32_t>& _imgSize, PolymorphicSwizzle* swizzle = nullptr)
     {
         using namespace asset;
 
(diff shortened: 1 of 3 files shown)


devshgraphicsprogramming:
so GLI does not support TGA?

AnastaZIuk:
> so GLI does not support TGA?

actually it seems to me it does https://github.com/g-truc/gli/blob/master/test/main.cpp#L34
so what is the decision?

AnastaZIuk:
@devshgraphicsprogramming should I get rid of the loader and replace it with gli tga loader?

devshgraphicsprogramming:
yes ofc you should as long as you can get the same functionality out of it!

AnastaZIuk:
checked, there is no opportunity to use GLI for loading TGA - TGA Loader will be refactored

include/irr/asset/format/convertColor.h
```diff
@@ -159,8 +159,17 @@ namespace irr { namespace video
...
     }
+	/*
+		It converts a data to desired texel format
+
+		To use it, you have to pass source data with texel format \bsF\b you want to exchange
+		with \bdF\b. srcPix data is an array due to planar formats, but if source data isn't
+		a planar format, you can use \bsrcPix[0]\b.
+		\b_pixOrBlockCnt\b is an amount of texels you want to convert and \b_imgSize\b is a size
+		in texels of an image they belong to.
+	*/
```
devshgraphicsprogramming, line 180: polymorphic swizzle is a class that can swizzle the RGBA componenents into a different arrangement at runtime, using it (non null pointer) obviously makes the whole process slower.

src/irr/asset/CImageLoaderTGA.cpp
```diff
@@ -115,20 +115,21 @@ static void convertColorFlip(const core::smart_refctd_ptr<ICPUImage> &image, con
 {
 	const T *in = (const T *) src;
 	T *out = (T *) image->getBuffer()->getPointer();
-	
-	auto size = image->getCreationParameters().extent;
+	const auto extent = image->getCreationParameters().extent;
+	irr::core::vector3d size = {extent.width, extent.height, extent.depth};
```
devshgraphicsprogramming: get the values with padding maybe? (or can you guarantee that you always use one region for the whole and only mip and its without padding)

src/irr/asset/CImageLoaderTGA.cpp
```diff
@@ -179,7 +179,7 @@ asset::SAssetBundle CImageLoaderTGA::loadAsset(io::IReadFile* _file, const asset
 	imgInfo.type = ICPUImage::ET_2D;
 	imgInfo.extent.width = header.ImageWidth;
 	imgInfo.extent.height = header.ImageHeight;
-	imgInfo.extent.depth = 1u;
+	imgInfo.extent.depth = header.PixelDepth / 8; // not sure about it
```
devshgraphicsprogramming: no depth as in dimension size depth is 1u (image is always 2D)

PixelDepth from TGA header probably means how many bits per pixel (color depth), so its used for picking the correct E_FORMAT

src/irr/asset/CImageLoaderTGA.cpp
```diff
@@ -276,6 +276,20 @@ asset::SAssetBundle CImageLoaderTGA::loadAsset(io::IReadFile* _file, const asset
 					// Do an OpenGL flip
 					convertColorFlip<uint8_t, EF_R8_SRGB, EF_R8_SRGB>(image, data, true);
 				}
+
+				const void* planarData[] = { data , nullptr, nullptr, nullptr};
+				void* outRGBData = data;
+				const size_t wholeSize = region.imageExtent.height * region.bufferRowLength * region.imageExtent.depth;
+				const auto wholeSizeInBytes = wholeSize * getTexelOrBlockBytesize(EF_R8G8B8_SRGB);
+
+				core::vector3d<uint32_t> imageSize;
```
devshgraphicsprogramming: you could just reinterpret cast, since its same bytewise stuff.

devshgraphicsprogramming:
Tell me when done and tested.

AnastaZIuk:
> Tell me when done and tested.

@devshgraphicsprogramming done and tested, but I had to insert some code into some switch cases, it looks awful, so I will do a small clean code refactor tomorrow and then it will be ready to merge

src/irr/asset/CImageLoaderTGA.cpp
```diff
@@ -273,32 +275,40 @@ asset::SAssetBundle CImageLoaderTGA::loadAsset(io::IReadFile* _file, const asset
...
+				video::convertColor<EF_R8_SRGB, EF_R8G8B8_SRGB>(planarData, outRGBData, wholeSize, *reinterpret_cast<core::vector3d<uint32_t>*>(&region.imageExtent));
+
+				texelBuffer = core::make_smart_refctd_ptr<ICPUBuffer>(wholeSizeInBytes);
+
+				memcpy(texelBuffer->getPointer(), outRGBData, wholeSizeInBytes);
+				_IRR_DELETE_ARRAY(outRGBData, wholeSizeInBytes); // it involves size in R8G8B8
+
+				image->setBufferAndRegions(std::move(texelBuffer), regions);
+
 				// Targa formats needs two y-axis flips. The first is a flip to get the Y conforms to OpenGL coords.
```
devshgraphicsprogramming: it would be easier on the CPU and memory to flip before expanding to 3x the size during the convert

src/irr/asset/CImageLoaderTGA.cpp
```diff
@@ -232,7 +235,6 @@ asset::SAssetBundle CImageLoaderTGA::loadAsset(io::IReadFile* _file, const asset
 		{
 			region.bufferRowLength = calcPitchInBlocks(region.imageExtent.width, getTexelOrBlockBytesize(EF_A1R5G5B5_UNORM_PACK16));
 			const auto bufferSize = endBufferSize = region.imageExtent.height * region.bufferRowLength * header.PixelDepth / 8;
-			texelBuffer = core::make_smart_refctd_ptr<ICPUBuffer>(bufferSize);
 			data = loadCompressedImage(_file, header);
```
devshgraphicsprogramming: why not make `data` an ICPUBuffer that you can then use directly?

src/irr/asset/CImageLoaderTGA.cpp
```diff
@@ -273,32 +275,40 @@ asset::SAssetBundle CImageLoaderTGA::loadAsset(io::IReadFile* _file, const asset
...
-				memcpy(data, outRGBData, wholeSizeInBytes);
-				_IRR_DELETE_ARRAY(outRGBData, wholeSizeInBytes);
 			}
 			break;
 		case 16:
 			{
 				imgInfo.format = asset::EF_A1R5G5B5_UNORM_PACK16;
 
+				texelBuffer = core::make_smart_refctd_ptr<ICPUBuffer>(endBufferSize);
+				memcpy(texelBuffer->getPointer(), data, endBufferSize);
```
devshgraphicsprogramming: https://github.com/buildaworldnet/IrrlichtBAW/pull/412/files/97860b3deaf472bfa2ba7df541aab63f8ceb074b..a80c4362992c58b2f38eb75732fc19b149146a90#r374048643

src/irr/asset/CImageLoaderTGA.cpp
```diff
@@ -308,6 +318,10 @@ asset::SAssetBundle CImageLoaderTGA::loadAsset(io::IReadFile* _file, const asset
 		case 24:
 			{
 				imgInfo.format = asset::EF_R8G8B8_SRGB;
+
+				texelBuffer = core::make_smart_refctd_ptr<ICPUBuffer>(endBufferSize);
+				memcpy(texelBuffer->getPointer(), data, endBufferSize);
```
devshgraphicsprogramming: https://github.com/buildaworldnet/IrrlichtBAW/pull/412/files/97860b3deaf472bfa2ba7df541aab63f8ceb074b..a80c4362992c58b2f38eb75732fc19b149146a90#r374048643

src/irr/asset/CImageLoaderTGA.cpp
```diff
@@ -319,6 +333,10 @@ asset::SAssetBundle CImageLoaderTGA::loadAsset(io::IReadFile* _file, const asset
 			{
 				imgInfo.format = asset::EF_R8G8B8A8_SRGB;
 
+				texelBuffer = core::make_smart_refctd_ptr<ICPUBuffer>(endBufferSize);
+				memcpy(texelBuffer->getPointer(), data, endBufferSize);
```
devshgraphicsprogramming: https://github.com/buildaworldnet/IrrlichtBAW/pull/412/files/97860b3deaf472bfa2ba7df541aab63f8ceb074b..a80c4362992c58b2f38eb75732fc19b149146a90#r374048643

src/irr/asset/CImageLoaderTGA.cpp
```diff
@@ -86,6 +86,9 @@ bool CImageLoaderTGA::isALoadableFileFormat(io::IReadFile* _file) const
 		return false;
 	}
 	
+	/*
+		I can't load anything because of bellow
```
devshgraphicsprogramming: wait, why, what happened?

AnastaZIuk: Well, all of the images placed in `color_space_test` doesn't contain informations about gamma. I was trying to generate it through GIMP, but I couldn't as well, so I decided to bypass it

devshgraphicsprogramming: well sure, but leave the code for handling gamma, if there is no info about gamma then assume 2.233333333..............

AnastaZIuk: okay, I will patch it

AnastaZIuk:
@devshgraphicsprogramming convertColor needed patching with default values on texels, but nut sure about alpha. Afaik there are some rules defining how to put default alpha values when inteeger or normalized etc

src/irr/asset/CImageLoaderTGA.cpp
```diff
@@ -115,22 +114,20 @@ bool CImageLoaderTGA::isALoadableFileFormat(io::IReadFile* _file) const
 
 // convertColorFlip() does color conversion as well as taking care of properly flipping the given image.
 template <typename T, E_FORMAT srcFormat, E_FORMAT destFormat>
```
devshgraphicsprogramming: why is there a template on T, why are you incrementing the `in` and `out` by the same stride that is in pixels and not in bytes?

devshgraphicsprogramming:
> Afaik there are some rules defining how to put default alpha values when inteeger or normalized etc

Default value is (0,0,0,1) whatever channel is not defined, you take from that default vector.

src/irr/asset/CImageLoaderTGA.cpp
AnastaZIuk: because somebody has created it, but if it comes to stride, I need to fix it

devshgraphicsprogramming: just use uint8_t* always.

src/irr/asset/CImageLoaderTGA.cpp
```diff
@@ -129,8 +126,8 @@ static void convertColorFlip(uint32_t regionBufferRowLenght, VkExtent3D imageExt
 		imageExtent.depth
 	};
 
-	auto stride = regionBufferRowLenght;
 	auto channels = getFormatChannelCount(destFormat);
+	auto stride = regionBufferRowLenght * channels;
```
devshgraphicsprogramming: you're using `channels` wrong when in fact you should be using pixel block bytesize (TGA formats always have a 1x1 block physical size)

devshgraphicsprogramming:
Are we done?

src/irr/asset/CImageLoaderTGA.cpp
AnastaZIuk: my mistake, to fix

devshgraphicsprogramming: still fucking wrong.

PR #413, opened 2020-01-24
Fork GLM and GLEW, specify tags

AnastaZIuk:
GLM - 0.9.9.5 tag 
GLEW - glew-cmake-2.1.0 tag


devshgraphicsprogramming:
open a similar PR against every single on of yours and @Crisspl 's branches.

PR #415, opened 2020-01-30
Added parallel-hashmap

Quinost:


.gitmodules
```diff
@@ -73,3 +73,6 @@
 [submodule "3rdparty/CEGUI/glew-cmake"]
 	path = 3rdparty/CEGUI/glew
 	url = https://github.com/AnastaZIuk/glew-cmake.git
+[submodule "3rdparty/parallel-hashmap"]
+	path = 3rdparty/parallel-hashmap
+	url = https://github.com/WieszKto/parallel-hashmap
```
devshgraphicsprogramming: no need to have it come from your own fork, its generally ok to go straight from main repo if the owner is not a maniac (unlike some repos we've had to clone)

source/Irrlicht/CMakeLists.txt
```diff
@@ -209,6 +209,11 @@ set(COMMON_INCLUDE_DIRS
 irr_adjust_flags() # macro defined in root CMakeLists
 irr_adjust_definitions() # macro defined in root CMakeLists
 
+#Paraller hashmap
+macro(irr_target_link_parallel_hashmap _trgt)
```
devshgraphicsprogramming: since you're only adding includes I wouldn't call the macro `link`

devshgraphicsprogramming:
cool, now change the typedef for `core::unordered_map` and similar `unordered` to parallel hashmap stuff and test the library (especially smooth normal computation).

.gitmodules
devshgraphicsprogramming: ?

include/IrrCompileConfig.h
```diff
@@ -4,6 +4,7 @@
 
 #ifndef __IRR_COMPILE_CONFIG_H_INCLUDED__
 #define __IRR_COMPILE_CONFIG_H_INCLUDED__
+#define _ENABLE_EXTENDED_ALIGNED_STORAGE
```
devshgraphicsprogramming: wrap this in a test for MSVC compiler (only enable if MSVC detected, we have plenty of these tests in the engine headers, so just copy one)

PR #416, opened 2020-01-31
BaW V3 - left handed and right handed meshes info

AnastaZIuk:
Status:

**in progress**


devshgraphicsprogramming (commented):
This commit is huge, I can't view all the changes easily!

3rdparty/CMakeLists.txt
```diff
@@ -35,6 +35,22 @@ update_git_submodule(./glm)
...
+set(_OLD_BUILD_SHARED_LIBS ${BUILD_SHARED_LIBS})
+set(_OLD_SKIP_INSTALL_ALL ${SKIP_INSTALL_ALL})
+set(BUILD_SHARED_LIBS OFF)
+set(SKIP_INSTALL_ALL ON)
+add_subdirectory(zlib zlib EXCLUDE_FROM_ALL)
+set(SKIP_INSTALL_ALL ${_OLD_SKIP_INSTALL_ALL})
+set(BUILD_SHARED_LIBS ${_OLD_BUILD_SHARED_LIBS})
+get_target_property(ZLIB_BINARY_DIR zlibstatic "BINARY_DIR")
+get_target_property(ZLIB_NAME zlibstatic "NAME")
+get_target_property(ZLIB_DEBUG_POSTFIX zlibstatic "DEBUG_POSTFIX")
```
devshgraphicsprogramming, line 52: won't that make it `zlibstaticzlibstatic` ?

AnastaZIuk: I think it won't, why should it make it like that?. I've checked paths in some VS projects - seems to be done well + launched, so it works

AnastaZIuk:
@devshgraphicsprogramming so I noticed functions responsible for converting into new one version up should be moved to legacy section. Where directly? because we have a place where there are some namespaces for legacy stuff, but you have written about loader cpp as I remember

devshgraphicsprogramming:
you want to provide new struct definitions (patch) for `SkinnedMeshBlobV3` and `MeshBlobV3`

devshgraphicsprogramming:
> because we have a place where there are some namespaces for legacy stuff, but you have written about loader cpp as I remember

https://github.com/buildaworldnet/IrrlichtBAW/tree/master/src/irr/asset/bawformat/legacy

This is the resting place for V2 function and struct definitions (that will not be simply typedef'ed/aliased into being V3), you can delete as much as you want (like asset instantiation) as the loader only needs the definitions to convert from V2 to V3.

devshgraphicsprogramming:
You usually only need the V2 struct, and the conversion is done manually.

devshgraphicsprogramming:
>so I noticed functions responsible for converting into new one version up should be moved to legacy section

I'd make another .cpp file in https://github.com/buildaworldnet/IrrlichtBAW/tree/master/src/irr/asset/bawformat/legacy

Just for the implementations of the loader's file-convert-up functions

src/irr/asset/CBAWMeshFileLoader.cpp
```diff
@@ -202,6 +203,30 @@ SAssetBundle CBAWMeshFileLoader::loadAsset(io::IReadFile* _file, const asset::IA
 #endif // _IRR_DEBUG
 
     asset::ICPUMesh* mesh = reinterpret_cast<asset::ICPUMesh*>(retval);
+
+    if(_params.loaderFlags & E_LOADER_PARAMETER_FLAGS::ELPF_RIGHT_HANDED_MESHES)
```
devshgraphicsprogramming: this needs to happen in the blob instantiation, as you need to check the flag in the blob (if mesh was saved LH or RH) then only flip if `SavedFlag!=RequestedFlag`

AnastaZIuk: how to save it if there is no member for blobs for that - so no need for patching constructors for those blobs?
should it be saved in header of blob?

AnastaZIuk: I have talked to Criss recently and afaik it can't be saved in it (but not sure)

src/irr/asset/CBAWMeshWriter.cpp
```diff
@@ -153,6 +153,29 @@ struct LzmaMemMngmnt
 
         const ICPUMesh* mesh = static_cast<const ICPUMesh*>(_params.rootAsset);
 
+		if (!(_params.writerFlags & IAssetWriter::EWPF_MESH_IS_RIGHT_HANDED))
```
devshgraphicsprogramming: don't flip, just save the flag into the blob (left or right handed)

src/irr/asset/CBAWMeshFileLoader.cpp
```diff
@@ -253,39 +282,69 @@ io::IReadFile* CBAWMeshFileLoader::createConvertIntoVer_spec<3>(SContext& _ctx,
...
+        // should I actually do anything with meshes if their flipping is dependent on flag placed somewhere in IAssetLoader/Writer?
+        // so it isn't a member of any of blobs
 
-            offsetDiff += static_cast<int32_t>(hdr.blobSizeDecompr) - static_cast<int32_t>(hdr.effectiveSize());
-            hdr.compressionType = asset::Blob::EBCT_RAW;
-            core::XXHash_256(blob, hdr.blobSizeDecompr, hdr.blobHash);
-            hdr.blobSize = hdr.blobSizeDecompr;
+        switch (hdr.blobType)
+        {
+            case asset::Blob::EBT_FINAL_BONE_HIERARCHY:
```
devshgraphicsprogramming: there's no reason for you to touch the final bone hierarchy blob because you won't be changing it

AnastaZIuk: so should I delete it?

src/irr/asset/CBAWMeshWriter.cpp
AnastaZIuk: okay

src/irr/asset/CBAWMeshFileLoader.cpp
devshgraphicsprogramming: yes

src/irr/asset/CBAWMeshFileLoader.cpp
devshgraphicsprogramming: You save the left/right hand flag as an extra member of the meshbuffer blob.
(old files get upgraded by making the new version blob a blob that always has LH-flag)

If the flag does not agree, then when loading (in the finalization step), you need to create a new `ICPUBuffer` per meshbuffer just for the positions (and normals, if present) and write out flipped values to it (create a copy of the meshbuffer with the old POSITION and NORMALS buffers attached, then change the ATTR0 and ATTR3 buffers, and via the use of `getAttribute` and `setAttribute` you can extract, flip and save the new values)

include/irr/asset/bawformat/blobs/MeshBufferBlob.h
```diff
@@ -32,22 +37,18 @@ struct IRR_FORCE_EBO MeshBufferBlobV0 : TypedBlob<MeshBufferBlobV0, ICPUMeshBuff
 	uint32_t baseInstance;
 	uint32_t primitiveType;
 	uint32_t posAttrId;
+	bool isRightHandedCoordinateSystem;
 } PACK_STRUCT;
 #include "irr/irrunpack.h"
-static_assert(sizeof(MeshBufferBlobV0::mat)==197, "sizeof(MeshBufferBlobV0::mat) must be 197");
+static_assert(sizeof(MeshBufferBlobV3::mat)==197, "sizeof(MeshBufferBlobV0::mat) must be 197");
```
devshgraphicsprogramming, line 39: you need to patch this assert, because you added a member the struct will grow in size.

include/irr/asset/bawformat/blobs/MeshBufferBlob.h
```diff
@@ -32,22 +37,18 @@ struct IRR_FORCE_EBO MeshBufferBlobV0 : TypedBlob<MeshBufferBlobV0, ICPUMeshBuff
 	uint32_t baseInstance;
 	uint32_t primitiveType;
 	uint32_t posAttrId;
+	bool isRightHandedCoordinateSystem;
```
devshgraphicsprogramming: don't use `bool` they have undefined size across platforms, uint8_t is more appropriate.

include/irr/asset/bawformat/blobs/MeshBufferBlob.h
```diff
@@ -32,22 +37,18 @@ struct IRR_FORCE_EBO MeshBufferBlobV0 : TypedBlob<MeshBufferBlobV0, ICPUMeshBuff
...
-    + sizeof(MeshBufferBlobV0::primitiveType) + sizeof(MeshBufferBlobV0::posAttrId),
+    sizeof(MeshBufferBlobV3) ==
+    sizeof(MeshBufferBlobV3::mat) + sizeof(MeshBufferBlobV3::box) + sizeof(MeshBufferBlobV3::descPtr) + sizeof(MeshBufferBlobV3::indexType) + sizeof(MeshBufferBlobV3::baseVertex)
+    + sizeof(MeshBufferBlobV3::indexCount) + sizeof(MeshBufferBlobV3::indexBufOffset) + sizeof(MeshBufferBlobV3::instanceCount) + sizeof(MeshBufferBlobV3::baseInstance)
+    + sizeof(MeshBufferBlobV3::primitiveType) + sizeof(MeshBufferBlobV3::posAttrId) + sizeof(MeshBufferBlobV3::isRightHandedCoordinateSystem),
     "MeshBufferBlobV0: Size of blob is not sum of its contents!"
 );
 
-using MeshBufferBlobV1 = MeshBufferBlobV0;
-using MeshBufferBlobV2 = MeshBufferBlobV1;
```
devshgraphicsprogramming, line 47: still need aliases for Ver2 to Ver0

src/irr/asset/bawformat/CBAWFile.cpp
```diff
@@ -82,10 +82,28 @@ MeshBufferBlobV0::MeshBufferBlobV0(const asset::ICPUMeshBuffer* _mb)
...
+	indexType = meshBufferV2->indexType;
+	baseVertex = meshBufferV2->baseVertex;
+	indexCount = meshBufferV2->indexCount;
+	indexBufOffset = meshBufferV2->indexBufOffset;
+	instanceCount = meshBufferV2->instanceCount;
+	baseInstance = meshBufferV2->baseInstance;
+	primitiveType = meshBufferV2->primitiveType;
+	posAttrId = meshBufferV2->posAttrId;
+
+	isRightHandedCoordinateSystem = false;
```
devshgraphicsprogramming: this bad boy should be set by the writer.

src/irr/asset/bawformat/legacy/CBAWLegacy.h
```diff
@@ -137,6 +137,52 @@ using BAWFileV2 = BAWFileVn<2>;
 using BlobHeaderV2 = BlobHeaderVn<2>;
 
 
+#include "irr/irrpack.h"
+//! Simple struct of essential data of ICPUMeshBuffer that has to be exported
+struct IRR_FORCE_EBO MeshBufferBlobV2 : TypedBlob<MeshBufferBlobV2, ICPUMeshBuffer>, FixedSizeBlob<MeshBufferBlobV2, ICPUMeshBuffer>
+{
+    //! Constructor filling all members
+    explicit MeshBufferBlobV2(const ICPUMeshBuffer* _mb)
```
devshgraphicsprogramming: this constructor will never be used again, it can go 
(you will never make a new legacy blob from object, only new blob from old blob)

include/irr/asset/bawformat/blobs/MeshBufferBlob.h
AnastaZIuk: Matt is it neseccary if there is no change since Ver0 to Ver2 up? I have exchaned meshbuffers using 0 version (that are the same till V3) for version 2 and I find it valid, it compiles

include/irr/asset/bawformat/blobs/MeshBufferBlob.h
AnastaZIuk: I haven't known it yet though, to fix

src/irr/asset/bawformat/legacy/CBAWLegacy.h
AnastaZIuk: okay, to remove

include/irr/asset/bawformat/blobs/MeshBufferBlob.h
devshgraphicsprogramming: We always provide all the V0 to VLatest at least as aliases.

include/irr/asset/IAssetWriter.h
```diff
@@ -94,6 +94,12 @@ class IAssetWriter : public virtual core::IReferenceCounted
             return ctx.params.flags;
         }
 
+        //! It allows you to get writer flags which specify some information on saving process like handedness coordinate system
+        inline virtual IAssetWriter::E_WRITER_PARAMETER_FLAGS getAssetWritingParameterFlags(const SAssetWriteContext& ctx, const IAsset* assetToWrite, const uint32_t& hierarchyLevel)
+        {
+            return ctx.params.writerFlags;
+        }
```
devshgraphicsprogramming: this duplicates the above function I think?

src/irr/asset/CBAWMeshFileLoader.h
```diff
@@ -88,6 +88,12 @@ class CBAWMeshFileLoader : public asset::IAssetLoader
 		unsigned char iv[16];
 	};
 
+    template<typename aType>
+    static inline void performActionBasedOnOrientationSystem(aType& varToHandle, void (*performOnCertainOrientation)(aType& varToHandle))
```
devshgraphicsprogramming: kill this function, we don't use it

src/irr/asset/CBAWMeshFileLoader.cpp
```diff
@@ -192,6 +194,147 @@ SAssetBundle CBAWMeshFileLoader::loadAsset(io::IReadFile* _file, const asset::IA
...
+	const core::matrix3x4SIMD xFlip(
+		-1.f,0.f,0.f,0.f,
+		 0.f,1.f,0.f,0.f,
+		 0.f,0.f,1.f,0.f
+	);
+	while (!params.meshesToFlip.empty())
+	{
+		auto mesh = params.meshesToFlip.top();
+		params.meshesToFlip.pop();
+
```
devshgraphicsprogramming, line 207: Little Optimization: you can skip processing (flipping) the mesh if the reference count `<2` (the stack was the only reference, mesh will get dropped as soon as stack lets go)

PR #422, opened 2020-02-12
Bindings reordering fix

Crisspl:


devshgraphicsprogramming:
whhy does this PR contain updates to submodules?

include/IDriver.h
```diff
@@ -293,6 +294,7 @@ class IDriver : public virtual core::IReferenceCounted, public IVideoCapabilityR
 
 		//! Create a renderpass independent graphics pipeline (@see ICPURenderpassIndependentPipeline)
         virtual core::smart_refctd_ptr<IGPURenderpassIndependentPipeline> createGPURenderpassIndependentPipeline(
+            IGPUPipelineCache* _pipelineCache,
             core::smart_refctd_ptr<IGPURenderpassIndependentPipeline>&& _parent,
```
devshgraphicsprogramming: we can get rid of `parent`

Crisspl: at all? won't it be needed for vulkan implementation?

Crisspl:
>whhy does this PR contain updates to submodules?

idk, added files with git add -u and those got added. I think it's no harm though

include/IDriver.h
devshgraphicsprogramming: nah, IHVs says that they don't care about parents at all.

devshgraphicsprogramming:
> idk, added files with git add -u and those got added. I think it's no harm though

No way to tell unless you do a clean checkout and see if you've added a problem to our already problem riddled submodules.

The submodules were at specific commits and tags for a reaason

include/IDriver.h
Crisspl: what are IHVs?

include/IDriver.h
```diff
@@ -42,6 +42,7 @@ class IDriver : public virtual core::IReferenceCounted, public IVideoCapabilityR
     protected:
 		core::smart_refctd_ptr<StreamingTransientDataBufferMT<> > defaultDownloadBuffer;
 		core::smart_refctd_ptr<StreamingTransientDataBufferMT<> > defaultUploadBuffer;
+        core::smart_refctd_ptr<IGPUPipelineCache> defaultPipelineCache;
```
devshgraphicsprogramming: don't make a default, let the user create one and pass it around.

include/irr/video/IGPUObjectFromAssetConverter.h
```diff
@@ -576,7 +576,7 @@ inline created_gpu_object_array<asset::ICPURenderpassIndependentPipeline> IGPUOb
                 shaders[local_shdr_count++] = (*gpuShaders)[shdrRedirs[shdrIter++]].get();
 
         (*res)[i] = m_driver->createGPURenderpassIndependentPipeline(
-            nullptr, //make and use some driver's default pipeline cache
+            m_driver->getDefaultPipelineCache(),
```
devshgraphicsprogramming: make it a parameter to the functions (also make a parameter struct to these functions of `IGPUObjectFromAssetConverter` for posteriority)

source/Irrlicht/CNullDriver.cpp
```diff
@@ -134,11 +135,11 @@ bool CNullDriver::endScene()
...
 void CNullDriver::bindDescriptorSets_generic(const IGPUPipelineLayout* _newLayout, uint32_t _first, uint32_t _count, const IGPUDescriptorSet* const* _descSets, const IGPUPipelineLayout** _destPplnLayouts)
 {
-    uint32_t compatibilityLimits[IGPUPipelineLayout::DESCRIPTOR_SET_COUNT]{}; //actually more like "compatibility limit + 1" (i.e. 0 mean not comaptible at all)
+    int32_t compatibilityLimits[IGPUPipelineLayout::DESCRIPTOR_SET_COUNT]{};
     for (uint32_t i=0u; i<IGPUPipelineLayout::DESCRIPTOR_SET_COUNT; i++)
     {
-        const uint32_t lim = _destPplnLayouts[i] ? //if no descriptor set bound at this index
-            _destPplnLayouts[i]->isCompatibleUpToSet(IGPUPipelineLayout::DESCRIPTOR_SET_COUNT-1u, _newLayout) : 0u;
+        const int32_t lim = _destPplnLayouts[i] ? //if no descriptor set bound at this index
+            _destPplnLayouts[i]->isCompatibleUpToSet(IGPUPipelineLayout::DESCRIPTOR_SET_COUNT-1u, _newLayout) : -1;
```
devshgraphicsprogramming, line 141: doesnt this change affect a lot more?

src/irr/video/COpenGLPipelineCache.h
```diff
@@ -48,13 +49,25 @@ class COpenGLPipelineCache final : public IGPUPipelineCache
 		return {0,nullptr};
 	}
 	//assumes that m_cache does not already contain an item with key==_key and layout fully compatible with _val.layout
-	void insert(SCacheKey&& _key, SCacheVal&& _val)
+	void insert(SCacheKey&& _key, SCacheVal&& _val, const asset::ICPUBuffer* _spirv)
```
devshgraphicsprogramming: you're caching the wrong thing here

src/irr/video/COpenGLSpecializedShader.h
```diff
@@ -43,6 +43,7 @@ class COpenGLSpecializedShader : public core::impl::ResolveAlignment<IGPUSpecial
 
 		const SInfo& getSpecializationInfo() const { return m_specInfo; }
 		const std::array<uint64_t, 4>& getSpirvHash() const { return m_spirvHash; }
+		const asset::ICPUBuffer* getSpirv() const { return m_spirv.get(); }
```
devshgraphicsprogramming, line 44: wrong thing to cache

src/irr/video/COpenGLSpecializedShader.cpp
```diff
@@ -272,6 +272,7 @@ void COpenGLSpecializedShader::setUniformsImitatingPushConstants(const uint8_t*
 auto COpenGLSpecializedShader::compile(const COpenGLPipelineLayout* _layout) const -> std::pair<GLuint, SProgramBinary>
 {
 	spirv_cross::CompilerGLSL comp(reinterpret_cast<const uint32_t*>(m_spirv->getPointer()), m_spirv->getSize()/4ull);
```
devshgraphicsprogramming: can you cache this compiler?

include/irr/video/IGPUPipelineCache.h
```diff
@@ -13,6 +14,8 @@ class IGPUPipelineCache : public core::IReferenceCounted
 
 public:
 	virtual void merge(uint32_t _count, const IGPUPipelineCache** _srcCaches) = 0;
```
devshgraphicsprogramming, line 16: does Vulkan have this? If not then only CPU cache should have.

include/IDriver.h
devshgraphicsprogramming: independent hardware vendors

source/Irrlicht/CNullDriver.cpp
Crisspl: it does but I changed everything

src/irr/video/COpenGLPipelineCache.h
Crisspl: hm?

include/irr/video/IGPUPipelineCache.h
Crisspl: yea, it has

src/irr/video/COpenGLPipelineCache.h
devshgraphicsprogramming: we shouldn't be caching the SPIR-V shader, just the hash of the shader instead.

Crisspl: yea, yea, ofc it does. This last param is just wrong approach to something else, doesnt matter, sorry. Will rework it today

devshgraphicsprogramming: ok.

src/irr/video/COpenGLSpecializedShader.h
```diff
@@ -52,22 +55,21 @@ class COpenGLSpecializedShader : public core::impl::ResolveAlignment<IGPUSpecial
...
-			SUniform(const SMember& _m, GLint _loc, uint8_t* _valptr) : m(_m), location(_loc), value(_valptr) {}
-			SMember m;
-			GLint location;
-			uint8_t* value;
-		};
-		mutable bool m_uniformsSetForTheVeryFirstTime = true;
-		alignas(128) uint8_t m_uniformValues[IGPUMeshBuffer::MAX_PUSH_CONSTANT_BYTESIZE];
+		SInfo m_specInfo;
+		core::smart_refctd_ptr<const asset::ICPUBuffer> m_spirv;
+		std::array<uint64_t, 4> m_spirvHash;
```
devshgraphicsprogramming, line 58: shouldn't hash be in `COpenGLShader` the unspecialized one?

devshgraphicsprogramming: also the SPIR-V?

src/irr/video/COpenGLSpecializedShader.h
```diff
@@ -19,28 +20,30 @@ namespace video
...
+		};
 
-	public:
 		COpenGLSpecializedShader(size_t _ctxCount, uint32_t _ctxID, uint32_t _GLSLversion, const asset::ICPUBuffer* _spirv, const asset::ISpecializedShader::SInfo& _specInfo, const asset::CIntrospectionData* _introspection);
 
-		inline GLuint getGLnameForCtx(uint32_t _ctxID) const
-		{
-			if ((*m_GLnames)[_ctxID])
-				return (*m_GLnames)[_ctxID];
+		void setUniformsImitatingPushConstants(const uint8_t* _pcData, GLuint _GLname, const GLint* _locations, uint8_t* _state) const;
```
devshgraphicsprogramming: that can move to the pipeline since that will hold program objects now.

src/irr/video/COpenGLSpecializedShader.h
Crisspl: no, why? unspecialized shader has nothing to do with pipelines + unspecialized shader holds spirv OR glsl, while specialized always holds spirv

devshgraphicsprogramming: > unspecialized shader holds spirv OR glsl

hash of either is fine.

Are there any mutations that get applied to the GLSL after we create an `IGPUShader` ?

source/Irrlicht/COpenGLDriver.cpp
```diff
@@ -2239,7 +2278,12 @@ void COpenGLDriver::SAuxContext::flushStateGraphics(uint32_t stateBits)
 			const COpenGLSpecializedShader* shdr = static_cast<const COpenGLSpecializedShader*>(currentState.pipeline.graphics.pipeline->getShaderAtIndex(i));
 			//assert(shdr); //this would be weird
 
-			shdr->setUniformsImitatingPushConstants(pushConstantsState[EPBP_GRAPHICS].data, this->ID);
+            //WARNING order of calls getShaderGLnameForCtx() and getUniformLocationsForStage() is important!!
```
devshgraphicsprogramming: why?

src/irr/video/COpenGLRenderpassIndependentPipeline.h
```diff
@@ -14,21 +14,35 @@ namespace video
...
     COpenGLRenderpassIndependentPipeline(
         core::smart_refctd_ptr<IGPURenderpassIndependentPipeline>&& _parent,
         core::smart_refctd_ptr<IGPUPipelineLayout>&& _layout,
         IGPUSpecializedShader** _shadersBegin, IGPUSpecializedShader** _shadersEnd,
         const asset::SVertexInputParams& _vertexInputParams,
         const asset::SBlendParams& _blendParams,
         const asset::SPrimitiveAssemblyParams& _primAsmParams,
-        const asset::SRasterizationParams& _rasterParams
+        const asset::SRasterizationParams& _rasterParams,
+        uint32_t _ctxCount, uint32_t _ctxID, GLuint _GLnames[SHADER_STAGE_COUNT], COpenGLSpecializedShader::SProgramBinary _binaries[SHADER_STAGE_COUNT]
```
devshgraphicsprogramming: `const` the `_GLnames`

src/irr/video/COpenGLRenderpassIndependentPipeline.h
```diff
@@ -53,9 +67,57 @@ class COpenGLRenderpassIndependentPipeline : public IGPURenderpassIndependentPip
...
+        const COpenGLSpecializedShader* glshdr = static_cast<const COpenGLSpecializedShader*>(m_shaders[_stageIx].get());
+        auto uniforms_rng = glshdr->getUniforms();
+        GLuint GLname = (*m_GLnames)[_ctxID*SHADER_STAGE_COUNT + _stageIx];
+
+        m_uniformLocations[_stageIx] = core::make_refctd_dynamic_array<decltype(m_uniformLocations[_stageIx])>(uniforms_rng.length());
+        for (size_t i = 0ull; i < uniforms_rng.length(); ++i)
+            (*m_uniformLocations[_stageIx])[i] = COpenGLExtensionHandler::extGlGetUniformLocation(GLname, uniforms_rng.begin()[i].m.name.c_str());
+
+        return m_uniformLocations[_stageIx]->data();
+    }
```
devshgraphicsprogramming: because you're using separable programs, the uniform locations can be precomputed (will be identical for all usages across all threads and pipelines) for a specialized shader (from a single compilation) and "cached" inside of the specialized shader

src/irr/video/COpenGLPipelineCache.h
```diff
@@ -0,0 +1,62 @@
...
+
+namespace irr { namespace video
+{
+
+class COpenGLPipelineCache final : public IGPUPipelineCache
+{
+public:
+	struct SCacheVal {
+		COpenGLSpecializedShader::SProgramBinary binary;
+		core::smart_refctd_ptr<COpenGLPipelineLayout> layout;
```
devshgraphicsprogramming: shouldn't layout be the key? just like specialization would be?

src/irr/video/COpenGLPipelineCache.h
```diff
@@ -0,0 +1,62 @@
...
+
+	COpenGLSpecializedShader::SProgramBinary find(const SCacheKey& _key, const COpenGLPipelineLayout* _layout) const
+	{
+		auto rng = m_cache.equal_range(_key);
+		for (auto it = rng.first; it != rng.second; ++it)
+			if (_layout->isCompatibleUpToSet(COpenGLPipelineLayout::DESCRIPTOR_SET_COUNT-1u, it->second.layout.get())==(COpenGLPipelineLayout::DESCRIPTOR_SET_COUNT-1u))
+				return it->second.binary;
+		return {0,nullptr};
+	}
+	//assumes that m_cache does not already contain an item with key==_key and layout fully compatible with _val.layout
```
devshgraphicsprogramming, line 69: in `IRR_DEBUG` assert that

src/irr/video/COpenGLSpecializedShader.h
Crisspl: well, yea, like everything. #include resolvements, binding reorders, push constants to uniform translation, specialization constants resolvements

src/irr/video/COpenGLPipelineCache.h
Crisspl: no because there's no way to make an operator< for it

src/irr/video/COpenGLSpecializedShader.h
devshgraphicsprogramming: > binding reorders, push constants to uniform translation, specialization constants resolvement

That's on the GLSL generated from SPIR-V.

But yeah I remember the run-time extension `include`s

source/Irrlicht/COpenGLDriver.cpp
```diff
@@ -1323,14 +1323,45 @@ core::smart_refctd_ptr<IGPURenderpassIndependentPipeline> COpenGLDriver::createG
...
+
+    GLuint GLname = 0u;
+    COpenGLSpecializedShader::SProgramBinary binary;
+    if (_pipelineCache)
+    {
+        COpenGLPipelineCache* cache = static_cast<COpenGLPipelineCache*>(_pipelineCache);
+        COpenGLPipelineLayout* layout = static_cast<COpenGLPipelineLayout*>(_layout.get());
+        COpenGLSpecializedShader* glshdr = static_cast<COpenGLSpecializedShader*>(_shader.get());
+
+        COpenGLPipelineCache::SCacheKey key{ glshdr->getSpirvHash(), glshdr->getSpecializationInfo() };
```
devshgraphicsprogramming: layout should be in the key though? since we want to get the full driver dependent shader binary (and that depends on GLSL generated from SPIR-V which depends on the layout)

source/Irrlicht/COpenGLDriver.cpp
```diff
@@ -2282,8 +2313,11 @@ void COpenGLDriver::SAuxContext::flushStateGraphics(uint32_t stateBits)
             GLuint GLname = currentState.pipeline.graphics.pipeline->getShaderGLnameForCtx(i, this->ID);
             const GLint* locations = currentState.pipeline.graphics.pipeline->getUniformLocationsForStage(i, this->ID);
             uint8_t* PCstate = currentState.pipeline.graphics.pipeline->getPushConstantsStateForStage(i, this->ID);
+            const bool uniformsEverSet = currentState.pipeline.graphics.pipeline->haveUniformsBeenEverSet(i, this->ID);
 
-			shdr->setUniformsImitatingPushConstants(pushConstantsState[EPBP_GRAPHICS].data, GLname, locations, PCstate);
+			shdr->setUniformsImitatingPushConstants(pushConstantsState[EPBP_GRAPHICS].data, GLname, locations, PCstate, !uniformsEverSet);
+
+            currentState.pipeline.graphics.pipeline->afterUniformsSet(i, this->ID);
```
devshgraphicsprogramming: dont put this in the hot loop (stuttering) actually intialize at the start.

source/Irrlicht/COpenGLDriver.cpp
```diff
@@ -2393,7 +2427,12 @@ void COpenGLDriver::SAuxContext::flushStateCompute(uint32_t stateBits)
 		{
 			const COpenGLSpecializedShader* shdr = static_cast<const COpenGLSpecializedShader*>(currentState.pipeline.compute.pipeline->getShader());
 
-			shdr->setUniformsImitatingPushConstants(pushConstantsState[EPBP_COMPUTE].data, this->ID);
+            GLuint GLname = currentState.pipeline.compute.pipeline->getShaderGLnameForCtx(0u, this->ID);
+            const GLint* locations = currentState.pipeline.compute.pipeline->getUniformLocationsForStage(this->ID);
+            uint8_t* PCstate = currentState.pipeline.compute.pipeline->getPushConstantsStateForStage(0u, this->ID);
+            const bool uniformsEverSet = currentState.pipeline.compute.pipeline->haveUniformsBeenEverSet(0u, this->ID);
+
+			shdr->setUniformsImitatingPushConstants(pushConstantsState[EPBP_COMPUTE].data, GLname, locations, PCstate, !uniformsEverSet);
```
devshgraphicsprogramming: same as https://github.com/buildaworldnet/IrrlichtBAW/pull/422/commits/98fcf894ea6ffb42cdbe37ba98615dab9fe4e6f8#r379923220

source/Irrlicht/COpenGLDriver.cpp
Crisspl: look in code that actually finds binaries

src/irr/video/COpenGLPipelineCache.h
devshgraphicsprogramming: let's invent one (don't actually give it a `operator<` but use a `map` with a custom comparator functor struct.

You can treat a layout as a 4d vector/array of descriptor layouts, so implement a `<` operator like you would for a 4bit LSB int (first set `<`, then if first set `==` then check second set, and so on)

Then a `<` function for descriptor layouts could be defined the same as you define a `<` operator for an `std::array`

src/irr/video/COpenGLPipelineCache.h
```diff
@@ -33,19 +34,31 @@ class COpenGLPipelineCache final : public IGPUPipelineCache
...
+			sz = std::max(sz, static_cast<const COpenGLPipelineCache*>(_srcCaches[i])->m_cache.getSize());
+
+		auto buf = core::make_refctd_dynamic_array<core::smart_refctd_dynamic_array<decltype(m_cache)::MutablePairType>>(sz);
 		for (uint32_t i = 0u; i < _count; ++i)
 		{
 			const auto& src = static_cast<const COpenGLPipelineCache*>(_srcCaches[i])->m_cache;
-			m_cache.insert(src.begin(), src.end());
+			src.outputAll(sz, buf->data());
+			for (size_t j = 0ull; j < sz; ++j)
+				m_cache.insert((*buf)[j].first, (*buf)[j].second);
```
devshgraphicsprogramming: use a single-threaded cache, and mutex the shit out of the pipeline, just using a thread-safe cache won't make the member functions of the owner safe to call from multiple threads.

src/irr/video/COpenGLPipelineCache.h
devshgraphicsprogramming: its possible to implement a comparator using very similar logic to this:
https://github.com/buildaworldnet/IrrlichtBAW/pull/422/files#diff-73a7b1e23025a75063836a1f6c71770aR80

Crisspl: so u're saying that having let's say layouts A, B,C and knowing relation of compatibility between them I can sort them in some order that will optimize process of searching the one compatible with some another layout D? I don't think so, how would it work? Compatibility between layouts A and B says nothing about compatibility A and D

Crisspl: >its possible to implement a comparator using very similar logic to this:

u dont want just a comparator, u want a clearly defined relation

src/irr/video/COpenGLPipelineCache.h
Crisspl: > just using a thread-safe cache won't make the member functions of the owner safe to call from multiple threads.

why?

Crisspl: hm, ok, i know what u mean, nvm

src/irr/video/COpenGLRenderpassIndependentPipeline.h
Crisspl: they're precomputed once for all threads, however u're right that those can be stored in COpenGLSpecializedShader, i'll change it

source/Irrlicht/COpenGLDriver.cpp
Crisspl: because `getShaderGLnameForCtx` might create the GL name and `getUniformLocationsForStage` will need GL name to query locations (ofc locations are cached once queried, but order of those calls are still important for the first time when locations are yet unknown)

src/irr/video/COpenGLSpecializedShader.h
Crisspl: not really because COpenGLSpecializedShader holds offsets of "uniforms"

src/irr/video/COpenGLSpecializedShader.h
```diff
@@ -61,6 +61,7 @@ class COpenGLSpecializedShader : public core::impl::ResolveAlignment<IGPUSpecial
 		//mutable bool m_uniformsSetForTheVeryFirstTime = true;
 		//alignas(128) uint8_t m_uniformValues[IGPUMeshBuffer::MAX_PUSH_CONSTANT_BYTESIZE];
 		core::vector<SUniform> m_uniformsList;
+		mutable core::smart_refctd_dynamic_array<GLint> m_locations;
```
devshgraphicsprogramming: either use `vector` for both or `smart_Refctd_dynamic_array`

src/irr/video/COpenGLPipelineCache.h
devshgraphicsprogramming: > so u're saying that having let's say layouts A, B,C and knowing relation of compatibility between them I can sort them in some order that will optimize process of searching the one compatible with some another layout D? I don't think so, how would it work? 

Well in order for a SPIR-V to give you the same GLSL shader, the layout needs to be FULLY compatible, which means that all descriptor sets need to be identically defined.

Fun fact, how STL implements all algorithms in terms of just `<`:
A<B: `A<B`
A>B: `B<A`
A!=B: `A<B || B<A`
A==B: `!(A!=B)`

So we don't need a relation that makes sense, just one that is consistent (no `A<B<C && C<A` loop) and will guarantee that equality will work.

> Compatibility between layouts A and B says nothing about compatibility A and D
> u dont want just a comparator, u want a clearly defined relation

Well compatibility for a SHADER binding reordering requires full compatibility, so if A and B are compatible for shader X and C is compatible with A or B up to set 4, then C will be compatible as the descriptor set layouts are identically defined.

P.S. Another way to establish a `<` relation is to first compare shader hash, then specialization info, and then the computed (set,binding) to (type,binding) remappings which could be cached in `COpenGLPipelineLayout` similarily to precomputed multibind params 
https://github.com/Crisspl/IrrlichtBAW/blob/shaders/src/irr/video/COpenGLPipelineLayout.h#L73

source/Irrlicht/COpenGLDriver.cpp
devshgraphicsprogramming: > because getShaderGLnameForCtx might create the GL name

We could just create all the thread's programs (all the duplicates) at once during pipeline creation so we never, ever, stutter during a frame.

> getUniformLocationsForStage will need GL name to query locations (ofc locations are cached once queried, but order of those calls are still important for the first time when locations are yet unknown)

we should always compile/create at least one shader program upon pipeline creation and hence the uniform locations should be queried at that time and stored in the specialized shader (they will be identical for all following copies).

To avoid any hitches during rendering.

src/irr/video/COpenGLSpecializedShader.h
devshgraphicsprogramming: ok

source/Irrlicht/COpenGLDriver.cpp
Crisspl: done

Crisspl: actually only the 2nd one
So u want to create for all threads at once? Now pipeline irrbaw objects store binaries and create GL names from it when a thread wants to use the pipeline

devshgraphicsprogramming: there's probably an overhead in creating a new shader even from a binary, so maybe doing all up-front is a better idea.

src/irr/video/COpenGLSpecializedShader.cpp
```diff
@@ -291,6 +288,9 @@ auto COpenGLSpecializedShader::compile(const COpenGLPipelineLayout* _layout) con
 	if (logbuf[0])
 		os::Printer::log(logbuf, ELL_ERROR);
 
+	if (m_locations.empty())
```
devshgraphicsprogramming, line 199: what if there are no uniforms at all? will we keep on hitting this?

devshgraphicsprogramming: ah nvm its in `compile` so only one hit ever.

src/irr/video/IOpenGLPipeline.h
```diff
@@ -0,0 +1,91 @@
...
+        m_uniformValues = reinterpret_cast<uint8_t*>(_IRR_ALIGNED_MALLOC(uVals_sz, 128));
+    }
+    ~IOpenGLPipeline()
+    {
+        //shader programs can be shared so all names can be freed by any thread
+        for (GLuint n : *m_GLnames)
+            if (n != 0u)
+                COpenGLExtensionHandler::extGlDeleteProgram(n);
+        _IRR_ALIGNED_FREE(m_uniformValues);
+    }
```
devshgraphicsprogramming: you're leaking binaries?

Crisspl: no, binaries are stored in smart array. It's in a raw one because of alignment

src/irr/video/IOpenGLPipeline.h
```diff
@@ -0,0 +1,91 @@
...
+        (*m_uniformsSetForTheVeryFirstTime)[ix] = false;
+    }
+
+    uint8_t* getPushConstantsStateForStage(uint32_t _stageIx, uint32_t _ctxID) const { return const_cast<uint8_t*>(m_uniformValues + ((_STAGE_COUNT*_ctxID + _stageIx)*IGPUMeshBuffer::MAX_PUSH_CONSTANT_BYTESIZE)); }
+
+    static core::smart_refctd_dynamic_array<GLint> gatherUniformLocations(const COpenGLSpecializedShader* _shader, GLuint _GLname)
+    {
+        auto uniforms_rng = _shader->getUniforms();
+
+        auto uniformLocations = core::make_refctd_dynamic_array<core::smart_refctd_dynamic_array<GLint>>(uniforms_rng.length());
```
devshgraphicsprogramming: do we always need to make a new array?

src/irr/video/IOpenGLPipeline.h
```diff
@@ -0,0 +1,91 @@
...
+            return GLname;
+        }
+
+        return (*m_GLnames)[name_ix];
+    }
+
+    //mutable for deferred GL objects creation
+    mutable core::smart_refctd_dynamic_array<GLuint> m_GLnames;
+    mutable core::smart_refctd_dynamic_array<COpenGLSpecializedShader::SProgramBinary> m_shaderBinaries;
+    mutable core::smart_refctd_dynamic_array<bool> m_uniformsSetForTheVeryFirstTime;
```
devshgraphicsprogramming: make these `std::array`, there's no need for dynamic allocation for arrays we 100% know the upper bound for.

Crisspl: so what's the upper bound?

src/irr/video/COpenGLSpecializedShader.cpp
Crisspl: not ever, every time pipeline with this shader is created. Yes, we hit every time then. And then it runs 0 iterations loop so whatever

src/irr/video/IOpenGLPipeline.h
devshgraphicsprogramming: argh shit `m_GLnames` and `m_uniformsSetForTheVeryFirstTime` actually needs a dynamic allocation (but you could join the `GLuint` and `bool` into a single struct/pair and only have one dynamically allocated array).

But `m_shaderBinaries` you only need up to MAX_STAGES_PER_PIPELINE (which is your template paramater right now).

src/irr/video/COpenGLSpecializedShader.cpp
devshgraphicsprogramming: yeah, I realized

src/irr/video/IOpenGLPipeline.h
Crisspl: not always, this code runs just once. Anyway it's no longer there, now locations are stored in COpenGLSpecializedShader

source/Irrlicht/COpenGLDriver.cpp
Crisspl: i dont know what you mean actually

devshgraphicsprogramming: what does `haveUniformsBeenEverSet` and `afterUniformsSet` do?

src/irr/video/COpenGLSpecializedShader.cpp
devshgraphicsprogramming: what did you end-up doing in the end?

Crisspl: would u prefer it to be cached in COpenGLSPecShader or somehow in pipeline cache?

CMakeLists.txt
```diff
@@ -23,17 +23,17 @@ project(Irrlicht LANGUAGES CXX C)
 # CUDA test
 include(CheckLanguage)
 check_language(CUDA)
-enable_language(CUDA)
```
devshgraphicsprogramming: what error did you get withthis?

devshgraphicsprogramming: https://cmake.org/cmake/help/v3.2/module/CheckLanguage.html

hmm seems I should check for `CMAKE_CUDA_COMPILER` before proceeding to enable language

src/irr/video/COpenGLSpecializedShader.cpp
devshgraphicsprogramming: pipeline cache, but in the second map/cache that's only for IGPUShader

source/Irrlicht/COpenGLDriver.cpp
Crisspl: former one gets flag and the latter one sets it to false

devshgraphicsprogramming: can't the pipeline handle it in `setUniformsImitatingPushConstants` ?

Also why is `setUniformsImitatingPushConstants` a member of the shader!?

src/irr/video/COpenGLPipelineCache.h
```diff
@@ -33,52 +33,75 @@ class COpenGLPipelineCache final : public IGPUPipelineCache
...
 
 		auto rng = m_cache.equal_range(_key);
 		for (auto it = rng.first; it != rng.second; ++it)
 			if (_layout->isCompatibleUpToSet(COpenGLPipelineLayout::DESCRIPTOR_SET_COUNT-1u, it->second.layout.get())==(COpenGLPipelineLayout::DESCRIPTOR_SET_COUNT-1u))
 				return it->second.binary;
 		return {0,nullptr};
 	}
-	//assumes that m_cache does not already contain an item with key==_key and layout fully compatible with _val.layout
-	void insert(SCacheKey&& _key, SCacheVal&& _val, const asset::ICPUBuffer* _spirv)
+	const spirv_cross::ParsedIR* findParsedSpirv(const std::array<uint64_t, 4>& _key)
```
devshgraphicsprogramming, line 58: maybe make the argument an `ICPUShader` then the `ICPUShader` can keep track of its own hash?

src/irr/video/COpenGLPipelineCache.h
```diff
@@ -33,52 +33,75 @@ class COpenGLPipelineCache final : public IGPUPipelineCache
...
+
+		return nullptr;
+	}
+
+	//assumes that m_cache does not already contain an item with key==_key and layout fully compatible with _val.layout
+	void insert(SCacheKey&& _key, SCacheVal&& _val)
+	{
+		const std::lock_guard<std::mutex> _(m_bin_cache_mutex);
 #ifdef _IRR_DEBUG
 		assert(!find(_key, _val.layout.get()).binary);
```
devshgraphicsprogramming: wrong assert, should be `!found || !isCompatible(found.layout.get(),_val.layout.get()`

source/Irrlicht/COpenGLDriver.cpp
Crisspl: because shader holds "uniforms' offsets"

src/irr/video/COpenGLPipelineCache.h
Crisspl: why? pipeline operates on specialized shaders so why even bother with unspecialized ones?

src/irr/video/COpenGLPipelineCache.h
Crisspl: see code of `find()`

CMakeLists.txt
Crisspl: >what error did you get withthis?

"no CUDA toolset found"

Crisspl: ok i fixed it

source/Irrlicht/COpenGLDriver.cpp
devshgraphicsprogramming: it can hold whatever it wants, but since pipelineholds both all the shaders, per-thread shader instances, and these actually hold state in driver's OpenGL implementation... the pipeline should be setting the uniforms with `setUniformsImitatingPushConstants` as a methdod not the shader

src/irr/video/COpenGLPipelineCache.h
devshgraphicsprogramming: but you're findind the parsed spirv of the unspecialized shader right?

src/irr/video/COpenGLPipelineCache.h
devshgraphicsprogramming: `find` assumes that `layout` is compatible with the key?

source/Irrlicht/COpenGLDriver.cpp
devshgraphicsprogramming: ?

devshgraphicsprogramming:
what are we doing about the unresolved conversations?

src/irr/video/COpenGLPipelineCache.h
Crisspl: `find` doesnt find anything (returns nullptr) if passed layout is not compatible with any of entries of equal key

src/irr/video/COpenGLPipelineCache.h
Crisspl: theres no spirv of unspecialized shader and spirv of specialized shader. Theres just spirv. Specialized shader is spirv + spec info

devshgraphicsprogramming: well we can compute the hash of spir-v and GLSL the same anyway, and make ICPUShader keep that.

src/irr/video/COpenGLSpecializedShader.cpp
devshgraphicsprogramming: what happened in the end?

source/Irrlicht/COpenGLDriver.cpp
devshgraphicsprogramming: ?

src/irr/video/COpenGLSpecializedShader.cpp
Crisspl: COpenGLPipelineCache caches parsed spir-v and upon GL shader creation tries to find parsed spir-v to use it for faster spirv-cross compiler construction (in order to specialize, reorder bindings and generate final GLSL)

source/Irrlicht/COpenGLDriver.cpp
Crisspl: done

src/irr/video/COpenGLPipelineCache.cpp
```diff
@@ -3,6 +3,76 @@
...
+		return A ? 1 : 0;
+
+	// non null descriptor set layouts, we can compare more
+	auto count = A->getBindings().length();
+	auto lendiff = int64_t(count) - int64_t(B->getBindings().length());
+	if (lendiff != 0)
+		return lendiff;
+
+	const auto* lhs = A->getBindings().begin();
+	const auto* rhs = B->getBindings().begin();
```
devshgraphicsprogramming, line 21: I'd compare the OpenGL multibind arguments here instead, cause after all you only care if the bindings got remapped the same to OpenGL slots (which means that as you iterate through the `(set,binding)` tuples of both layouts, do all 4 running total counts of UBO, SSBO, Texture and Image stay the same?)

But I guess that can be left as a TODO in the comment, since this seems to be "good enough"

src/irr/asset/CPLYMeshFileLoader.cpp
```diff
@@ -643,7 +643,7 @@ bool CPLYMeshFileLoader::genVertBuffersForMBuffer(asset::ICPUMeshBuffer* _mbuf,
 		const auto currentBitmask = core::getBitmask({ attrib });
 		inputParams.enabledBindingFlags |= currentBitmask;
 		inputParams.enabledAttribFlags |= currentBitmask;
-		inputParams.bindings[attrib] = { asset::getTexelOrBlockBytesize((E_FORMAT)vertexAttribParamsAllOptions[attrib].format), EVIR_PER_INSTANCE };
+		inputParams.bindings[attrib] = { 16, EVIR_PER_VERTEX };
```
devshgraphicsprogramming, line 646: cmon guys, you know I hate constant literals.

PR #423, opened 2020-02-23
CPLY Loader

AnastaZIuk:
Status:

- **refactored**

TODO:

- optimalization while loading, get rid of attribs vector and use SBufferBinding instead
- not sure what about default shaders being used by the loader


AnastaZIuk:
shit I think we should move the PR to Criss's shaders branch

Crisspl:
if your CPLY branch comes from matt's master then yea probably :P because my shaders branch is already merged with matt/master

AnastaZIuk:
moving to https://github.com/Crisspl/IrrlichtBAW/pull/26

PR #426, opened 2020-03-03
small fix

AnastaZIuk:


devshgraphicsprogramming:
whats the reason for the change?

AnastaZIuk:
We had 2 enums and you had told me that we don't want to divide flags, so righthanded flag should be moved and we didn't remove the second one enum

PR #432, opened 2020-03-11
Normal Recalculation Partial-Fix

devshgraphicsprogramming:


devshgraphicsprogramming:
mergining provisionally because I want to work on it.

PR #437, opened 2020-03-14
Normal Quantization Cache

Przemog1:


Przemog1:
hash function may not make sense at this moment :V

include/irr/asset/normal_quantization.h
```diff
@@ -15,6 +15,28 @@ namespace irr
 {
 namespace asset
 {
+	struct QuantNormalHash
+	{
+		size_t operator()(const core::vectorSIMDf& vec) const noexcept
+		{
+			static constexpr uint32_t primeNumber1 = 73856093;
+			static constexpr uint32_t primeNumber2 = 19349663;
+			static constexpr uint32_t primeNumber3 = 83492791;
```
devshgraphicsprogramming: if you're hashing in 64bits then you need 64bit long prime numbers (big big numbers)

Also make the hash 2D, cause you'll be storing projections

include/irr/asset/normal_quantization.h
```diff
@@ -15,6 +15,28 @@ namespace irr
...
+
+			return	((static_cast<size_t>(vec.x * std::numeric_limits<uint32_t>::max())* primeNumber1) ^
```
devshgraphicsprogramming: `float(std::numeric_limits<uint64_t>::max())`

include/irr/asset/normal_quantization.h
```diff
@@ -15,6 +15,28 @@ namespace irr
...
+				(static_cast<size_t>(vec.y * std::numeric_limits<uint32_t>::max())* primeNumber2) ^
+				(static_cast<size_t>(vec.z * std::numeric_limits<uint32_t>::max())* primeNumber3))& (std::numeric_limits<size_t>::max());
```
devshgraphicsprogramming: no need for the `&` operation

include/irr/asset/normal_quantization.h
```diff
@@ -58,6 +80,11 @@ namespace asset
 	extern core::vector<QuantizationCacheEntry16_16_16>     normalCacheFor16_16_16Quant;
 	extern core::vector<QuantizationCacheEntryHalfFloat>    normalCacheForHalfFloatQuant;
 
+	extern core::unordered_map<core::vectorSIMDf, uint32_t, QuantNormalHash, QuantNormalEqualTo> normalCacheFor2_10_10_10QuantUm;
+	extern core::unordered_map<core::vectorSIMDf, uint32_t, QuantNormalHash, QuantNormalEqualTo> normalCacheFor8_8_8QuantUm;
+	extern core::unordered_map<core::vectorSIMDf, uint32_t, QuantNormalHash, QuantNormalEqualTo> normalCacheFor16_16_16QuantUm;
+	extern core::unordered_map<core::vectorSIMDf, uint32_t, QuantNormalHash, QuantNormalEqualTo> normalCacheForHalfFloatQuantUm;
```
devshgraphicsprogramming: these must store a `struct {uint16_t[3]};` not a uint32_t

include/irr/asset/normal_quantization.h
```diff
@@ -160,7 +192,7 @@ namespace asset
         uint32_t bestFit = snormVec[0]|(snormVec[1]<<quantizationBits)|(snormVec[2]<<(quantizationBits*2u));
 
 		dummySearchVal.value = bestFit;
-        normalCacheFor2_10_10_10Quant.insert(found,dummySearchVal);
+        normalCacheFor2_10_10_10QuantUm.insert(std::make_pair(found->first, bestFit));
```
devshgraphicsprogramming: why are you inserting into the cache what you've already found? Your key needs to be `projFitting` that you've tried searching with

For the `std::vector::insert` function the iterator is a placement hint.

include/irr/asset/normal_quantization.h
```diff
@@ -144,11 +171,16 @@ namespace asset
...
         dummySearchVal.key = normal;
-        auto found = std::lower_bound(normalCacheFor2_10_10_10Quant.begin(),normalCacheFor2_10_10_10Quant.end(),dummySearchVal);
+        /*auto found = std::lower_bound(normalCacheFor2_10_10_10Quant.begin(),normalCacheFor2_10_10_10Quant.end(),dummySearchVal);
         if (found!=normalCacheFor2_10_10_10Quant.end()&&(found->key==normal).all())
         {
             return found->value;
-        }
+        }*/
+
+		auto found = normalCacheFor2_10_10_10QuantUm.find(normal);
```
devshgraphicsprogramming: First of all use `fittingVector` not `normal`, we save space by noting the symmetry and using a `abs(normal)` for the fitting and caching (after fit is found or computed the function should restore the signs).

Should be applied to the search result too now
https://github.com/buildaworldnet/IrrlichtBAW/pull/437/files#diff-6e17f886f5c5ef84f97cb8acf20c9926R188

you want to take the `fittingVector` and transform it into a 2d surface projection `projFitting`

my suggestion is barycentric coordinate of the intersection of a line originating at (0,0,0) in direction of `fittingVector` with the triangle defined by vertices (0,0,1),(0,1,0),(1,0,0)

CMakeLists.txt
```diff
@@ -20,6 +20,8 @@ cmake_minimum_required(VERSION 3.14)
 
 project(Irrlicht LANGUAGES CXX C)
 
+add_definitions("/D_ENABLE_EXTENDED_ALIGNED_STORAGE")
```
Przemog1: will take care of it when I'm done with cache

include/irr/asset/normal_quantization.h
```diff
@@ -169,13 +189,17 @@ namespace asset
...
+		if (found != normalCacheFor2_10_10_10QuantUm.end() && (found->first == core::abs(normal)).all())
+		{
 
-		if (found != normalCacheFor2_10_10_10QuantUm.end() && (found->first == normal).all())
 			return found->second;
+		}
 
 		constexpr uint32_t quantizationBits = 10u;
-		const auto xorflag = core::vectorSIMDu32((0x1u<<quantizationBits)-1u);
+		const auto xorflag = core::vectorSIMDu32((0x1u<<quantizationBits)-1);
```
Przemog1: oops

CMakeLists.txt
devshgraphicsprogramming: ok, thanks for pointing out!

include/irr/asset/normal_quantization.h
```diff
@@ -10,11 +10,31 @@
...
+
+	inline VectorUV mapToBarycentric(const core::vectorSIMDf& vec)
+	{
+		//normal to A = [1,0,0], B = [0,1,0], C = [0,0,1] triangle
+		static const core::vector3df_SIMD n(0.577f, 0.577f, 0.577f);
+
+		//point of intersection of vec and triangle - ( n dot A ) / (n dot vec) ) * vec
+		const float r = 0.577f / core::dot(n, vec).x;
+		
+		//[0, 1, 0] + u * [1, -1, 0] + v * [0, -1, 1] = P, so u = Px and v = Py
```
devshgraphicsprogramming: update the comment that its actually `v=Pz`

include/irr/asset/normal_quantization.h
```diff
@@ -10,11 +10,31 @@
...
 {
 namespace asset
 {
+	struct VectorUV
+	{
+		float u;
+		float v;
+	};
+
+	inline VectorUV mapToBarycentric(const core::vectorSIMDf& vec)
```
devshgraphicsprogramming: on the whole looks good!

include/irr/asset/normal_quantization.h
devshgraphicsprogramming: what oops?

include/irr/asset/normal_quantization.h
```diff
@@ -30,22 +35,32 @@ namespace asset
...
+	//will remove later
 	struct QuantNormalHash
 	{
 		size_t operator()(const core::vectorSIMDf& vec) const noexcept
 		{
-			static constexpr uint32_t primeNumber1 = 73856093;
-			static constexpr uint32_t primeNumber2 = 19349663;
-			static constexpr uint32_t primeNumber3 = 83492791;
+			static constexpr size_t primeNumber1 = 73856093;
+			static constexpr size_t primeNumber2 = 19349663;
```
devshgraphicsprogramming: bigger primes!

Przemog1:
@devshgraphicsprogramming now on my machine, in debug mode time to load sponza decreased from ~1:30 to ~0:45
since sponza uses only `quantizeNormal2_10_10_10` functions `quantizeNormal888` and `quantizeNormal16_16_16` are not tested yet

include/irr/asset/normal_quantization.h
```diff
@@ -76,8 +76,8 @@ namespace asset
 	{
 		size_t operator()(const VectorUV& vec) const noexcept
 		{
-			static constexpr size_t primeNumber1 = 73856093;
-			static constexpr size_t primeNumber2 = 19349663;
+			static constexpr size_t primeNumber1 = 18446744073709551557;
+			static constexpr size_t primeNumber2 = 4611686018427388273;
```
Przemog1: may be not the best choice, I will try to figure it out

src/irr/asset/CMeshManipulator.cpp
```diff
@@ -24,15 +24,10 @@ namespace asset
...
-core::vector<QuantizationCacheEntry2_10_10_10> normalCacheFor2_10_10_10Quant;
-core::vector<QuantizationCacheEntry8_8_8> normalCacheFor8_8_8Quant;
-core::vector<QuantizationCacheEntry16_16_16> normalCacheFor16_16_16Quant;
-core::vector<QuantizationCacheEntryHalfFloat> normalCacheForHalfFloatQuant;
-
 core::unordered_map<VectorUV, uint32_t, QuantNormalHashUV, QuantNormalEqualTo> normalCacheFor2_10_10_10QuantUm;
-core::unordered_map<core::vectorSIMDf, uint32_t, QuantNormalHash, QuantNormalEqualTo> normalCacheFor8_8_8QuantUm;
-core::unordered_map<core::vectorSIMDf, uint64_t, QuantNormalHash, QuantNormalEqualTo> normalCacheFor16_16_16QuantUm;
-core::unordered_map<core::vectorSIMDf, uint64_t, QuantNormalHash, QuantNormalEqualTo> normalCacheForHalfFloatQuantUm;
+core::unordered_map<VectorUV, Vector16u, QuantNormalHashUV, QuantNormalEqualTo> normalCacheFor8_8_8QuantUm;
```
devshgraphicsprogramming: Vector8u ?

Przemog1: .. right

src/irr/asset/CMeshManipulator.cpp
```diff
@@ -1482,15 +1480,10 @@ bool CMeshManipulator::calcMaxQuantizationError(const SAttribTypeChoice& _srcTyp
         case EF_R16G16_SFLOAT:
         case EF_R16G16B16_SFLOAT:
         case EF_R16G16B16A16_SFLOAT:
```
devshgraphicsprogramming: just remove these cases and let them drop to default.

devshgraphicsprogramming: and remove the enums from here
https://github.com/buildaworldnet/IrrlichtBAW/blob/6b7aeccf2d8030799992b651fcd5546d10f79e48/src/irr/asset/CMeshManipulator.cpp#L1318

devshgraphicsprogramming:
Nice, we just now need to wrap up the cache in a class and allow for saving, loading and mergining caches (so that we can accelerate between runs)

source/Irrlicht/CQuantNormalCache.h
```diff
@@ -0,0 +1,108 @@
...
+
+		return restoredAsInt;
+	}
+
+	core::vectorSIMDf findBestFit(const uint32_t& bits, const core::vectorSIMDf& normal) const;
+
+private:
+	core::unordered_map<VectorUV, uint32_t, QuantNormalHash, QuantNormalEqualTo> normalCacheFor2_10_10_10Quant;
+	core::unordered_map<VectorUV, Vector8u, QuantNormalHash, QuantNormalEqualTo> normalCacheFor8_8_8Quant;
+	core::unordered_map<VectorUV, Vector16u, QuantNormalHash, QuantNormalEqualTo> normalCacheFor16_16_16Quant;
```
Przemog1, line 342: I wonder if I should make these static

src/irr/asset/CMeshManipulator.cpp
```diff
@@ -1534,6 +1534,19 @@ core::smart_refctd_ptr<ICPUBuffer> IMeshManipulator::idxBufferFromTrianglesFanTo
...
 
+void IMeshManipulator::loadNormalQuantCacheFromBuffer(E_QUANT_NORM_CACHE_TYPE type, asset::ICPUBuffer* buffer)
+{
+
+}
+void IMeshManipulator::addCacheContentsFromBuffer(E_QUANT_NORM_CACHE_TYPE type, asset::ICPUBuffer* buffer)
+{
+
+}
+void IMeshManipulator::saveCacheToFile(const std::string& path)
```
devshgraphicsprogramming: export to buffer, make it take an `SBufferBinding<ICPUBuffer>` as input and write to the correct offset in the buffer (return error is not enough space)

src/irr/asset/CMeshManipulator.cpp
```diff
@@ -1534,6 +1534,19 @@ core::smart_refctd_ptr<ICPUBuffer> IMeshManipulator::idxBufferFromTrianglesFanTo
 	return nullptr;
 }
 
+void IMeshManipulator::loadNormalQuantCacheFromBuffer(E_QUANT_NORM_CACHE_TYPE type, asset::ICPUBuffer* buffer)
+{
+
+}
+void IMeshManipulator::addCacheContentsFromBuffer(E_QUANT_NORM_CACHE_TYPE type, asset::ICPUBuffer* buffer)
```
devshgraphicsprogramming: actually take an `SBufferBinding` its because it would be handy to support an offset into the buffer from which to read.

src/irr/asset/CMeshManipulator.cpp
devshgraphicsprogramming: this is still an issue

source/Irrlicht/CQuantNormalCache.h
```diff
@@ -8,69 +8,69 @@ namespace irr
...
-	struct QuantNormalHash
+	inline bool operator()(const VectorUV& lval, const VectorUV& rval) const noexcept
 	{
-		inline size_t operator()(const VectorUV& vec) const noexcept
-		{
-			static constexpr size_t primeNumber1 = 18446744073709551557;
-			static constexpr size_t primeNumber2 = 4611686018427388273;
+		return (lval.u == rval.u && lval.v == rval.v);
+	}
+};
```
devshgraphicsprogramming: keep these structs in `CQuantNormalCache` don't want them to pollute the global namespace

source/Irrlicht/CQuantNormalCache.h
```diff
@@ -8,69 +8,69 @@ namespace irr
...
-		inline bool operator()(const VectorUV& lval, const VectorUV& rval) const noexcept
-		{
-			return (lval.u == rval.u && lval.v == rval.v);
-		}
-	};
-
-	inline VectorUV mapToBarycentric(const core::vectorSIMDf& vec) const
+	static void insertIntoCache2_10_10_10(const core::vectorSIMDf& key, const uint32_t quantizedNormal);
+	static void insertIntoCache8_8_8(const core::vectorSIMDf& key, const uint32_t quantizedNormal);
+	static void insertIntoCache16_16_16(const core::vectorSIMDf& key, const uint32_t quantizedNormal);
```
devshgraphicsprogramming: dont make the caches static, I hate globals!

source/Irrlicht/CQuantNormalCache.h
```diff
@@ -95,12 +95,12 @@ class CQuantNormalCache
...
-	core::vectorSIMDf findBestFit(const uint32_t& bits, const core::vectorSIMDf& normal) const;
+	static core::vectorSIMDf findBestFit(const uint32_t& bits, const core::vectorSIMDf& normal);
 
 private:
-	core::unordered_map<VectorUV, uint32_t, QuantNormalHash, QuantNormalEqualTo> normalCacheFor2_10_10_10Quant;
-	core::unordered_map<VectorUV, Vector8u, QuantNormalHash, QuantNormalEqualTo> normalCacheFor8_8_8Quant;
-	core::unordered_map<VectorUV, Vector16u, QuantNormalHash, QuantNormalEqualTo> normalCacheFor16_16_16Quant;
+	static core::unordered_map<VectorUV, uint32_t, QuantNormalHash, QuantNormalEqualTo> normalCacheFor2_10_10_10Quant;
+	static core::unordered_map<VectorUV, Vector8u, QuantNormalHash, QuantNormalEqualTo> normalCacheFor8_8_8Quant;
+	static core::unordered_map<VectorUV, Vector16u, QuantNormalHash, QuantNormalEqualTo> normalCacheFor16_16_16Quant;
```
devshgraphicsprogramming: dont make the caches static, I hate globals!

source/Irrlicht/CQuantNormalCache.cpp
```diff
@@ -0,0 +1,181 @@
...
+namespace irr
+{
+namespace asset
+{
+
+core::vectorSIMDf CQuantNormalCache::findBestFit(const uint32_t& bits, const core::vectorSIMDf& normal) const
+{
+	core::vectorSIMDf fittingVector = normal;
+	fittingVector.makeSafe3D();
+	fittingVector = abs(fittingVector);
```
devshgraphicsprogramming: there should be a requirement that the input to this function already be a an absolute value vector

devshgraphicsprogramming: because `bestFit` returned will be an abs value vector

source/Irrlicht/CQuantNormalCache.cpp
```diff
@@ -0,0 +1,181 @@
...
+
+}
+void CQuantNormalCache::insertIntoCache8_8_8(const core::vectorSIMDf& key, const uint32_t quantizedNormal)
+{
+
+}
+void CQuantNormalCache::insertIntoCache16_16_16(const core::vectorSIMDf& key, const uint32_t quantizedNormal)
+{
+
+}
```
devshgraphicsprogramming: you could template these functions and turn them into one, by templating on the cache type (enum or stored value type)

source/Irrlicht/CQuantNormalCache.cpp
```diff
@@ -0,0 +1,181 @@
...
+
+	core::vectorSIMDf fit = findBestFit(quantizationBits, normal);
+	auto absIntFit = core::vectorSIMDu32(core::abs(fit)) & xorflag;
+
+	Vector16u bestFit = { absIntFit[0], absIntFit[1], absIntFit[2] };
+
+	normalCacheFor16_16_16Quant.insert(std::make_pair(mapToBarycentric(normal), bestFit));
+
+	return restoreSign<uint64_t>(absIntFit, xorflag, negativeMask, quantizationBits);
+}
```
devshgraphicsprogramming: these functions are practically identical, the only thing that changes is the number of quantization bits, that you could deduce them from the template type of an enum or stored value type

source/Irrlicht/CQuantNormalCache.h
Przemog1: @devshgraphicsprogramming should i make this class singleton then? I don't think there is need for more than one instance of this class

devshgraphicsprogramming: no singletons, no globals, user might want to make bazillion instances of asset manager (and ergo normal caches)

src/irr/asset/CMeshManipulator.cpp
Przemog1: shouldn't it be range instead of single offset?

source/Irrlicht/CQuantNormalCache.cpp
Przemog1: Since `quantizeNormal16_16_16` returns uint64_t this solution would require different return type as well:

```
template <E_QUANT_NORM_CACHE_TYPE Type, typename ReturnType>
ReturnType quantizeNormal(const core::vectorSIMDf& normal);
```

and usage would look smth like this:

```
quantizeNormal<E_QUANT_NORM_CACHE_TYPE::Q_2_10_10_10, uint32_t>(vec) ;
quantizeNormal<E_QUANT_NORM_CACHE_TYPE::Q_8_8_8, uint32_t>(vec) ;
quantizeNormal<E_QUANT_NORM_CACHE_TYPE::Q_16_16_16, uint64_t>(vec) ;
```

It's error prone, and kinda wierd imo, I know current solution violates DRY rule but personally i would leave it as it is..

source/Irrlicht/CQuantNormalCache.cpp
Przemog1: not anymore

Przemog1: not after recent changes

source/Irrlicht/CQuantNormalCache.cpp
devshgraphicsprogramming: Try this

```cpp
template<E_QUANT_NORM_CACHE_TYPE CacheType>
struct vector_for_cache;

template<> struct vector_for_cache<Q_8_8_8>
{
   typedef type uint32_t;
};
template<> struct vector_for_cache<Q_2_10_10_10>
{
   typedef type uint32_t;
};
template<> struct vector_for_cache<Q_16_16_16>
{
   typedef type uint32_t;
};

template<E_QUANT_NORM_CACHE_TYPE CacheType>
using vector_for_cache_t = typename vector_for_cache<CacheType>::type;

template<E_QUANT_NORM_CACHE_TYPE CacheType>
vector_for_cache_t<CacheType> quantizeNormal(const core::vectorSIMDf& normal);
```

include/irr/asset/IMeshManipulator.h
```diff
@@ -368,9 +377,9 @@ namespace asset
 		}
 		
 		//!
-		static void loadNormalQuantCacheFromBuffer(E_QUANT_NORM_CACHE_TYPE type, asset::ICPUBuffer* buffer);
+		static bool loadNormalQuantCacheFromBuffer(E_QUANT_NORM_CACHE_TYPE type, SBufferRange<ICPUBuffer>& buffer, CQuantNormalCache& quantNormalCache);
 
-		//!
+		//! idk if there is need for this function ?\_(?)_/?
```
devshgraphicsprogramming: instead, maybe make a boolean/flag in `loadNormalQuantCacheFromBuffer` whether to replace or merge/add the contents

include/irr/asset/IMeshManipulator.h
```diff
@@ -21,6 +21,15 @@ namespace asset
 	class CQuantNormalCache;
 	enum class E_QUANT_NORM_CACHE_TYPE;
 
+	//I will delete it later (or move somewhere else)
+	template<typename BufferType>
+	struct SBufferRange
```
devshgraphicsprogramming: you can put it next to `SBufferBinding`

src/irr/asset/CMeshManipulator.cpp
```diff
@@ -1530,14 +1530,80 @@ core::smart_refctd_ptr<ICPUBuffer> IMeshManipulator::idxBufferFromTrianglesFanTo
 	return nullptr;
 }
 
-void IMeshManipulator::loadNormalQuantCacheFromBuffer(E_QUANT_NORM_CACHE_TYPE type, asset::ICPUBuffer* buffer)
+bool IMeshManipulator::loadNormalQuantCacheFromBuffer(E_QUANT_NORM_CACHE_TYPE type, SBufferRange<ICPUBuffer>& buffer, CQuantNormalCache& quantNormalCache)
```
devshgraphicsprogramming: this should be a method of the `CQuantNormalCache` not its owner.

src/irr/asset/CMeshManipulator.cpp
devshgraphicsprogramming: still

src/irr/asset/CMeshManipulator.cpp
```diff
@@ -1541,6 +1530,145 @@ core::smart_refctd_ptr<ICPUBuffer> IMeshManipulator::idxBufferFromTrianglesFanTo
...
+    }
+
+    return true;
+}
+void IMeshManipulator::addCacheContentsFromBuffer(E_QUANT_NORM_CACHE_TYPE type, asset::ICPUBuffer* buffer)
+{
+
+}
+
+bool IMeshManipulator::saveCacheToBuffer(E_QUANT_NORM_CACHE_TYPE type, SBufferBinding<ICPUBuffer>& buffer, CQuantNormalCache& quantNormalCache)
```
devshgraphicsprogramming: and these should also be methods of `CQuantNormalCache`

devshgraphicsprogramming:
merge my branch into yours first.... and lets get this MF-cker merged!

include/irr/asset/IAssetLoader.h
```diff
@@ -101,6 +101,9 @@ class IAssetLoader : public virtual core::IReferenceCounted
         const E_CACHING_FLAGS cacheFlags;
         const char* relativeDir;
         const E_LOADER_PARAMETER_FLAGS loaderFlags;				//!< Flags having an impact on extraordinary tasks during loading process
+
+		//smart ptr?
+		CQuantNormalCache& quantNormalCache = IMeshManipulator::getQuantNormalCache();
```
Przemog1: @devshgraphicsprogramming is it ok?

Przemog1: https://github.com/Przemog1/IrrlichtBAW/blob/shader_pipeline/include/irr/asset/IMeshManipulator.h#L374

devshgraphicsprogramming: just plain old pointer and  non-static member function

include/irr/asset/IMeshManipulator.h
```diff
@@ -375,17 +366,12 @@ namespace asset
...
-		//! idk if there is need for this function ?\_(?)_/?
-		static void addCacheContentsFromBuffer(E_QUANT_NORM_CACHE_TYPE type, asset::ICPUBuffer* buffer);
 
-		//!
-		static bool saveCacheToBuffer(E_QUANT_NORM_CACHE_TYPE type, SBufferBinding<ICPUBuffer>& buffer, CQuantNormalCache& quantNormalCache);
+		static inline CQuantNormalCache& getQuantNormalCache() { return quantNormalCache; }
+		static inline const CQuantNormalCache& getQuantNormalcache() { return quantNormalCache; }
 
     protected:
+		static CQuantNormalCache quantNormalCache;
```
devshgraphicsprogramming: STOP WITH THE STATICS!

Przemog1: that will complicate things but ok

source/Irrlicht/CQuantNormalCache.h
```diff
@@ -68,11 +69,114 @@ class CQuantNormalCache
...
-	uint32_t quantizeNormal2_10_10_10(const core::vectorSIMDf& normal);
-	uint32_t quantizeNormal8_8_8(const core::vectorSIMDf& normal);
-	uint64_t quantizeNormal16_16_16(const core::vectorSIMDf& normal);
+		IRR_PSEUDO_IF_CONSTEXPR_BEGIN(CacheType == E_QUANT_NORM_CACHE_TYPE::Q_2_10_10_10)
+		{
+			auto found = normalCacheFor2_10_10_10Quant.find(uvMappedNormal);
+			if (found != normalCacheFor2_10_10_10Quant.end() && (found->first == uvMappedNormal))
+			{
+				const uint32_t absVec = found->second;
+				const auto vec = core::vectorSIMDu32(absVec, absVec >> quantizationBits, absVec >> quantizationBits * 2) & xorflag;
```
devshgraphicsprogramming: careful about an expression like `absVec >> quantizationBits * 2` the bitwise operators have weird precedence, its better to write `absVec >> (quantizationBits * 2)`

source/Irrlicht/CQuantNormalCache.h
```diff
@@ -68,11 +69,114 @@ class CQuantNormalCache
...
+		}
+		IRR_PSEUDO_IF_CONSTEXPR_END
+		IRR_PSEUDO_IF_CONSTEXPR_BEGIN(CacheType == E_QUANT_NORM_CACHE_TYPE::Q_8_8_8)
+		{
+			quantizationBits = 8u;
+		}
+		IRR_PSEUDO_IF_CONSTEXPR_END
+		IRR_PSEUDO_IF_CONSTEXPR_BEGIN(CacheType == E_QUANT_NORM_CACHE_TYPE::Q_16_16_16)
+		{
+			quantizationBits = 10u; //for sure?
```
devshgraphicsprogramming: 16

source/Irrlicht/CQuantNormalCache.h
```diff
@@ -68,11 +69,114 @@ class CQuantNormalCache
...
+		{
+			Vector16u bestFit = { absIntFit[0], absIntFit[1], absIntFit[2] };
+			normalCacheFor16_16_16Quant.insert(std::make_pair(uvMappedNormal, bestFit));
+		}
+		IRR_PSEUDO_IF_CONSTEXPR_END
+
+		return restoreSign<vector_for_cache_t<CacheType>>(absIntFit, xorflag, negativeMask, quantizationBits);
+	}
 
 	void insertIntoCache2_10_10_10(const VectorUV key, const uint32_t quantizedNormal);
```
devshgraphicsprogramming: template those

source/Irrlicht/CQuantNormalCache.h
```diff
@@ -68,11 +69,114 @@ class CQuantNormalCache
...
+			auto found = normalCacheFor16_16_16Quant.find(uvMappedNormal);
+			if (found != normalCacheFor16_16_16Quant.end() && (found->first == uvMappedNormal))
+			{
+				const auto absVec = core::vectorSIMDu32(found->second.x, found->second.y, found->second.z);
+
+				return restoreSign<uint64_t>(absVec, xorflag, negativeMask, quantizationBits);
+			}
+		}
+
+		core::vectorSIMDf fit = findBestFit(quantizationBits, normal);
```
devshgraphicsprogramming: remember about the `abs` on the normal, so the input makes sense (moving it out of the function

src/irr/asset/COBJMeshFileLoader.cpp
```diff
@@ -285,7 +285,7 @@ asset::SAssetBundle COBJMeshFileLoader::loadAsset(io::IReadFile* _file, const as
 					core::vectorSIMDf simdNormal;
 					simdNormal.set(normalsBuffer[Idx[2]].data);
                     simdNormal.makeSafe3D();
-					v.normal32bit = _params.quantNormalCache.quantizeNormal2_10_10_10(simdNormal);
+					v.normal32bit = _params.quantNormalCache.quantizeNormal<E_QUANT_NORM_CACHE_TYPE::Q_2_10_10_10>(simdNormal);
```
devshgraphicsprogramming: remember about the fallback if `_params.quantNormalCache==nullptr` (get the default from assetManager->getMeshManipulator()->getNormalCache())

src/irr/asset/CMeshManipulator.h
```diff
@@ -121,8 +121,6 @@ class CMeshManipulator : public IMeshManipulator
 			}
 			return output;
 		}
-
-		CQuantNormalCache quantNormalCache;
```
devshgraphicsprogramming: you can keep it as a member, just not a static member

devshgraphicsprogramming:
can you merge my `image_manipulator` branch into this?

devshgraphicsprogramming:
Did you merge rebase or something like that??

I'm seeing 209 modified files!?

include/irr/asset/IAssetManager.h
```diff
@@ -257,14 +257,14 @@ class IAssetManager : public core::IReferenceCounted
...
         SAssetBundle getAssetInHierarchy(const std::string& _filePath, const IAssetLoader::SAssetLoadParams& _params, uint32_t _hierarchyLevel, IAssetLoader::IAssetLoaderOverride* _override)
         {
-            IAssetLoader::SAssetLoadContext ctx(_params, nullptr);
-
-            if (_params.quantNormalCache == nullptr)
+            if (_params.meshManipulatorToUse == nullptr)
             {
                 //meh
-                const_cast<IAssetLoader::SAssetLoadParams&>(_params).quantNormalCache = m_meshManipulator->_getQuantNormalCache();
+                const_cast<IAssetLoader::SAssetLoadParams&>(_params).meshManipulatorToUse = m_meshManipulator.get();
```
devshgraphicsprogramming: do not const_cast its dangerous

devshgraphicsprogramming: just create new parameters or something

include/irr/asset/IGeometryCreator.h
```diff
@@ -94,7 +98,8 @@ class IGeometryCreator : public core::IReferenceCounted
 		\param oblique (to be documented)
 		\return Generated mesh.
 		*/
-		virtual return_type createConeMesh(float radius, float length, uint32_t tesselation,
+		virtual return_type createConeMesh(IMeshManipulator* const meshManipulatorToUse,
```
devshgraphicsprogramming: make `IMeshManipulator* const meshManipulatorToUse` last parameter (as `IMeshManipulator* meshManipulatorOverride=nullptr`)

src/irr/asset/CGeometryCreator.cpp
```diff
@@ -178,11 +179,12 @@ CGeometryCreator::return_type CGeometryCreator::createArrowMesh(const uint32_t t
 
 
 /* A sphere with proper normals and texture coords */
-CGeometryCreator::return_type CGeometryCreator::createSphereMesh(float radius, uint32_t polyCountX, uint32_t polyCountY) const
+CGeometryCreator::return_type CGeometryCreator::createSphereMesh(IMeshManipulator* const meshManipulatorToUse, float radius, uint32_t polyCountX, uint32_t polyCountY) const
 {
 	// we are creating the sphere mesh here.
 	return_type retval;
 	constexpr size_t vertexSize = sizeof(CGeometryCreator::SphereVertex);
+	CQuantNormalCache* const quantNormalCache = meshManipulatorToUse->getQuantNormalCache();
```
devshgraphicsprogramming: `(meshManipulatorToUse ? meshManipulatorToUse:assetManager->getMeshManipulator())->getQuantNormalCache()`

src/irr/asset/COBJMeshFileLoader.cpp
```diff
@@ -80,13 +80,15 @@ asset::SAssetBundle COBJMeshFileLoader::loadAsset(io::IReadFile* _file, const as
         _override
     );
 
-	if (_params.quantNormalCache == nullptr)
+	if (_params.meshManipulatorToUse == nullptr)
 	{
-		os::Printer::log("quantNormalCache is nullptr", ELL_ERROR);
+		os::Printer::log("meshManipulatorToUse is nullptr", ELL_ERROR);
```
devshgraphicsprogramming: if its nulll, you just take the default one from the asset manager !

Przemog1: I do it here:
https://github.com/Przemog1/IrrlichtBAW/blob/shader_pipeline/include/irr/asset/IAssetManager.h#L260

devshgraphicsprogramming:
not merging this with so many modified files, branch again, and hand-copy the changes

PR #444, opened 2020-03-24
NonProfit IrrlichtBaW tutorial example

AnastaZIuk:
status:

**done**

It will be covered in wiki section.


devshgraphicsprogramming:
You can remove example 5 and replace it with this... and turn examples 1-4 into tutorials ;)

PR #452, opened 2020-04-02
Make latest branch!

devshgraphicsprogramming:


devshgraphicsprogramming:
@Crisspl if you just fix-up `ITexturePacker` and `ICPUTexturePacker` to use filters then I could merge ;)

PR #473, opened 2020-04-20
Warnings elimination

AnastaZIuk:
status:

- entry warnings amount: **over 6073**
- flags: **-Wall**
- compiler: **msvc**

### DIFF
--- a/include/irr/asset/IAssetManager.h
+++ b/include/irr/asset/IAssetManager.h
@@ -599,7 +599,7 @@ class IAssetManager : public core::IReferenceCounted
             const char** exts = _writer->getAssociatedFileExtensions();
             for (uint32_t i = 0u; i < IAsset::ET_STANDARD_TYPES_COUNT; ++i)
             {
-                const IAsset::E_TYPE type = IAsset::E_TYPE(1u << i);
+                const IAsset::E_TYPE type = IAsset::E_TYPE((uint64_t)1 << (uint64_t)i);
                 if ((suppTypes>>i) & 1u)
                 {
                     m_writers.perType.insert(type, _writer.get());
@@ -618,7 +618,7 @@ class IAssetManager : public core::IReferenceCounted
             {
                 if ((suppTypes >> i) & 1u)
                 {
-                    const IAsset::E_TYPE type = IAsset::E_TYPE(1u << i);
+                    const IAsset::E_TYPE type = IAsset::E_TYPE((uint64_t)1 << (uint64_t)i);
                     m_writers.perType.removeObject(_writer, type);
                     while (const char* ext = exts[extIx++])
                         m_writers.perTypeAndFileExt.removeObject(_writer, {type, ext});
--- a/include/irr/core/math/plane3dSIMD.h
+++ b/include/irr/core/math/plane3dSIMD.h
@@ -98,7 +98,7 @@ class plane3dSIMDf : private vectorSIMDf
 		}
 
 		//!
-        static inline plane3dSIMDf transform(const plane3dSIMDf& _in, const matrix3x4SIMD& _mat)
+        static inline [[maybe_unused]] plane3dSIMDf transform(const plane3dSIMDf& _in, const matrix3x4SIMD& _mat)
         {
             matrix3x4SIMD inv;
             _mat.getInverse(inv);
--- a/include/quaternion.h
+++ b/include/quaternion.h
@@ -359,7 +359,7 @@ inline void quaternion::toEuler(vector3df_SIMD& euler) const
 {
 	vectorSIMDf sqr = *reinterpret_cast<const vectorSIMDf*>(this);
 	sqr *= sqr;
-	const double test = 2.0 * (Y*W - X*Z);
+	const double test = 2.0 * (double)((double)Y * (double)W - (double)X * (double)Z);
 
 	if (core::equals(test, 1.0, 0.000001))
 	{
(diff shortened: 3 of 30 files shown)


include/irr/asset/IAssetManager.h
```diff
@@ -599,7 +599,7 @@ class IAssetManager : public core::IReferenceCounted
             const char** exts = _writer->getAssociatedFileExtensions();
             for (uint32_t i = 0u; i < IAsset::ET_STANDARD_TYPES_COUNT; ++i)
             {
-                const IAsset::E_TYPE type = IAsset::E_TYPE(1u << i);
+                const IAsset::E_TYPE type = IAsset::E_TYPE((uint64_t)1 << (uint64_t)i);
```
devshgraphicsprogramming, line 602: actually if you want a 64bit uint constant the suffix is `ull`, e.g. `1ull`

include/irr/core/math/plane3dSIMD.h
```diff
@@ -98,7 +98,7 @@ class plane3dSIMDf : private vectorSIMDf
 		}
 
 		//!
-        static inline plane3dSIMDf transform(const plane3dSIMDf& _in, const matrix3x4SIMD& _mat)
+        static inline [[maybe_unused]] plane3dSIMDf transform(const plane3dSIMDf& _in, const matrix3x4SIMD& _mat)
```
devshgraphicsprogramming, line 101: i dont think that function declaration return type is the place to put `[[maybe_unused]]`

include/irr/asset/IAssetManager.h
AnastaZIuk: I know those literals, right

include/irr/core/math/plane3dSIMD.h
AnastaZIuk: I didn't solve the problem though, so it will be removed probably

include/quaternion.h
```diff
@@ -382,9 +382,9 @@ inline void quaternion::toEuler(vector3df_SIMD& euler) const
...
 	{
 		// heading = rotation about z-axis
-		euler.Z = (float) atan2(2.0 * (X*Y +Z*W),(sqr.X - sqr.Y - sqr.Z + sqr.W));
+		euler.Z = (float) atan2(2.0f * (X*Y + Z*W),(sqr.X - sqr.Y - sqr.Z + sqr.W));
 		// bank = rotation about x-axis
-		euler.X = (float) atan2(2.0 * (Y*Z +X*W),(-sqr.X - sqr.Y + sqr.Z + sqr.W));
+		euler.X = (float) atan2(2.0f * (Y*Z +X*W),(-sqr.X - sqr.Y + sqr.Z + sqr.W));
 		// attitude = rotation about y-axis
 		euler.Y = (float) asin( core::clamp(test, -1.0, 1.0) );
 	}
```
devshgraphicsprogramming, line 390: actually while you're at it, this function should be changed to use floats throughout

include/irr/core/math/plane3dSIMD.h
AnastaZIuk: but probably should be at def

devshgraphicsprogramming:
Prefer `static_cast<T>(var)` to C-style cast `(T)var`, C-style cast will cast anything to anything pretty much !

careful with `decltype(variable)` especially with regards to ints, and similar, because `decltype(var)` on something that was declared as `T var` actually evaluates to `T&` or `const T&` (depending on current scope's access const qualifiers), and you really dont want to be casting constant literals to references (this ties into why you want to use `static_cast` and not C-style cast).

You'd actually need a std::decay for that... ask @Crisspl how to best handle

include/irr/core/math/plane3dSIMD.h
devshgraphicsprogramming: its probably being unused where its being called.

include/irr/asset/IAssetManager.h
devshgraphicsprogramming: so why the cast?

include/irr/core/math/plane3dSIMD.h
AnastaZIuk: yes, I wanted to use attribute to avoid it

include/irr/asset/IAssetManager.h
AnastaZIuk: to fix

AnastaZIuk:
> Prefer `static_cast<T>(var)` to C-style cast `(T)var`, C-style cast will cast anything to anything pretty much !
> 
> careful with `decltype(variable)` especially with regards to ints, and similar, because `decltype(var)` on something that was declared as `T var` actually evaluates to `T&` or `const T&` (depending on current scope's access const qualifiers), and you really dont want to be casting constant literals to references (this ties into why you want to use `static_cast` and not C-style cast).
> 
> You'd actually need a std::decay for that... ask @Crisspl how to best handle

I need to change for `static_cast`, but I find it probably not true (quite) with evaluating to `T&` or `const T&`, since it depends on the way you call decltype

devshgraphicsprogramming:
> I need to change for static_cast, but I find it probably not true (quite) with evaluating to T& or const T&, since it depends on the way you call decltype

I think `static_cast<decltype()>` should protect us against most things

AnastaZIuk:
> > I need to change for static_cast, but I find it probably not true (quite) with evaluating to T& or const T&, since it depends on the way you call decltype
> 
> I think `static_cast<decltype()>` should protect us against most things

yeah, it's true

include/irr/core/math/plane3dSIMD.h
devshgraphicsprogramming: its the correct place to put it, where its being called and unused, not the global definition for everyone

devshgraphicsprogramming:
Update to match the HEAD of `shader_pipeline` fix my comments and I'll merge

devshgraphicsprogramming:
closing temporarily.... reopen when ready for merge

PR #475, opened 2020-04-22
Vulkan Hello World Test

devshgraphicsprogramming:
I have just two questions:
- How on earth does this work without linking a Vulkan library? i.e. without `vkGetInstanceProcAddr` becoming an undefined reference.
- Does this run and produce the correct "all red window" result?
- Could you port this sample to use SDL2 (you can make SDL2 a submodule in `3rdparty`) instead of glfw, really don't want to add glfw as a dependendency


devshgraphicsprogramming:
I added SDL2 submodule to 3rdparty.

Cmake will need touching up to add it as a dependency of `Irrlicht` target

Here's a list of resources on SDL2 and Vulkan:
https://docs.tizen.org/application/native/guides/graphics/vulkan
https://github.com/Z80Fan/VulkanDemos
https://sopyer.github.io/b/post/minimal-vulkan-sample/
https://www.gamedev.net/forums/topic/699117-vulkan-with-sdl2-getting-started/

devshgraphicsprogramming:
I also included `volk`

https://github.com/zeux/volk

no need to make it a dependency, you can use it in header-only-mode

3rdparty/CMakeLists.txt
```diff
@@ -36,9 +39,61 @@ update_git_submodule(./glm)
 update_git_submodule(./freetype2)
 update_git_submodule(./portable-file-dialogs)
 
-
+# openssl
 add_subdirectory(openssl openssl EXCLUDE_FROM_ALL)
 
+# volk
+if (Vulkan_FOUND)
+	set(VOLK_PULL_IN_VULKAN ON)
```
devshgraphicsprogramming: Did I want this?

devshgraphicsprogramming: oh, you merged my branch

devshgraphicsprogramming:
why did you copy my commit and not merge my branch, this is a recipe for conflicts when merging

PR #476, opened 2020-04-22
VAO tracking issues fixed

Crisspl:


source/Irrlicht/COpenGLDriver.cpp
```diff
@@ -2489,6 +2484,9 @@ void COpenGLDriver::SAuxContext::flushStateGraphics(uint32_t stateBits)
                     }
                 }
             }
+            //upon vao change, force binding all vtx/idx buffers (by setting their current state to null)
+            std::fill(currentState.vertexInputParams.bindings, currentState.vertexInputParams.bindings+IGPUMeshBuffer::MAX_ATTR_BUF_BINDING_COUNT, std::decay_t<decltype(currentState.vertexInputParams.bindings[0])>{});
```
devshgraphicsprogramming: couldn't you cache the VAO's buffer bindings, and fill with the retrieved cached VAO's bindings instead? then the total number of state changes will be lower

devshgraphicsprogramming: N.B. no matter what you decide on the above, you need to do the same on the index buffer binding ;)

source/Irrlicht/COpenGLDriver.cpp
```diff
@@ -2484,14 +2489,12 @@ void COpenGLDriver::SAuxContext::flushStateGraphics(uint32_t stateBits)
                     }
                 }
             }
-            //upon vao change, force binding all vtx/idx buffers (by setting their current state to null)
-            std::fill(currentState.vertexInputParams.bindings, currentState.vertexInputParams.bindings+IGPUMeshBuffer::MAX_ATTR_BUF_BINDING_COUNT, std::decay_t<decltype(currentState.vertexInputParams.bindings[0])>{});
-
             //vertex and index buffer bindings are done outside this if-statement because no change in hash doesn't imply no change in those bindings
         }
         GLuint GLvao = currentState.vertexInputParams.vao.second.GLname;
         if (GLvao)
```
devshgraphicsprogramming: shouldn't you assert that `GLvao` is not 0?

PR #480, opened 2020-04-25
IrrlichtBaW CEGUI

AnastaZIuk:
status:

Well, I launched new version and it looks and behave like **old** one **without**

```C++
GLStateManager.saveOpenGLState();         
setOpenGLClip();
// here it renders GUI
GLStateManager.restoreOpenGLState();
```
[those are here](https://github.com/buildaworldnet/IrrlichtBAW/blob/shader_pipeline/ext/CEGUI/ExtCEGUI.cpp#L119)

the results are as follows

**old** with _commented_ **those 3 above**:

image

**new** with working fonts:

image

**old** with _uncommented_ **those 3 above**:

image

What I don't understand is why it happens something with a state has an impact on half GUI that most is invisible, but we won't use those states and clips probably however. Need to figure out what to do with that.

I'm sure that the widgets are placed in new example correctly, since I can interact with them but I don't see them **due to those states and clip**.

notes:

- I have created a debug example for learning and testing purposes that will be deleted soon
- The strategy is as follows. I will learn CEGUI and CEED, launch alifsko creating a 2D example with OpenGL renderer wrapping everything in my code classes implementation. When it's done - I will change the renderer API to ours.


examples_tests/07a.CEGUIIrrlichtBaWRenderer/bin/CEGUI.log
```diff
@@ -0,0 +1,430 @@
+25/04/2020 22:00:44 (Std) 	+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
```
AnastaZIuk: file to remove

devshgraphicsprogramming:
when the code is commented out the OpenGL vs.Vulkan NDC set-up is wrong (it doesn't get reset to OpenGL default before CEGUI is invoked)

However when its uncommented there's probably something wrong with saving the blend state (`glEnable` + `glBlendFunc` stuff)

ext/CEGUI/CMakeLists.txt
```diff
@@ -1,26 +1,42 @@
...
 	CEGUIOpenGLClip.h
 	Helpers.h
+	IrrlichtBaW/IrrlichtBaWRenderer.hpp
+	IrrlichtBaW/IrrlichtBaWRenderTarget.hpp
+	IrrlichtBaW/IrrlichtBaWRenderQueue.hpp
+	IrrlichtBaW/IrrlichtBaWGeometryBuffer.hpp
+	IrrlichtBaW/IrrlichtBaWRenderingWindow.hpp
+	IrrlichtBaW/IrrlichtBaWRenderEffect.hpp
+	IrrlichtBaW/IrrlichtBaWTextureTarget.hpp
+	IrrlichtBaW/IrrlichtBaWTexture.hpp
```
devshgraphicsprogramming: I wouldn't bother with the extra `IrrlichtBaW` directory since you already prefix the file/class names

ext/CEGUI/IrrlichtBaW/IrrlichtBaWRenderEffect.cpp
```diff
@@ -0,0 +1 @@
+#include "IrrlichtBaWRenderer.hpp"
```
devshgraphicsprogramming: why add `Effect` to the name

ext/CEGUI/IrrlichtBaW/IrrlichtBaWRenderQueue.hpp
```diff
@@ -0,0 +1,25 @@
+#ifndef _IRR_EXT_IRRLICHT_BAW_RENDERER_INCLUDED_
+#define _IRR_EXT_IRRLICHT_BAW_RENDERER_INCLUDED_
```
devshgraphicsprogramming: duplicate file?

ext/CEGUI/IrrlichtBaW/IrrlichtBaWRenderTarget.cpp
```diff
@@ -0,0 +1,48 @@
...
+
+void IrrlichtBaWRenderTarget::activate() 
+{
+
+}
+
+void IrrlichtBaWRenderTarget::deactivate() 
+{
+
+}
```
devshgraphicsprogramming: these will probably be FBO bind and unbind, check if OpenGL backend does more than that

ext/CEGUI/IrrlichtBaW/IrrlichtBaWRenderer.cpp
```diff
@@ -0,0 +1,131 @@
...
+}
+
+const CEGUI::String& IrrlichtBaWRenderer::getIdentifierString() const
+{
+	return d_rendererID;
+}
+
+void IrrlichtBaWRenderer::initializeDescriptorSetLayout()
+{
+	IGPUDescriptorSetLayout::SBinding binding{ 0u, EDT_COMBINED_IMAGE_SAMPLER, 1u, IGPUSpecializedShader::ESS_FRAGMENT, nullptr };
```
devshgraphicsprogramming: you might need to add a UBO or SSBO for some per-quad data such as transforms, colors etc.

ext/CEGUI/IrrlichtBaW/IrrlichtBaWRenderer.hpp
```diff
@@ -0,0 +1,124 @@
...
+		static CEGUI::String d_rendererID;
+
+		/*
+			Standard shaders defined by CEGUI
+		*/
+
+		static constexpr std::string_view ceguiVertexShader = R"===(
+        #version 430 core
+
+        uniform mat4 modelViewPerspMatrix;
```
devshgraphicsprogramming: you want this to come from an SSBO
```glsl
struct PerQuadData
{
   mat4 modelViewPerspMatrix;
};

layout(set=3, binding=1) restrict readonly buffer QuadData
{
   PerQuadData perQuadData[];
};
```

ext/CEGUI/IrrlichtBaW/IrrlichtBaWRenderer.hpp
```diff
@@ -0,0 +1,124 @@
...
+        in vec3 inPosition;
+        in vec2 inTexCoord;
+        in vec4 inColour;
```
devshgraphicsprogramming: check if CEGUI supports gradients, if not (all vertices get assigned the same color) put the color in `PerQuadData`

ext/CEGUI/IrrlichtBaW/IrrlichtBaWRenderer.hpp
```diff
@@ -0,0 +1,124 @@
...
+        }
+		)===";
+
+		static constexpr std::string_view ceguiFragmentShader = R"===(
+        #version 430 core
+        
+        layout(set = 3, binding = 0) uniform sampler2D texture0;
+        in vec2 exTexCoord;
+        in vec4 exColour;
+        out vec4 out0;
```
devshgraphicsprogramming: for GLSL shaders to compile into SPIR-V you need to specify matching `layout(location=` on all `in` and `out` variables

ext/CEGUI/IrrlichtBaW/IrrlichtBaWRenderer.hpp
devshgraphicsprogramming: then naturally a push constant to supply offset into array, so you could do this
```glsl
perQuadData[pcOffset+gl_InstanceIndex];
```

ext/CEGUI/IrrlichtBaW/IrrlichtBaWRenderer.hpp
```diff
@@ -0,0 +1,124 @@
...
+		/*
+			Standard shaders defined by CEGUI
+		*/
+
+		static constexpr std::string_view ceguiVertexShader = R"===(
+        #version 430 core
+
+        uniform mat4 modelViewPerspMatrix;
+        in vec3 inPosition;
+        in vec2 inTexCoord;
```
devshgraphicsprogramming: if everything CEGUI ever draws are quads, then you could easily make these `const` global arrays (like the Vulkan triangle example) for a single quad indexed by `gl_VertexIndex` then if you want more quads in different places, you'd handle that via instancing and a per-quad `modelViewPerspMatrix`

ext/CEGUI/IrrlichtBaW/IrrlichtBaWTexture.hpp
```diff
@@ -0,0 +1,44 @@
...
+
+        virtual void loadFromFile(const CEGUI::String& filename, const CEGUI::String& resourceGroup) override;
+
+        virtual void loadFromMemory(const void* buffer, const CEGUI::Sizef& buffer_size, PixelFormat pixel_format) override;
+
+        virtual void blitFromMemory(const void* sourceData, const CEGUI::Rectf& area) override;
+
+        virtual void blitToMemory(void* targetData) override;
+
+        virtual bool isPixelFormatSupported(const PixelFormat fmt) const override;
```
devshgraphicsprogramming: btw. the final override of a `virtual` function does not need the `virtual` keyword, only `override`

ext/CEGUI/IrrlichtBaW/IrrlichtBaWTexture.cpp
```diff
@@ -0,0 +1,50 @@
...
+
+void IrrlichtBaWTexture::blitToMemory(void* targetData)
+{
+
+}
+
+bool IrrlichtBaWTexture::isPixelFormatSupported(const PixelFormat fmt) const
+{
+
+}
```
devshgraphicsprogramming: these should be ridiculously simple for you to implement ;)

Start here.

ext/CEGUI/IrrlichtBaW/IrrlichtBaWTextureTarget.cpp
```diff
@@ -0,0 +1,25 @@
...
+}
+
+void IrrlichtBaWTextureTarget::declareRenderSize(const CEGUI::Sizef& sz)
+{
+
+}
+
+bool IrrlichtBaWTextureTarget::isRenderingInverted() const
+{
+
```
devshgraphicsprogramming: here do the opposite of what OpenGL says

ext/CEGUI/IrrlichtBaW/IrrlichtBaWRenderTarget.cpp
AnastaZIuk: it sets viewport, the matrix and cegui render target

```C++
template <typename T>
void OpenGLRenderTarget<T>::activate()
{
    glViewport(static_cast<GLsizei>(d_area.left()),
               static_cast<GLsizei>(d_area.top()),
               static_cast<GLsizei>(d_area.getWidth()),
               static_cast<GLsizei>(d_area.getHeight()));

    if (!d_matrixValid)
        updateMatrix();

    d_owner.setViewProjectionMatrix(d_matrix);

    d_owner.setActiveRenderTarget(this);
}

//----------------------------------------------------------------------------//
template <typename T>
void OpenGLRenderTarget<T>::deactivate()
{
}
```

devshgraphicsprogramming: that would be a:
https://github.com/buildaworldnet/IrrlichtBAW/blob/shader_pipeline/include/IVideoDriver.h#L139
https://github.com/buildaworldnet/IrrlichtBAW/blob/shader_pipeline/include/IVideoDriver.h#L112

The matrix stuff you need to compute yourself and send to shader via Push Constant or UBO

ext/CEGUI/IrrlichtBaWRenderTarget.cpp
```diff
@@ -1,30 +1,47 @@
 #include "IrrlichtBaWRenderTarget.hpp"
+#include "IrrlichtBaWGeometryBuffer.hpp"
 
 using namespace irr;
 using namespace ext;
 using namespace cegui;
 
+const double IrrlichtBaWRenderTarget::d_yfov_tan = 0.267949192431123;
```
devshgraphicsprogramming, line 8: a GUI projection is orthographic, not perspective

ext/CEGUI/IrrlichtBaWRenderTarget.hpp
```diff
@@ -34,6 +34,28 @@ class IrrlichtBaWRenderTarget final : public CEGUI::RenderTarget
...
+		//! holds defined area for the RenderTarget
+		CEGUI::Rectf d_area;
+		//! tangent of the y FOV half-angle; used to calculate viewing distance.
+		static const double d_yfov_tan;
+		//! saved copy of projection matrix
+		mutable core::matrix4SIMD d_matrix;
+		//! true if saved matrix is up to date
+		mutable bool d_matrixValid;
+		//! tracks viewing distance (this is set up at the same time as d_matrix)
+		mutable double d_viewDistance;
```
devshgraphicsprogramming, line 58: you need a simple orthographic matrix

ext/CEGUI/IrrlichtBaWTexture.cpp
```diff
@@ -0,0 +1,167 @@
...
+    else
+    {
+        assert(status);
+        return { 0, 0 };
+    }
+}
+
+void IrrlichtBaWTexture::loadFromFile(const CEGUI::String& filename, const CEGUI::String& resourceGroup)
+{
+    auto path = resourceGroup + filename;
```
devshgraphicsprogramming, line 100: oh, so that's how they do it XD

ext/CEGUI/IrrlichtBaWTexture.cpp
```diff
@@ -0,0 +1,167 @@
...
+    auto assetManager = device->getAssetManager();
+
+    auto params = asset::IAssetLoader::SAssetLoadParams();
+    auto bundle = assetManager->getAsset(path, params);
```
devshgraphicsprogramming: you could possibly load with an override that overrides the cache searching so you get a separate asset cache just for CEGUI (you'd also need to override the GPU object creator slightly for `getGPUObjectsFromAssets`)

CC: @Crisspl

ext/CEGUI/IrrlichtBaWTexture.cpp
```diff
@@ -0,0 +1,167 @@
...
+    auto cpuImage = asset::IAsset::castDown<asset::ICPUImage>(bundle.getContents().first[0]);
+    assert(cpuImage.get(), "something went wrong while casting, empty pointer detected!");
+
+    bufferRowLength = cpuImage->getRegions().begin()->bufferRowLength;
```
devshgraphicsprogramming, line 108: what is this needed for?

ext/CEGUI/IrrlichtBaWTexture.cpp
```diff
@@ -0,0 +1,167 @@
...
+
+    asset::ICPUImageView::SCreationParams imageViewInfo;
```
devshgraphicsprogramming, line 110: I'd first create a gpuImage, then create a `IGPUImageView` directly from it, doing it via the CPU object and using `getGPUObjectsFromAssets` introduces more objects into caches

ext/CEGUI/IrrlichtBaWTexture.cpp
```diff
@@ -0,0 +1,167 @@
...
+    imageViewInfo.viewType = asset::IImageView<asset::ICPUImage>::ET_2D;
+    imageViewInfo.flags = static_cast<asset::ICPUImageView::E_CREATE_FLAGS>(0u);
+    imageViewInfo.subresourceRange.baseArrayLayer = 0u;
+    imageViewInfo.subresourceRange.baseMipLevel = 0u;
+    imageViewInfo.subresourceRange.layerCount = imageViewInfo.image->getCreationParameters().arrayLayers;
+    imageViewInfo.subresourceRange.levelCount = imageViewInfo.image->getCreationParameters().mipLevels;
+
+    auto cpuImageView = asset::ICPUImageView::create(std::move(imageViewInfo));
+
+    // cachingName = ""     TODO Okay, so we have to get likely GL id or something similar. Should I cast it to OpenGL image or maybe we should get CPU cache string?
```
devshgraphicsprogramming: what does CEGUI use `cachingName` for ?

ext/CEGUI/IrrlichtBaWTexture.cpp
```diff
@@ -0,0 +1,167 @@
...
+    asset::IImage::SCreationParams imageParams;
+    imageParams.flags = static_cast<asset::IImage::E_CREATE_FLAGS>(0);
+    imageParams.type = asset::IImage::ET_2D;
+    imageParams.format = newFormat;
+    imageParams.extent = { buffer_size.d_width, buffer_size.d_height, 1 };
+    imageParams.mipLevels = 1;
+    imageParams.arrayLayers = 1;
+
+    auto gpuImage = device->getVideoDriver()->createGPUImage(std::move(imageParams));
+    // TODO copy buffer to image
```
devshgraphicsprogramming: call `blitFromMemory(buffer,{0.f,0.f,buffer_size.d_width, buffer_size.d_height})` and do the implementation there

ext/CEGUI/IrrlichtBaWTexture.cpp
```diff
@@ -0,0 +1,167 @@
...
+}
+
+void IrrlichtBaWTexture::blitFromMemory(const void* sourceData, const CEGUI::Rectf& area)
+{
+
+}
+
+void IrrlichtBaWTexture::blitToMemory(void* targetData)
+{
+
```
devshgraphicsprogramming, line 217: this is pretty much what the screenshot extension does

ext/CEGUI/IrrlichtBaWTexture.cpp
```diff
@@ -0,0 +1,167 @@
...
+void IrrlichtBaWTexture::loadFromMemory(const void* buffer, const CEGUI::Sizef& buffer_size, PixelFormat pixel_format)
+{
+    asset::E_FORMAT newFormat = [&]()
+    {
+        switch (pixel_format)
+        {
+            case PixelFormat::PF_RGB: return asset::EF_R8G8B8_SRGB;
+            case PixelFormat::PF_RGBA: return asset::EF_R8G8B8A8_SRGB;
+            case PixelFormat::PF_RGBA_4444: return asset::EF_R4G4B4A4_UNORM_PACK16;
+            case PixelFormat::PF_RGB_565: return asset::EF_R5G6B5_UNORM_PACK16;
```
devshgraphicsprogramming: don't support these, they are not SRGB, and CEGUI probably gets them wrong

ext/CEGUI/IrrlichtBaWTexture.cpp
```diff
@@ -0,0 +1,167 @@
...
+}
+
+void IrrlichtBaWTexture::blitToMemory(void* targetData)
+{
+
+}
+
+bool IrrlichtBaWTexture::isPixelFormatSupported(const PixelFormat fmt) const
+{
+    return true;
```
devshgraphicsprogramming: maybe change the `asset::E_FORMAT newFormat = [&]()` lambda into a static private function, so you can check it here.

ext/CEGUI/IrrlichtBaWTexture.cpp
AnastaZIuk: yeah, you've got an alias to various kind of resources hah

ext/CEGUI/IrrlichtBaWTexture.cpp
```diff
@@ -77,6 +142,9 @@ const CEGUI::Vector2f& IrrlichtBaWTexture::getTexelScaling() const
 
 void IrrlichtBaWTexture::loadFromFile(const CEGUI::String& filename, const CEGUI::String& resourceGroup)
 {
+    if (frameBuffer)
+        frameBuffer->drop();
```
devshgraphicsprogramming: why are you turning every single texture into a framebuffer!?

ext/CEGUI/IrrlichtBaWTexture.cpp
```diff
@@ -127,6 +198,11 @@ void IrrlichtBaWTexture::loadFromMemory(const void* buffer, const CEGUI::Sizef&
         }
     }();
 
+    auto texelOrBlockByteSize = asset::getTexelOrBlockBytesize(newFormat);
+    auto bufferByteSize = buffer_size.d_width * buffer_size.d_height * texelOrBlockByteSize;
+    core::smart_refctd_ptr<asset::ICPUBuffer> cpuPixelBuffer = core::make_smart_refctd_ptr<asset::CCustomAllocatorCPUBuffer<core::null_allocator<uint8_t>>>(bufferByteSize, buffer, core::adopt_memory);
+    core::smart_refctd_ptr<video::IGPUBuffer> gpuPixelBuffer = driver->getGPUObjectsFromAssets(&cpuPixelBuffer.get(), &cpuPixelBuffer.get() + 1u)->front();
```
devshgraphicsprogramming: create a GPU buffer of the given `bufferByteSize` with `driver->createGPUBufferDeviceLocalOnDedMem` and fill it with `driver->updateBufferRangeViaStagingBuffer`, no need for `getGPUObjectsFromAssets` and creation of CPU buffer

ext/CEGUI/IrrlichtBaWTexture.cpp
```diff
@@ -148,17 +237,52 @@ void IrrlichtBaWTexture::loadFromMemory(const void* buffer, const CEGUI::Sizef&
...
+        auto& viewParams = gpuTexture->getCreationParameters();
+        auto& extent = viewParams.image->getCreationParameters().extent;
+        irr::core::recti coverArea(area.left(), area.top(), area.getWidth(), area.getHeight());
+
+        auto temporaryInFrameBuffer = createFrameBuffer(sourceData, viewParams.format, extent.width, extent.height);
+        driver->blitRenderTargets(temporaryInFrameBuffer, frameBuffer, false, false, coverArea, coverArea);
+        temporaryInFrameBuffer->drop();
+    }
+    else
+        assert(status);
```
devshgraphicsprogramming: do not blit with fraebuffers!

you should be using `copyBufferToImage` here, and you could have implemented `IrrlichtBaWTexture::loadFromMemory` with this function partially

ext/CEGUI/IrrlichtBaWTexture.cpp
```diff
@@ -148,17 +237,52 @@ void IrrlichtBaWTexture::loadFromMemory(const void* buffer, const CEGUI::Sizef&
...
+        auto gpuTexture = frameBuffer->getAttachment(textureColorAttachment);
+        auto& viewParams = gpuTexture->getCreationParameters();
+        auto& extent = viewParams.image->getCreationParameters().extent;
+
+        auto temporaryOutFrameBuffer = createFrameBuffer(targetData, viewParams.format, extent.width, extent.height);
+        driver->blitRenderTargets(frameBuffer, temporaryOutFrameBuffer, false, false);
+ 
+        auto bufferByteSize = viewParams.image->getImageDataSizeInBytes();
+        auto destinationBoundMemory = const_cast<video::IDriverMemoryAllocation*>(viewParams.image->getBoundMemory()); // not sure
+        destinationBoundMemory->mapMemoryRange(video::IDriverMemoryAllocation::EMCAF_READ, { 0u, bufferByteSize });
```
devshgraphicsprogramming: that's not how it works and it won't work this way.

devshgraphicsprogramming: you need `copyImageToBuffer`

ext/CEGUI/IrrlichtBaWTexture.cpp
devshgraphicsprogramming: just copy what it does with the downstreaming buffer and `copyImageToBuffer`

ext/CEGUI/IrrlichtBaWTexture.cpp
```diff
@@ -15,35 +16,99 @@ IrrlichtBaWTexture::~IrrlichtBaWTexture()
...
+    return frameBuffer;
+}
+
+irr::video::IFrameBuffer* IrrlichtBaWTexture::createFrameBuffer(core::smart_refctd_ptr<video::IGPUImageView> gpuImageView)
+{
+    auto frameBuffer = driver->addFrameBuffer();
+    frameBuffer->attach(textureColorAttachment, std::move(gpuImageView));
+
+    return frameBuffer;
+}
```
devshgraphicsprogramming: you don't need to use the framebuffer at all

ext/CEGUI/IrrlichtBaWTexture.cpp
AnastaZIuk: adjusted in following commit

ext/CEGUI/IrrlichtBaWTexture.cpp
AnastaZIuk: adjusted in following commit

ext/CEGUI/IrrlichtBaWTexture.cpp
AnastaZIuk: adjusted in following commit

ext/CEGUI/IrrlichtBaWTexture.cpp
AnastaZIuk: adjusted in following commit

ext/CEGUI/IrrlichtBaWRenderTarget.cpp
AnastaZIuk: it is copied from their opengl implementation

ext/CEGUI/IrrlichtBaW/IrrlichtBaWTexture.cpp
AnastaZIuk: resolved

ext/CEGUI/IrrlichtBaWTexture.cpp
AnastaZIuk: adjusted in following commit

AnastaZIuk: but the area is covering area of framebuffer likely, how in that case I can copy it correctly?

AnastaZIuk: got an idea

AnastaZIuk: oh forget about regions haha

ext/CEGUI/IrrlichtBaWTexture.cpp
AnastaZIuk: do we implement `copy` for gpu objects?

AnastaZIuk: adjusted in following commit

ext/CEGUI/IrrlichtBaWTexture.cpp
AnastaZIuk: adjusted in following commit

ext/CEGUI/IrrlichtBaWTexture.cpp
AnastaZIuk: adjusted in following commit

ext/CEGUI/IrrlichtBaW/IrrlichtBaWTexture.hpp
AnastaZIuk: I know, but it's my habbit likely, I will remove those

ext/CEGUI/CMakeLists.txt
AnastaZIuk: resolved

ext/CEGUI/IrrlichtBaWTexture.cpp
AnastaZIuk: changed it a little

ext/CEGUI/IrrlichtBaWTexture.cpp
AnastaZIuk: adjusted in folllowing commit

ext/CEGUI/IrrlichtBaWTexture.cpp
AnastaZIuk: I'm not sure if GPU extent takes into account buffer row length, if for it it is indifferent, I will just use width

ext/CEGUI/IrrlichtBaWTexture.cpp
AnastaZIuk: they use it for caching afaik, doing like folllowing

```C++
//! container type used to hold Textures we create.
    typedef std::map<String, OpenGLTexture*, StringFastLessCompare
                     CEGUI_MAP_ALLOC(String, OpenGLTexture*)> TextureMap;
    //! Container used to track textures.
    TextureMap d_textures;
```

AnastaZIuk: the solution will be https://github.com/buildaworldnet/IrrlichtBAW/pull/480#discussion_r422015730

ext/CEGUI/IrrlichtBaWTexture.cpp
```diff
@@ -220,72 +152,117 @@ void IrrlichtBaWTexture::loadFromMemory(const void* buffer, const CEGUI::Sizef&
...
+    asset::E_FORMAT newFormat = [&]()
+    {
+        switch (ceguiFormat)
+        {
+            case PixelFormat::PF_RGB: return asset::EF_R8G8B8_SRGB;
+            case PixelFormat::PF_RGBA: return asset::EF_R8G8B8A8_SRGB;
+            case PixelFormat::PF_RGBA_4444: return asset::EF_R4G4B4A4_UNORM_PACK16;
+            case PixelFormat::PF_RGB_565: _IRR_FALLTHROUGH;
+            case PixelFormat::PF_PVRTC2: _IRR_FALLTHROUGH;
+            case PixelFormat::PF_PVRTC4: return asset::EF_UNKNOWN;
```
devshgraphicsprogramming, line 255: I think we have an SRGB enum for these two

ext/CEGUI/IrrlichtBaWTexture.cpp
```diff
@@ -220,72 +152,117 @@ void IrrlichtBaWTexture::loadFromMemory(const void* buffer, const CEGUI::Sizef&
...
+
+asset::E_FORMAT IrrlichtBaWTexture::getTranslatedFormat(const PixelFormat ceguiFormat)
+{
+    asset::E_FORMAT newFormat = [&]()
+    {
+        switch (ceguiFormat)
+        {
+            case PixelFormat::PF_RGB: return asset::EF_R8G8B8_SRGB;
+            case PixelFormat::PF_RGBA: return asset::EF_R8G8B8A8_SRGB;
+            case PixelFormat::PF_RGBA_4444: return asset::EF_R4G4B4A4_UNORM_PACK16;
```
devshgraphicsprogramming, line 252: don't support this one

ext/CEGUI/IrrlichtBaWTexture.cpp
```diff
@@ -220,72 +152,117 @@ void IrrlichtBaWTexture::loadFromMemory(const void* buffer, const CEGUI::Sizef&
...
+            case PixelFormat::PF_RGBA_DXT1: return asset::EF_BC1_RGBA_SRGB_BLOCK;
+            case PixelFormat::PF_RGBA_DXT3: return asset::EF_BC2_SRGB_BLOCK;
+            case PixelFormat::PF_RGBA_DXT5: return asset::EF_BC3_SRGB_BLOCK;
+
+        default:
+            return asset::EF_UNKNOWN;
+        }
+    }();
+
+    assert(newFormat != asset::EF_UNKNOWN, _IRR_CEGUI_UNSUPPORTED_FORMAT_.data());
```
devshgraphicsprogramming, line 266: don't assert, its not an error to ask

ext/CEGUI/IrrlichtBaWTexture.cpp
```diff
@@ -220,72 +152,117 @@ void IrrlichtBaWTexture::loadFromMemory(const void* buffer, const CEGUI::Sizef&
...
     }
     else
-        assert(status);
+        assert(status, _IRR_TEXTURE_MUST_EXIST_.data());
 }
 
 bool IrrlichtBaWTexture::isPixelFormatSupported(const PixelFormat fmt) const
 {
-    return true;
+    return getTranslatedFormat(fmt);
```
devshgraphicsprogramming, line 241: !=EF_UNNKNOWN

ext/CEGUI/IrrlichtBaWTexture.cpp
```diff
@@ -220,72 +152,117 @@ void IrrlichtBaWTexture::loadFromMemory(const void* buffer, const CEGUI::Sizef&
...
-        destinationBoundMemory->mapMemoryRange(video::IDriverMemoryAllocation::EMCAF_READ, { 0u, bufferByteSize });
-        core::smart_refctd_ptr<asset::ICPUBuffer> pixelBuffer = core::make_smart_refctd_ptr<asset::CCustomAllocatorCPUBuffer<core::null_allocator<uint8_t>>>(bufferByteSize, destinationBoundMemory->getMappedPointer(), core::adopt_memory);
-
-        memcpy(targetData, pixelBuffer->getPointer(), pixelBuffer->getSize());
-
-        temporaryOutFrameBuffer->drop();
+        const auto bufferByteSize = asset::getTexelOrBlockBytesize(viewParams.format) * extent.width * extent.height;
+
+        video::IDriverMemoryBacked::SDriverMemoryRequirements requirements = getMemoryRequirements(bufferByteSize);
+        auto gpuBuffer = driver->createGPUBufferOnDedMem(requirements);
```
devshgraphicsprogramming, line 216: don't create a new buffer, we have the downstreaming buffer, same thing we use in the screenshot extension!

ext/CEGUI/IrrlichtBaWTexture.cpp
```diff
@@ -220,72 +152,117 @@ void IrrlichtBaWTexture::loadFromMemory(const void* buffer, const CEGUI::Sizef&
...
+        region.bufferRowLength = 0;
+        region.bufferImageHeight = 0u;
+        region.imageOffset = { 0, 0, 0u };
+        region.imageExtent = extent;
+
+        driver->copyImageToBuffer(viewParams.image.get(), gpuBuffer.get(), regions->size(), regions->begin());
+        auto clientBoundMemory = gpuBuffer->getBoundMemory();
+        clientBoundMemory->mapMemoryRange(video::IDriverMemoryAllocation::EMCAF_READ, { 0u, bufferByteSize });
+        memcpy(targetData, clientBoundMemory->getMappedPointer(), bufferByteSize);
+        clientBoundMemory->unmapMemory();
```
devshgraphicsprogramming, line 233: maps in IrrlichtBaW are persistent, meaning they will not perform a fence before returning the pointer to the buffer!

downstreaming buffer is persistently mapped.

ext/CEGUI/IrrlichtBaWRenderTarget.cpp
devshgraphicsprogramming: well in that case its wrong XD

ext/CEGUI/IrrlichtBaWTexture.cpp
```diff
@@ -220,72 +152,117 @@ void IrrlichtBaWTexture::loadFromMemory(const void* buffer, const CEGUI::Sizef&
...
         auto& extent = viewParams.image->getCreationParameters().extent;
-        irr::core::recti coverArea(area.left(), area.top(), area.getWidth(), area.getHeight());
 
-        auto temporaryInFrameBuffer = createFrameBuffer(sourceData, viewParams.format, extent.width, extent.height);
-        driver->blitRenderTargets(temporaryInFrameBuffer, frameBuffer, false, false, coverArea, coverArea);
-        temporaryInFrameBuffer->drop();
+        auto requirements = getMemoryRequirements(area.getWidth() * area.getHeight() * asset::getTexelOrBlockBytesize(viewParams.format));
+
+        auto gpuBuffer = driver->createGPUBufferOnDedMem(requirements);
+        driver->updateBufferRangeViaStagingBuffer(gpuBuffer.get(), 0, requirements.vulkanReqs.size, sourceData);
```
devshgraphicsprogramming, line 186: you can `copyBufferToImage` straight from the streaming/staging buffer... here you're creating one buffer too many

ext/CEGUI/IrrlichtBaW/IrrlichtBaWTexture.hpp
devshgraphicsprogramming: on other classes too

ext/CEGUI/IrrlichtBaWTexture.cpp
```diff
@@ -5,80 +5,35 @@ using namespace irr;
...
-{
-
-}
+static constexpr std::string_view _IRR_TEXTURE_MUST_EXIST_ = "To perform the function, gpu texture must exist at that point!";
+static constexpr std::string_view _IRR_CEGUI_UNSUPPORTED_FORMAT_ = "Unsupported pixel format detected!";
+static constexpr std::string_view _IRR_WRONG_CASTING_ = "Something went wrong while casting, empty pointer detected!";
+static constexpr std::string_view _IRR_NOT_TIGHTLY_PACKED_ = "Width of memory data passed doesn't pass requirements - it isn't tightly packed!";
 
-IrrlichtBaWTexture::~IrrlichtBaWTexture()
+video::IDriverMemoryBacked::SDriverMemoryRequirements getMemoryRequirements(size_t vulkanReqsSize)
```
devshgraphicsprogramming, line 13: you don't need this function, because you don't need to create transient buffers (we already have those, they're called the up and downstreaming buffer)

ext/CEGUI/IrrlichtBaWTexture.cpp
```diff
@@ -88,43 +43,43 @@ const CEGUI::String& IrrlichtBaWTexture::getName() const
...
 const CEGUI::Vector2f& IrrlichtBaWTexture::getTexelScaling() const
 {
-    bool status = frameBuffer;
+    bool status = gpuImageViewTexture.get();
 
     if (status)
     {
-        auto& params = frameBuffer->getAttachment(textureColorAttachment)->getCreationParameters().image->getCreationParameters();
+        auto& params = gpuImageViewTexture->getCreationParameters().image->getCreationParameters();
         auto width = bufferRowLength ? bufferRowLength : params.extent.width;
```
devshgraphicsprogramming, line 83: that's not the correct width for texture scaling to return.

ext/CEGUI/IrrlichtBaWTexture.cpp
devshgraphicsprogramming: you're not even using this variable

ext/CEGUI/IrrlichtBaWTexture.cpp
devshgraphicsprogramming: you're still doing it wrong, you've creating a ICPUImageView not an IGPUImageView

ext/CEGUI/IrrlichtBaWTexture.cpp
devshgraphicsprogramming: I think its possible to create a "clean" asset manager class (`core::make_smart_refctd_ptr<IAssetManager>()`) that could be used just by the CEGUI renderer components, or at least use an override during loading that could prepend `?CEGUI?` to all the caching paths when inserting into cache and searching the cache

ext/CEGUI/IrrlichtBaWTexture.cpp
```diff
@@ -220,72 +152,117 @@ void IrrlichtBaWTexture::loadFromMemory(const void* buffer, const CEGUI::Sizef&
...
     region.bufferOffset = 0u;
     region.bufferRowLength = 0;
-    region.bufferImageHeight = 0u; 
+    region.bufferImageHeight = 0u;
     region.imageOffset = { 0u, 0u, 0u };
     region.imageExtent = imageParams.extent;
 
-    auto gpuImage = device->getVideoDriver()->createGPUImage(std::move(imageParams));
-    driver->copyBufferToImage(gpuPixelBuffer.get(), gpuImage.get(), regions->size(), regions->begin());
+    driver->copyBufferToImage(gpuBuffer.get(), gpuImage.get(), regions->size(), regions->begin());
```
devshgraphicsprogramming, line 159: first create the GPU Image, then do a `copyBufferToImage` but via the upstreaming staging buffer, don't create new buffers!

devshgraphicsprogramming: also DRY, so create the IGPUImage and the views, then call `blitFromMemory`

devshgraphicsprogramming:
closing temporarily.... reopen when ready for merge

PR #484, opened 2020-05-05
Stuff doesn't work need to start with clean repo

devshgraphicsprogramming:


devshgraphicsprogramming:
What could be upstreamed, had been upstreamed

PR #487, opened 2020-05-11
LRU cache

Hazardu:


devshgraphicsprogramming:
Does this work and has it been actually tested?

src/irr/core/LRUcache.h
```diff
@@ -0,0 +1,58 @@
...
+			inline void insert(Key &k, Value &v)
+			{
+				if (map.find(k) == map.end() && stored_keys.size() == cap) {
+					auto last = stored_keys.back();
+					stored_keys.pop_back();
+					map.erase(last);
+				}
+				else
+					stored_keys.remove(k);
+				stored_keys.push_front(k);
```
devshgraphicsprogramming: this works wrong.

If the cap would be overflown, we always need to remove the oldest value.

See here
https://www.geeksforgeeks.org/lru-cache-implementation/

src/irr/core/LRUcache.h
```diff
@@ -0,0 +1,58 @@
...
+
+namespace irr {
+	namespace core {
+
+		template<typename Key, typename Value,typename MapHash,typename MapEquals>
+		class LRUcache {
+			list<Key> stored_keys;
+			unordered_map<Key, Value, MapHash,MapEquals> map;
+			uint32_t cap;
+			std::atomic_uint32_t timestamp;
```
devshgraphicsprogramming: timestamp can be made redundant (its implicit)

src/irr/core/LRUcache.h
```diff
@@ -0,0 +1,58 @@
...
+		public:
+			inline LRUcache() = default;
+			inline LRUcache(uint32_t& capacity)
+			{
+				cap = capacity;
```
devshgraphicsprogramming: reserve memory in the `map`

src/irr/core/LRUcache.h
```diff
@@ -0,0 +1,58 @@
...
+#include "irr/core/Types.h"
+#include "COpenGLDriver.h"
+
+namespace irr {
+	namespace core {
+
+		template<typename Key, typename Value,typename MapHash,typename MapEquals>
+		class LRUcache {
+			list<Key> stored_keys;
+			unordered_map<Key, Value, MapHash,MapEquals> map;
```
devshgraphicsprogramming: use a `list<pair<Key,Value>>`, I use it like that in the `DeferredEventHandler` to implement a queue (although I only need forward linked list there).

However you need to create your own implementation of a Doubly Linked List using a pool address allocator such that every node is not dynamically allocated from the heap (its possible that its just enough to implement a custom allocator on top of `PoolAddressAllocator` and pass that in as a template argument for `list`, but only if `list` calls the allocator to allocate one node at a time).

then `unordered_map<Key,stored_keys::iterator,MapHash,MapEquals>` to be the "skip list"

First you need a `common_peek`:
1) look up the hashmap with key
2) use hashmap's iterator (pointer) to the linked list node

When you `peek`:
1) call `common_peek` to find yout node and return node's value

When you `get`:
1) call `common_peek` to find your node
2) move the node from its current position in the list to the back
3) return node's value

When you `insert` (you assume that user cannot call `insert` twice on the same key):
1) if cache cap has been reached, you call `common_erase` on the front linked list node
2) construct a linked list node out of `Key,Value` and add it to the back of the list
3) record the iterator (pointer) to that node into the `unordered_map`

When you `erase`:
1) get the node by calling `common_peek`
2) call `common_erase` on the found node

In `common_erase`:
1) remove node from the linked list
2) look up the node's "skip link" in the unordered map and remove it

Hazardu: Since timestamp is redundant, what is the difference between peek and get?

devshgraphicsprogramming: explained on discord

src/irr/core/CDoublyLinkedList.h
```diff
@@ -0,0 +1,83 @@
+#ifndef __DOUBLY_LINKED_LIST_H_INCLUDED__
+#define __DOUBLY_LINKED_LIST_H_INCLUDED__
+#include "irr/core/alloc/PoolAddressAllocator.h"
+#include "irr/core/Types.h"
+
+namespace irr {
+	namespace core {
+
+		template<typename Value>
```
devshgraphicsprogramming: add a pointer/iterator type, so you're not always forced to use `struct Snode*` as the pointer type

src/irr/core/CDoublyLinkedList.h
```diff
@@ -0,0 +1,83 @@
...
+		class DoublyLinkedList
+		{
+			struct Snode
+			{
+				struct Snode* prev;
+				Value data;
+				struct Snode* next;
```
devshgraphicsprogramming: correct layout is
```cpp
Value data;
iterator_type prev,next;
```

because `Value` could be a type with complex alignment requirements (i.e. SSE/AVX data structure aligned to 16 or 32 bytes or more) and it would create padding inbetween your members

src/irr/core/CDoublyLinkedList.h
```diff
@@ -0,0 +1,83 @@
...
+				node->prev = nullptr;
+				begin = node;
+			}
+			DoublyLinkedList(uint32_t& capacity) 
+			{
+				alloc = new PoolAddressAllocator<uint32_t>(malloc(sizeof(Snode) * capacity), 1u, 1u, 
+					sizeof(Snode) * (capacity - 1u),	//max allocatable alligment - should it be without `sizeof(Snode) `
+					sizeof(Snode), 1u);
+
+			}
```
devshgraphicsprogramming: did you pull this code off the internet or something?

src/irr/core/CDoublyLinkedList.h
```diff
@@ -0,0 +1,83 @@
...
+				Value data;
+				struct Snode* next;
+
+				Snode(Value v) { data = v; }
+			};
+
+			uint32_t cap;
+			Snode* begin;
+			Snode* end;
+			PoolAddressAllocator<uint32_t> alloc;
```
devshgraphicsprogramming: you always want to place the fattest-most-aligned member first... biggest to smallest.

src/irr/core/CDoublyLinkedList.h
Hazardu: no

src/irr/core/CDoublyLinkedList.h
```diff
@@ -0,0 +1,83 @@
...
+		template<typename Value>
+		class DoublyLinkedList
+		{
+			struct Snode
+			{
+				struct Snode* prev;
+				Value data;
+				struct Snode* next;
+
+				Snode(Value v) { data = v; }
```
devshgraphicsprogramming: make value a `const Value&`
then add a move constructor

as well as copy assignment and move assignment.

src/irr/core/CDoublyLinkedList.h
```diff
@@ -0,0 +1,83 @@
...
+				begin->prev = node;
+				node->next = begin;
+				node->prev = nullptr;
+				begin = node;
+			}
+			DoublyLinkedList(uint32_t& capacity) 
+			{
+				alloc = new PoolAddressAllocator<uint32_t>(malloc(sizeof(Snode) * capacity), 1u, 1u, 
+					sizeof(Snode) * (capacity - 1u),	//max allocatable alligment - should it be without `sizeof(Snode) `
+					sizeof(Snode), 1u);
```
devshgraphicsprogramming: if you take out all the `sizeof(Snode)` you will get good alignment, and sequential addresses.

src/irr/core/CDoublyLinkedList.h
```diff
@@ -0,0 +1,83 @@
...
+			{
+				if (begin == node) return;
+				begin->prev = node;
+				node->next = begin;
+				node->prev = nullptr;
+				begin = node;
+			}
+			DoublyLinkedList(uint32_t& capacity) 
+			{
+				alloc = new PoolAddressAllocator<uint32_t>(malloc(sizeof(Snode) * capacity), 1u, 1u,
```
devshgraphicsprogramming: the memory you give to an address allocator is not for storing data.. its for storing internal state, familiarize yourself with `PoolAddressAllocator::reserved_size`.

src/irr/core/CDoublyLinkedList.h
```diff
@@ -0,0 +1,83 @@
...
+			Snode* end;
+			PoolAddressAllocator<uint32_t> alloc;
+
+			inline void popBack()
+			{
+				if(end->prev != nullptr)
+					end->prev->next = nullptr;
+				Snode* temp = end;
+				end = end->prev;
+				alloc.free_addr(reinterpret_cast<uint32_t>(temp), sizeof(Snode));
```
devshgraphicsprogramming: you need to explictly call the destructor on the node you're removing, `Value` may be a non-trivial type

src/irr/core/CDoublyLinkedList.h
```diff
@@ -0,0 +1,83 @@
...
+			}
+
+		public:
+
+			inline void pushFront(Value &val) 
+			{
+				auto p = alloc.alloc_addr(sizeof(Snode),0u)
+				Snode n = new(reinterpret_cast<void*>(p)) Snode(val);
```
devshgraphicsprogramming: `alloc_addr` gets you an offset into a data store you've created and maintain by yourself.... not a pointer to a data store

correct solution would be
```cpp
uint32_t addr = alloc.alloc_addr(1u,1u);
Snode* node = new(myDataStore+addr) Snode(val);
```

src/irr/core/CDoublyLinkedList.h
```diff
@@ -0,0 +1,83 @@
+#ifndef __DOUBLY_LINKED_LIST_H_INCLUDED__
+#define __DOUBLY_LINKED_LIST_H_INCLUDED__
+#include "irr/core/alloc/PoolAddressAllocator.h"
+#include "irr/core/Types.h"
+
+namespace irr {
+	namespace core {
+
+		template<typename Value>
+		class DoublyLinkedList
```
devshgraphicsprogramming: call it a `FixedCapacityLinkedList` since it cant be grown over its initial capacity.

src/irr/core/CDoublyLinkedList.h
```diff
@@ -0,0 +1,83 @@
...
+		{
+			struct Snode
```
devshgraphicsprogramming: I'd define this struct outside in its own header, might be useful on its own

devshgraphicsprogramming:
can you merge `shader_pipeline` into your branch so the diff is not insane?

src/irr/core/CFixedCapacityDoublyLinkedList.h
```diff
@@ -21,20 +21,19 @@ namespace irr {
 		template<typename Value>
 		class DoublyLinkedList
 		{
-		public:
-
+			typedef Snode<Value>* iteratorptr_t;
```
devshgraphicsprogramming: make it a `uint32_t`

Hazardu:
@devshgraphicsprogramming could you assist me with assigning the proper parameters for allocator's constructor in CFixedCapacityDoublyLinkedList.h?

src/irr/core/CFixedCapacityDoublyLinkedList.h
Hazardu: Change the `typedef` to `uint32_t` and cast to `Snode< >` in every function where its used? Isn't that a bit lengthy way to do this?

devshgraphicsprogramming: no use the `uint32_t` to index into an array of `SNode` owned by the `CFixedCapacityDoublyLinkedList`.

Hazardu: I've had quite a bit of trouble find a way to do it in a performance friendly way, checking online yields only a result similar to my own conclusion. Indexing a linked list will slow things down. Inserting and removing will require a lot more operations, and the action of removing last element and inserting a new one will become `O(2n)`

Hazardu: > the action of removing last element and inserting a new one will become O(2n)

actually, just O(n), but still bad 💯

src/irr/core/CFixedCapacityDoublyLinkedList.h
```diff
@@ -0,0 +1,117 @@
...
+
+namespace irr {
+	namespace core {
+
+		template<typename Value>
+		struct Snode
+		{
+			Value data;
+			Snode* prev;
+			Snode* next;
```
devshgraphicsprogramming: these I want to be addresses not pointers

src/irr/core/CFixedCapacityDoublyLinkedList.h
```diff
@@ -0,0 +1,117 @@
...
+				this.prev = std::move(other.prev);
+				this.next = std::move(other.next);
+				return *this;
+			}
+		};
+
+		template<typename Value>
+		class DoublyLinkedList
+		{
+			typedef Snode<Value>* iteratorptr_t;
```
devshgraphicsprogramming: same here, address not pointer

src/irr/core/CFixedCapacityDoublyLinkedList.h
```diff
@@ -0,0 +1,117 @@
...
+			void* reservedSpace;
+			PoolAddressAllocator<uint32_t> alloc;
+			iteratorptr_t p_begin;
+			iteratorptr_t p_end;
+			uint32_t cap;
+
+			inline void popBack()
+			{
+				if(p_end->prev != nullptr)
+					p_end->prev->next = nullptr;
```
devshgraphicsprogramming: instead of `nullptr` define an `invalid_iterator` (which in this case would be the pool allocator's invalid address)

devshgraphicsprogramming: also `p_end` doesn't and shouldn't exist, you should keep `p_back` which is your `p_end->prev`

devshgraphicsprogramming: usually in STL compatible containers "end" means one past the container, where as "back" is the true last element.. see std::vector and std::array

src/irr/core/CFixedCapacityDoublyLinkedList.h
```diff
@@ -0,0 +1,117 @@
...
+				alloc.free_addr(reinterpret_cast<uint32_t>(temp), 1u);
+			}
+
+		public:
+
+			inline void pushFront(Value &val) 
+			{
+
+				uint32_t addr = alloc.alloc_addr(1u, 1u);
+				iteratorptr_t n = new(reinterpret_cast<iteratorptr_t>(reservedSpace) + addr) Snode(val);
```
devshgraphicsprogramming: no that's still not how it works!

reserved space is for the ALLOCATOR INTERNAL STATE.

You need to create a separate array of `SNode[cap]`

src/irr/core/CFixedCapacityDoublyLinkedList.h
devshgraphicsprogramming: it seems like your `p_end` functions like a `p_back`, if so, rename it.

src/irr/core/CFixedCapacityDoublyLinkedList.h
```diff
@@ -0,0 +1,117 @@
...
+
+			inline iteratorptr_t getLast() { return p_end; }
+
+			inline void erase(iteratorptr_t node)
+			{
+				if(node->prev != nullptr)
+				node->prev->next = node->next;
+				if (node->next != nullptr)
+				node->next->prev = node->prev;
+				delete node;
```
devshgraphicsprogramming: nope.. `node->~SNode();` delete is a call to system's `free()`

src/irr/core/CFixedCapacityDoublyLinkedList.h
```diff
@@ -0,0 +1,117 @@
...
+				alloc.free_addr(reinterpret_cast<uint32_t>(node), 1u);
+			}
+
+			inline void moveToFront(iteratorptr_t node)
+			{
+				if (p_begin == node) return;
+				p_begin->prev = node;
+				node->next = p_begin;
+				node->prev = nullptr;
+				p_begin = node;
```
devshgraphicsprogramming: what about the `node->next->prev` and `node->prev->next` ? you break the chain at `node` !

src/irr/core/CFixedCapacityDoublyLinkedList.h
```diff
@@ -0,0 +1,117 @@
...
+			}
+
+			DoublyLinkedList(const uint32_t& capacity) :
+				alloc(reservedSpace, 0u,0u, (capacity-1), capacity, 1u),
+				reservedSpace(malloc(sizeof(Snode<Value>)* capacity)),
+				cap(capacity)
```
devshgraphicsprogramming: first allocate the reserved space,
```cpp
reservedSpace(_IRR_ALIGNED_MALLOC(PoolAddressAllocator<uint32_t>::reserved_size(1u,MAX_NUMBER_OF_NODES_IN_YOUR_LINKED_LIST,1u),alignof(void*)))
```

devshgraphicsprogramming: first allocate the reserved space,
```cpp
reservedSpace(_IRR_ALIGNED_MALLOC(PoolAddressAllocator<uint32_t>::reserved_size(1u,MAX_NUMBER_OF_NODES_IN_YOUR_LINKED_LIST,1u),alignof(void*)))
```

devshgraphicsprogramming: first allocate the reserved space,
```cpp
reservedSpace(_IRR_ALIGNED_MALLOC(PoolAddressAllocator<uint32_t>::reserved_size(1u,MAX_NUMBER_OF_NODES_IN_YOUR_LINKED_LIST,1u),alignof(void*)))
```

devshgraphicsprogramming: then you can construct the allocator
```cpp
alloc(reservedSpace, 0u,0u,1u, MAX_NUMBER_OF_NODES_IN_YOUR_LINKED_LIST, 1u)
```

The `reservedSpace` is REQUIRED TO BE AVIALABLE AT ALOCATOR CONSTRUCTION TIME BECAUSE INTERNAL STATE GETS SET UP IMMEDIATELY!

P.S. you've `malloc`ed the reserved space, but where is the `free` (or better `_IRR_ALIGNED_FREE` ) ?

devshgraphicsprogramming: the actualy storage for the `SNode`s is yet another array you have to create for yourself.

src/irr/core/CFixedCapacityDoublyLinkedList.h
Hazardu: `node->next->prev` would be line 94 p_begin->prev = node;
inserted node is put at the front, therefore `node->prev == invalid_iterator`

src/irr/core/CFixedCapacityDoublyLinkedList.h
```diff
@@ -10,23 +10,23 @@ namespace irr {
 		struct Snode
 		{
 			Value data;
-			Snode* prev;
-			Snode* next;
+			uint32_t prev;
+			uint32_t next;
 
 			Snode(const Value& val) : data(val)
```
devshgraphicsprogramming, line 20: init the `prev` and `next` to invalid address

src/irr/core/CFixedCapacityDoublyLinkedList.h
```diff
@@ -38,73 +38,84 @@ namespace irr {
...
 			void* reservedSpace;
 			PoolAddressAllocator<uint32_t> alloc;
-			iteratorptr_t p_begin;
-			iteratorptr_t p_end;
+			node_address m_back;
+			node_address m_begin;
+			Snode<Value>* m_array;
 			uint32_t cap;
 
+			#define invalid_iterator alloc.invalid_address	//0xdeadbeef
```
devshgraphicsprogramming: IRR_STATIC_INLINE_CONSTEXPR

src/irr/core/CFixedCapacityDoublyLinkedList.h
```diff
@@ -38,73 +38,84 @@ namespace irr {
...
 			{
-				if(p_end->prev != nullptr)
-					p_end->prev->next = nullptr;
-				iteratorptr_t temp = p_end;
-				p_end = p_end->prev;
-				alloc.free_addr(reinterpret_cast<uint32_t>(temp), 1u);
+				if(m_back->prev != invalid_iterator)
+					m_back->prev->next = invalid_iterator;
+				uint32_t temp = m_end;
+				m_end = m_back->prev;
```
devshgraphicsprogramming: what is `m_end` ?

src/irr/core/CFixedCapacityDoublyLinkedList.h
```diff
@@ -38,73 +38,84 @@ namespace irr {
...
-				n->prev = nullptr;
-				
-				n->next = p_begin;
-
-				if (p_begin != nullptr)
-					p_begin->prev = n;
-				if (p_end == nullptr)
-					p_end = n;
-				p_begin = n;
+				Snode<Value>* n = (reinterpret_cast<Snode<Value>*>(reservedSpace) + addr);
```
devshgraphicsprogramming: FACEPALM

src/irr/core/CFixedCapacityDoublyLinkedList.h
```diff
@@ -38,73 +38,84 @@ namespace irr {
...
+				if(m_back->prev != invalid_iterator)
+					m_back->prev->next = invalid_iterator;
+				uint32_t temp = m_end;
+				m_end = m_back->prev;
+				alloc.free_addr(temp, 1u);
 			}
 
-		public:
-
 			inline void pushFront(Value &val)
```
devshgraphicsprogramming: make this an r-value reference

src/irr/core/CFixedCapacityDoublyLinkedList.h
```diff
@@ -38,73 +38,84 @@ namespace irr {
...
-				n->next = p_begin;
-
-				if (p_begin != nullptr)
-					p_begin->prev = n;
-				if (p_end == nullptr)
-					p_end = n;
-				p_begin = n;
+				Snode<Value>* n = (reinterpret_cast<Snode<Value>*>(reservedSpace) + addr);
+				n->prev = invalid_iterator;
+				n->data = val;
```
devshgraphicsprogramming: std::move the val

src/irr/core/CFixedCapacityDoublyLinkedList.h
```diff
@@ -38,73 +38,84 @@ namespace irr {
...
+				n->next = m_begin;
+
+				m_array[addr] = *n;
```
devshgraphicsprogramming: this line shouldn't be needed.

src/irr/core/CFixedCapacityDoublyLinkedList.h
```diff
@@ -38,73 +38,84 @@ namespace irr {
...
+				if (m_begin != invalid_iterator)
+					m_begin->prev = n;
+				m_begin = n;
```
devshgraphicsprogramming: `m_begin = addr;`

src/irr/core/CFixedCapacityDoublyLinkedList.h
devshgraphicsprogramming: also you cant just use `m_end->` and `m_begin->` , you'd need to implement a tiny allocator wrapper struct over the `uint32_t` to be able to do that.... or just use `m_array[m_end].` instead.

src/irr/core/CFixedCapacityDoublyLinkedList.h
```diff
@@ -38,73 +38,84 @@ namespace irr {
...
 			}
 
-			inline iteratorptr_t getFirst() { return p_begin; }
+			inline node_address getBegin() { return m_begin; }
 
-			inline iteratorptr_t getLast() { return p_end; }
+			inline node_address getBack() { return m_back; }
```
devshgraphicsprogramming: this needs to return a pointer to SNode not an address

src/irr/core/CFixedCapacityDoublyLinkedList.h
```diff
@@ -38,73 +38,84 @@ namespace irr {
...
 
-			inline void erase(iteratorptr_t node)
+			inline void erase(uint32_t& nodeAddr)
```
devshgraphicsprogramming: no point for a reference of a type that's less than or equal 8 bytes

src/irr/core/CFixedCapacityDoublyLinkedList.h
```diff
@@ -38,73 +38,84 @@ namespace irr {
...
-				reservedSpace(malloc(sizeof(Snode<Value>)* capacity)),
+				alloc(reservedSpace, 0u, 0u, 1u, capacity, 1u),
+				reservedSpace(_IRR_ALIGNED_MALLOC(PoolAddressAllocator<uint32_t>::reserved_size(1u, capacity, 1u), alignof(void*))),
 				cap(capacity)
 			{
-				p_begin = nullptr;
-				p_end = nullptr;
+				m_back = invalid_iterator;
+				m_begin = invalid_iterator;
+				m_array = new (reservedSpace) Snode<Value>;
```
devshgraphicsprogramming: you dont use placement new, also RESERVED SPACE DOES NOT OVERLAP WITH DATA STORAGE!

src/irr/core/CFixedCapacityDoublyLinkedList.h
```diff
@@ -0,0 +1,128 @@
...
+				if (node->prev != invalid_iterator)
+					node->next->prev = node->prev;
+				node->next = m_begin;
+				node->prev = invalid_iterator;
+				m_begin = node;
+			}
+
+			DoublyLinkedList(const uint32_t& capacity) :
+				alloc(reservedSpace, 0u, 0u, 1u, capacity, 1u),
+				reservedSpace(_IRR_ALIGNED_MALLOC(PoolAddressAllocator<uint32_t>::reserved_size(1u, capacity, 1u), alignof(void*))),
```
devshgraphicsprogramming: YOU NEED THE RESERVED SPACE TO BE AVAILABLE TO THE ALLOCATOR FROM THE START, WRONG ORDER OF INITIALIZATION!

src/irr/core/CFixedCapacityDoublyLinkedList.h
```diff
@@ -5,18 +5,25 @@
 
 namespace irr {
 	namespace core {
-
 		template<typename Value>
 		struct Snode
 		{
+			_IRR_STATIC_INLINE_CONSTEXPR uint32_t invalid_iterator = 0xdeadbeefu;
```
devshgraphicsprogramming: take the value from the address allocator or pass it through somehow

src/irr/core/CFixedCapacityDoublyLinkedList.h
```diff
@@ -0,0 +1,161 @@
...
+
+			void* reservedSpace;
+			PoolAddressAllocator<uint32_t> alloc;
+			uint32_t m_back;
+			uint32_t m_begin;
+			Snode<Value>* m_array;
+			uint32_t cap;
+
+		public:
+			_IRR_STATIC_INLINE_CONSTEXPR uint32_t invalid_iterator = 0xdeadbeefu;
```
devshgraphicsprogramming: this needs to come from the address allocator `invalid_address` constexpr (we plan to change from `0deadbeefu` eventually)

src/irr/core/CFixedCapacityDoublyLinkedList.h
```diff
@@ -0,0 +1,161 @@
...
+
+			//get node at iterator
+			inline Snode<Value>* get(uint32_t address)
```
devshgraphicsprogramming, line 67: make this get private, the public get should take a key and use `common_peek`

https://github.com/buildaworldnet/IrrlichtBAW/pull/487#discussion_r472068256

src/irr/core/CFixedCapacityDoublyLinkedList.h
```diff
@@ -0,0 +1,161 @@
...
+				n->prev = invalid_iterator;
+				n->next = m_begin;
+
+				if (m_begin != invalid_iterator)
+					getBegin()->prev = addr;
+				m_begin = addr;
+			}
+
+			//get node ptr of the first item in the list
+			inline Snode<Value>* getBegin() { return &(m_array[m_begin]); }
```
devshgraphicsprogramming: I'd rather use `m_array+m_begin` notation instead of `&(m_array[m_begin])`

src/irr/core/CFixedCapacityDoublyLinkedList.h
```diff
@@ -0,0 +1,161 @@
...
+				alloc(reservedSpace, 0u, 0u, 1u, capacity, 1u)
+			{
+				m_back = invalid_iterator;
+				m_begin = invalid_iterator;
+				m_array = new Snode<Value>[capacity];
+			}
+			~DoublyLinkedList()
+			{
+				_IRR_ALIGNED_FREE(reservedSpace);
+				free(m_array);
```
devshgraphicsprogramming: you call `delete []` here... because you used array new! (otherwise the destructors of the nodes never get called, also its bad C++ practice and ASAN will scream at you)

devshgraphicsprogramming: if you don't intend to call destructors automatically, you should not use`new []` in the first place.

src/irr/core/CFixedCapacityDoublyLinkedList.h
```diff
@@ -0,0 +1,161 @@
...
+			//remove the last element in the list
+			inline void popBack()
+			{
+				if (m_back == invalid_iterator)	return;
+				auto backNode = getBack();
+				if (backNode->prev != invalid_iterator)
+					get(backNode->prev)->next = invalid_iterator;
+				uint32_t temp = m_back;
+				m_back = backNode->prev;
+				alloc.free_addr(temp, 1u);
```
devshgraphicsprogramming: you need to call the destructor on the memory of the back node you've just deallocated.

src/irr/core/CFixedCapacityDoublyLinkedList.h
```diff
@@ -0,0 +1,161 @@
...
+			}
+
+			//add new item to the list. This function does not make space to store the new node. in case the list is full, popBack() needs to be called beforehand
+			inline void pushFront(Value&& val)
+			{
+				uint32_t addr = alloc.alloc_addr(1u, 1u);
+				assert(addr < cap);
+				m_array[addr] = *(new Snode<Value>(std::move(val)));
```
devshgraphicsprogramming: that's absolutely not how it works, do a proper placement new!

devshgraphicsprogramming: `Snode<value>* node = new(myDataStore+addr) Snode<value>(val);`

src/irr/core/CFixedCapacityDoublyLinkedList.h
```diff
@@ -0,0 +1,161 @@
...
+			//remove a node at nodeAddr from the list
+			inline void erase(uint32_t nodeAddr)
+			{
+				assert(nodeAddr != invalid_iterator);
+				auto node = get(nodeAddr);
+				if (node->prev != invalid_iterator)
+					get(node->prev)->next = node->next;
+				if (node->next != invalid_iterator)
+					get(node->next)->prev = node->prev;
+				//node->~Snode<Value>();	destructor not defined
```
devshgraphicsprogramming: this needs to happen, as in the other erase

devshgraphicsprogramming: define the destructor.

src/irr/core/CFixedCapacityDoublyLinkedList.h
```diff
@@ -0,0 +1,161 @@
...
+				alloc.free_addr(nodeAddr, 1u);
+			}
+
+			//move a node at nodeAddr to the front of the list
+			inline void moveToFront(uint32_t nodeAddr)
+			{
+				if (m_begin == nodeAddr) return;
```
devshgraphicsprogramming: also protect against invalid addr

src/irr/core/CFixedCapacityDoublyLinkedList.h
```diff
@@ -0,0 +1,161 @@
...
+				getBegin()->prev = nodeAddr;
+
+				auto node = get(nodeAddr);
+				if (node->next != invalid_iterator)
+					get(node->next)->prev = node->prev;
+				if (node->prev != invalid_iterator)
+					get(node->next)->prev = node->prev;
```
devshgraphicsprogramming: typo.

src/irr/core/CFixedCapacityDoublyLinkedList.h
```diff
@@ -0,0 +1,161 @@
...
+				node->next = m_begin;
+				node->prev = invalid_iterator;
+				m_begin = nodeAddr;
```
devshgraphicsprogramming, line 163: erm what about the `prev` of the old `m_begin` node?

src/irr/core/CFixedCapacityDoublyLinkedList.h
devshgraphicsprogramming: but you'd need to call the destructors for all nodes that were still allocated at the time of the destructor of the double linked list (you can abuse knowledge of the pool allocator internal state, it actually keeps a stack of all "free" addresses in its reserved memory, you can sort that stack and iterate through the whole address space skipping the destructor call on the "free" addresses)

It will be O(nlogn) or O(n) for the sort of free list, then just O(n) to walk the list and skip addresses.

src/irr/core/LRUcache.h
```diff
@@ -0,0 +1,131 @@
...
+	DoublyLinkedList<list_template_t> m_list;
+	unordered_map<Key, uint32_t, MapHash, MapEquals> m_shortcut_map;
+	uint32_t m_capacity;
+
+	//get iterator associated with a key, or invalid_iterator if key is not within the cache
+	inline uint32_t common_peek(Key& key)
+	{
+		if (m_shortcut_map.find(key) != m_shortcut_map.end())
+		{
+			return m_shortcut_map[key];
```
devshgraphicsprogramming: return the dereferenced found iterator, this `operator[]` will trigger yet another search that you don't need.

src/irr/core/LRUcache.h
```diff
@@ -0,0 +1,131 @@
...
+//Stores fixed size amount of elements. 
+//When inserting new, prevents overflow by removing the least recently used element.
+template<typename Key, typename Value, typename MapHash = std::hash<Key>, typename MapEquals = std::equal_to<Key>>
+class LRUcache
+{
+	typedef std::pair<Key, Value> list_template_t;
+	_IRR_STATIC_INLINE_CONSTEXPR uint32_t invalid_iterator = Snode<list_template_t>::invalid_iterator;
+
+	DoublyLinkedList<list_template_t> m_list;
+	unordered_map<Key, uint32_t, MapHash, MapEquals> m_shortcut_map;
```
devshgraphicsprogramming: why do you need the `Key` to live both in the linked list and the unordered map?

I know I gave a such example but it seems not necessary to duplicate it, @Crisspl, your opinion?

src/irr/core/LRUcache.h
```diff
@@ -0,0 +1,131 @@
...
+	uint32_t m_capacity;
+
+	//get iterator associated with a key, or invalid_iterator if key is not within the cache
+	inline uint32_t common_peek(Key& key)
```
devshgraphicsprogramming, line 24: make a `const` qualifier overload as well

src/irr/core/LRUcache.h
devshgraphicsprogramming: mostly why does the key live in the linked list?

src/irr/core/LRUcache.h
```diff
@@ -0,0 +1,131 @@
...
+		auto iterator = m_shortcut_map.find(k);
+		if (iterator == m_shortcut_map.end())
+		{
+			if (m_shortcut_map.size() >= m_capacity)
+			{
+				m_shortcut_map.erase(m_list.getBack()->data.first);
+				m_list.popBack();
+			}
+			m_list.pushFront(std::pair<Key, Value>(k, v));
+			m_shortcut_map[k] = m_list.getFirstAddress();
```
devshgraphicsprogramming: use insert with a iterator hint, the insert might be faster.

src/irr/core/LRUcache.h
```diff
@@ -0,0 +1,131 @@
...
+	}
+#endif // _IRR_DEBUG
+
+	//add element to the cache, or move to the front if key exists. In case of the latter, it doesnt update the value
+	inline void insert(Key& k, const Value& v)
+	{
+		auto iterator = m_shortcut_map.find(k);
+		if (iterator == m_shortcut_map.end())
+		{
+			if (m_shortcut_map.size() >= m_capacity)
```
devshgraphicsprogramming, line 48: shouldn't this be `common_erase` or something?

src/irr/core/LRUcache.h
```diff
@@ -0,0 +1,131 @@
...
+			{
+				m_shortcut_map.erase(m_list.getBack()->data.first);
+				m_list.popBack();
+			}
+			m_list.pushFront(std::pair<Key, Value>(k, v));
+			m_shortcut_map[k] = m_list.getFirstAddress();
+		}
+		else
+		{
+			m_list.get(iterator->second)->data = std::pair<Key, Value>(k, v);
```
devshgraphicsprogramming: you only need to swap the value.

src/irr/core/LRUcache.h
```diff
@@ -0,0 +1,131 @@
...
+			std::cout << node->data.first << ' ' << node->data.second << std::endl;
+			if (node->next == invalid_iterator)
+				break;
+			node = m_list.get(node->next);
+		}
+	}
+#endif // _IRR_DEBUG
+
+	//add element to the cache, or move to the front if key exists. In case of the latter, it doesnt update the value
+	inline void insert(Key& k, const Value& v)
```
devshgraphicsprogramming: `Value` should be always an "r-value", there should probably be two overloads of this method for a `const Key&` and a `Key&&`

src/irr/core/LRUcache.h
```diff
@@ -0,0 +1,131 @@
...
+			m_list.pushFront(std::pair<Key, Value>(k, std::move(v)));
+			m_shortcut_map[k] = m_list.getFirstAddress();
+
+		}
+		else
+		{
+			m_list.get(iterator->second)->data = std::pair<Key, Value>(k, std::move(v));
+			m_list.moveToFront(iterator->second);
+		}
+	}
```
devshgraphicsprogramming: dont duplicate code, if you have a override with slightly different reference types, then pass through appropriately to a "master implemenetation".

src/irr/core/LRUcache.h
```diff
@@ -0,0 +1,131 @@
...
+		if (i != invalid_iterator)
+		{
+			m_list.moveToFront(i);
+			return m_list.get(i)->data.second;
+		}
+		return nullptr;
+	}
+
+	//get value without reordering the list
+	inline Value peek(Key& key)
```
devshgraphicsprogramming: peek should be `const` qualified, and so its argument

src/irr/core/LRUcache.h
```diff
@@ -0,0 +1,131 @@
...
+		{
+			std::cout << node->data.first << ' ' << node->data.second << std::endl;
+			if (node->next == invalid_iterator)
+				break;
+			node = m_list.get(node->next);
+		}
+	}
+#endif // _IRR_DEBUG
+
+	//add element to the cache, or move to the front if key exists. In case of the latter, it doesnt update the value
```
devshgraphicsprogramming: comment doesnt match behaviour, update the comment.

src/irr/core/LRUcache.h
devshgraphicsprogramming: return by reference, not by value.

src/irr/core/LRUcache.h
```diff
@@ -0,0 +1,131 @@
...
+
+		}
+		else
+		{
+			m_list.get(iterator->second)->data = std::pair<Key, Value>(k, std::move(v));
+			m_list.moveToFront(iterator->second);
+		}
+	}
+	//get the value from the cache, or null if key doesnt exist in cache or has been removed. Moves the element to the front of the cache.
+	inline Value get(Key& key)
```
devshgraphicsprogramming: return by reference not by value

devshgraphicsprogramming: also const and non-const reference return value overloads

src/irr/core/LRUcache.h
devshgraphicsprogramming: also const and non-const overloads

src/irr/core/LRUcache.h
```diff
@@ -0,0 +1,131 @@
...
+	}
+
+	//remove element at key if present
+	inline void erase(Key& key)
+	{
+		uint32_t i = common_peek(key);
+		if (i != invalid_iterator)
+		{
+			m_shortcut_map.erase(key);
+			m_list.erase(i);
```
devshgraphicsprogramming: `common_erase` ?

src/irr/core/LRUcache.h
devshgraphicsprogramming: P.S. Example I gave https://github.com/buildaworldnet/IrrlichtBAW/pull/487#discussion_r472068256

devshgraphicsprogramming: It would be possible to only have one copy of `Key`, it would need a bit of a hack.

You'd need to change `unordered_map<Key, uint32_t, MapHash, MapEquals>` to `unordered_set<uint32_t, HashWrapper<MapHash>, EqualsWrapper<MapEquals>>`

Then you'd reserve the `invalid_iterator` to mean "input key to the function" (making a huge assertion here that `m_shortcut_map` can never contain `invalid_iterator`).

What would happen is that you could define your own `struct HashWrapper` and `struct EqualsWrapper` which thanks to C++11 could actually be made stateful and store the reference to the `m_list`, as well as `const Key* invalid_iterator_key;` and `std::size_t invalid_iterator_hash;`, you'd construct the wrappers in the constructor of `LRUcache` and pass them into the constructor of `m_shortcut_map`.

Then whenever you call `m_shortcut_map.find`, you'd first set the pointer to `invalid_iterator_key` and compute the `invalid_iterator_hash` for the stateful wrappers.

Then whenever the `operator()` of either the hash or equals gets called:
- If the value is `invalid_iterator` return the relevant state's hash or use the state's pointer for the `operator==`
- If the value is not `invalid_iterator` then fetch the relevant node from `m_list` use it's pair's `Key` member for hash or compare

src/irr/core/LRUcache.h
Hazardu: yep

src/irr/core/CFixedCapacityDoublyLinkedList.h
Hazardu: was assigned 9 lines above

src/irr/core/LRUcache.h
Hazardu: i hope i got this right in the next commit

src/irr/core/CFixedCapacityDoublyLinkedList.h
devshgraphicsprogramming: TODO.

src/irr/core/CFixedCapacityDoublyLinkedList.h
```diff
@@ -61,13 +65,17 @@ namespace irr {
...
 		public:
-			_IRR_STATIC_INLINE_CONSTEXPR uint32_t invalid_iterator = 0xdeadbeefu;
+			_IRR_STATIC_INLINE_CONSTEXPR uint32_t invalid_iterator = PoolAddressAllocator<uint32_t>::invalid_address;
 
 			//get node at iterator
 			inline Snode<Value>* get(uint32_t address)
 			{
 				return (m_array+address);
 			}
+			inline Snode<Value>* get(const uint32_t address) const
```
devshgraphicsprogramming: its a const pointer that should be returned

src/irr/core/CFixedCapacityDoublyLinkedList.h
```diff
@@ -114,21 +125,21 @@ namespace irr {
 					get(node->prev)->next = node->next;
 				if (node->next != invalid_iterator)
 					get(node->next)->prev = node->prev;
-				//node->~Snode<Value>();	destructor not defined
+				node->~Snode<Value>();
```
devshgraphicsprogramming: waiting on that `common_delete` or something.

src/irr/core/LRUcache.h
devshgraphicsprogramming: waiting on that common_delete or something.

src/irr/core/LRUcache.h
```diff
@@ -63,65 +50,79 @@ class LRUcache
...
-	inline Value peek(Key& key)
+	//get the value from cache at an associated Key, or nullptr if Key is not contained within cache. Does not alter the value use order
+	inline Value* peek(Key& key) const
 	{
 		uint32_t i = common_peek(key);
 		if (i == invalid_iterator) return nullptr;
-		else return  m_list.get(i)->data.second;
+		else return &(m_list.get(i)->data.second);
+	}
+	inline const Value* peek(const Key& key) const
```
devshgraphicsprogramming: why are you returning pointers instead of references?

src/irr/core/LRUcache.h
devshgraphicsprogramming: TODO: I don't like having a duplicate `Key` lingering around.

Also another solution would be to swap out the `Key` in unordered map for a `Key*` in the `unordered_map` (don't need a weird `unordered_set` anymore) however the hash and equals would still need to be wrapped to dereference the pointer.

The `Key*` pointer would need to point to the `SNode<>.value.first` so its basically almost a duplicate of the `uint32_t` field, at which point we degenerate into the `unordered_set` solution.

src/irr/core/LRUcache.h
Hazardu: to be able to return a nullptr

src/irr/core/CFixedCapacityDoublyLinkedList.h
```diff
@@ -0,0 +1,161 @@
...
+			}
+			//Constructor, capacity determines the amount of allocated space
+			DoublyLinkedList(const uint32_t capacity) :
+				cap(capacity),
+				reservedSpace(_IRR_ALIGNED_MALLOC(PoolAddressAllocator<uint32_t>::reserved_size(1u, capacity, 1u), alignof(void*))),
+				alloc(reservedSpace, 0u, 0u, 1u, capacity, 1u)
+			{
+				m_back = invalid_iterator;
+				m_begin = invalid_iterator;
+				m_array = new Snode<Value>[capacity];
```
Crisspl, line 173: i think you'll have to allocate raw storage and create `Snode`s with placement-new, just like std vector does. Otherwise you're unable to store not-default-constructible types

Crisspl: @devshgraphicsprogramming ?

src/irr/core/LRUcache.h
Hazardu: I've changed the LRU cache to store `Key*` 😊👍

src/irr/core/LRUcache.h
devshgraphicsprogramming: ah ok, for invalid.

src/irr/core/CFixedCapacityDoublyLinkedList.h
```diff
@@ -119,15 +125,20 @@ namespace irr {
 					m_back = addr;
 				m_begin = addr;
 			}
+			//add new item to the list. This function does not make space to store the new node. in case the list is full, popBack() needs to be called beforehand
+			inline void pushFront(Value&& val)
+			{
+				insertAt(reserveAddress(), std::move(val))
```
devshgraphicsprogramming, line 131: why are `insertAt` and `reserveAddress` public?

src/irr/core/CFixedCapacityDoublyLinkedList.h
devshgraphicsprogramming: this should have been a malloc/free not a new/delete

PR #491, opened 2020-05-19
Envmap Example

AnastaZIuk:
status:

- need to display cube env another way


AnastaZIuk:
actually shouldn't we have .hdr loader/writer?

devshgraphicsprogramming:
> actually shouldn't we have .hdr loader/writer?
Nah its not really a popular format.

devshgraphicsprogramming:
I'm going to see if I can help tomorrow

PR #495, opened 2020-05-25
SSBO benchmark

Przemog1:


examples_tests/31.SSBOBenchmark/main.cpp
```diff
@@ -0,0 +1,401 @@
...
+
+    //get random tesselation for disks
+    {
+        std::random_device rd;
+        std::mt19937 mt(rd());
+        std::uniform_int_distribution<uint32_t> dist(15, 14999);
+
+        //TODO: test
+        //`dist(mt) | 0x0001` so vertexCount is always odd (only when diskCount is odd as well)
+        std::generate(tesselation.begin(), tesselation.end(), [&]() { return dist(mt) | 0x0001; });
```
devshgraphicsprogramming, line 273: formula for vertex count in a disk 1+triangleCount, which means that the number of triangles should be even to have odd numbers of vertices (and hence vertex shader invocations)

devshgraphicsprogramming: its 2+triangleCount if there's no vertex re-use of the vertex creating the seam at the circumference

examples_tests/31.SSBOBenchmark/main.cpp
```diff
@@ -0,0 +1,401 @@
...
+            {0u,EF_R32G32B32_SFLOAT,offsetof(Vertex,normal)}
+        },
+        {vertexSize,EVIR_PER_VERTEX} 
+    };
+
+
+    core::vector<core::matrix3x4SIMD> SSBOData; //bone matrices will also reside here
+    SSBOData.reserve(diskCount);
+
+    //create disks and join them into one mesh buffer
```
devshgraphicsprogramming, line 306: this I want you to separate into a utility class like we talked with @AnastaZIuk and @Crisspl

PR #505, opened 2020-06-15
Examples API correction

AnastaZIuk:
Is there something else I can correct?


examples_tests/09.ColorSpaceTest/ApplicationHandler.cpp
```diff
@@ -126,7 +126,7 @@ void ApplicationHandler::performImageTest(std::string path)
 	auto cpuTexture = assetManager->getAsset(path, lp);
 	auto cpuTextureContents = cpuTexture.getContents();
 	
-	if (cpuTextureContents.first == cpuTextureContents.second)
+	if (cpuTextureContents.begin() == cpuTextureContents.end())
```
devshgraphicsprogramming, line 129: SRange has a `size()` now

PR #508, opened 2020-06-25
Criss' Shader Changes

devshgraphicsprogramming:


include/irr/builtin/glsl/bxdf/brdf/specular/beckmann_smith.glsl
```diff
@@ -1,38 +1,22 @@
...
-
 irr_glsl_BSDFSample irr_glsl_beckmann_smith_cos_generate(in irr_glsl_AnisotropicViewSurfaceInteraction interaction, in vec2 _sample, in float ax, in float ay)
 {
     vec2 u = _sample;
     
     mat3 m = irr_glsl_getTangentFrame(interaction);
 
     vec3 localV = interaction.isotropic.V.dir;
-    localV = normalize(V*m);//transform to tangent space
+    localV = normalize(localV*m);//transform to tangent space
```
devshgraphicsprogramming, line 19: your tanget frame shall be orthonormal, localV should also be normalized! so no need for normalization.

PR #520, opened 2020-07-23
Indirect structs

Przemog1:


include/irr/asset/IMeshBuffer.h
```diff
@@ -42,6 +42,13 @@ struct DrawElementsIndirectCommand_t
     uint32_t baseInstance;
 };
 
+struct DispatchIndirectCommand_t
```
Crisspl: I'd move this to IComputePipeline.h or make separate utility header and move all 3 there

Crisspl: Probably separate header is best option

Przemog1: @devshgraphicsprogramming what do you think?

devshgraphicsprogramming: I second criss, this (all indirect structs) should move to `irr/asset/IPipeline.h`

Issue #460, opened 2020-04-13
Image Filters Bugs


AnastaZIuk:
### List is for gathering found bugs and documenting them. 

Map:

- _Unsolved_: bugs that still need to be fixed
- _Probability of bugs_: those ones are suspected of being bugs, but they are not confirmed

**If you found a bug or you suspect something to be a bug, put your description here in following way**:

- short description of the target
- error compilation code if it's a syntax error
- link to appropriate line on github

In case of probability of being a bug, you won't pass error compile code, but you have to describe in a detail what's wrong and what's going on.

_Unsolved_:

_Probability of bugs_:

1. Copying regions with multilayered and mipmaped images. ImageView data is filled with black value and that's what renderDoc sees. ImageView is returned from GLI Loader which passes region validation, but probably there is something wrong while performing `cpu2gpu` conversion somewhere in `copybuffertoimage`.

_Fixed_:
1. `CState *` to `const Swizzle &` conversion
2. https://github.com/Crisspl/IrrlichtBAW/blob/5274cf63c26491483f61215f351f255e49f3df2e/include/irr/asset/IImageAssetHandlerBase.h#L120
and
https://github.com/Crisspl/IrrlichtBAW/blob/5274cf63c26491483f61215f351f255e49f3df2e/include/irr/asset/IImageAssetHandlerBase.h#L191

devshgraphicsprogramming:
> Copying regions with multilayered and mipmaped images. ImageView data is filled with black value and that's what renderDoc sees. ImageView is returned from GLI Loader which passes region validation, but probably there is something wrong while performing cpu2gpu conversion somewhere in copybuffertoimage.

I need a concrete reproducible example

devshgraphicsprogramming:
`CState` bug fixed

AnastaZIuk:
Another one `CState *` to `const Swizzle &` conversion bug
```
Severity	Code	Description	Project	Path	File	Line	Suppression State
[...]

PR #521, opened 2020-07-25
Filters API

AnastaZIuk:
status:

**done**:

- PNG Fix
- clamping 
- precomputed dithering
- blue noise dithering in denoiser tonemapper example
- gli swizzle feature
- compile time utilities
- swizzle in blit filter


include/irr/asset/filters/CSwizzleAndConvertImageFilter.h
```diff
@@ -188,20 +188,50 @@ class CSwizzleAndConvertImageFilter<EF_UNKNOWN,EF_UNKNOWN,Swizzle> : public CIma
...
+
+					return buffer;
+				};
+
+				auto swizzle = [&commonExecuteData,&blockDims,inFormat,outFormat,&state,&getClampedSourcePixels](uint32_t readBlockArrayOffset, core::vectorSIMDu32 readBlockPos)
 				{
 					constexpr auto MaxPlanes = 4;
 					const void* srcPix[MaxPlanes] = { commonExecuteData.inData+readBlockArrayOffset,nullptr,nullptr,nullptr };
 
+					auto clampedBuffer = getClampedSourcePixels(srcPix); // todo - compile time, do not waste memory
```
devshgraphicsprogramming, line 221: you do not clamp on input, you clamp on output!

devshgraphicsprogramming: use `if constexpr` instead of a lambda, also clamp the pixels on output not input.

include/irr/asset/filters/CSwizzleAndConvertImageFilter.h
```diff
@@ -188,20 +188,50 @@ class CSwizzleAndConvertImageFilter<EF_UNKNOWN,EF_UNKNOWN,Swizzle> : public CIma
...
 			auto perOutputRegion = [&blockDims,inFormat,outFormat,&state](const CMatchedSizeInOutImageFilterCommon::CommonExecuteData& commonExecuteData, CBasicImageFilterCommon::clip_region_functor_t& clip) -> bool
 			{
-				auto swizzle = [&commonExecuteData,&blockDims,inFormat,outFormat,&state](uint32_t readBlockArrayOffset, core::vectorSIMDu32 readBlockPos)
+				const auto inTexelOrBlockByteSize = asset::getTexelOrBlockBytesize(inFormat);
+				const auto inChannelAmount = asset::getFormatChannelCount(inFormat);
+
+				// TESTING
+				auto getClampedSourcePixels = [&](auto sourcePixels)
+				{
+					constexpr auto MaxPlanes = 4;
```
devshgraphicsprogramming: is there a "plane count query" for `E_FORMAT` yet?

include/irr/asset/filters/CSwizzleAndConvertImageFilter.h
```diff
@@ -188,20 +188,50 @@ class CSwizzleAndConvertImageFilter<EF_UNKNOWN,EF_UNKNOWN,Swizzle> : public CIma
...
+					auto buffer = core::make_smart_refctd_ptr<asset::ICPUBuffer>(blockDims.y * blockDims.x * inTexelOrBlockByteSize);
```
devshgraphicsprogramming: you just want a `double pixelVals[4];` instead of the buffer.

`decodePixels` takes an intrablock X,Y coordinate  to decode one pixel at a time.

include/irr/asset/filters/CSwizzleAndConvertImageFilter.h
```diff
@@ -188,20 +188,50 @@ class CSwizzleAndConvertImageFilter<EF_UNKNOWN,EF_UNKNOWN,Swizzle> : public CIma
...
+						std::array<double, MaxPlanes> decodeBuffer = {};
 						auto localOutPos = readBlockPos*blockDims+commonExecuteData.offsetDifference;
 						uint8_t* dstPix = commonExecuteData.outData+commonExecuteData.oit->getByteOffset(localOutPos,commonExecuteData.outByteStrides);
+
 						if constexpr(std::is_base_of<PolymorphicSwizzle,Swizzle>::value)
-							convertColor<Swizzle>(inFormat,outFormat,srcPix,dstPix,blockX,blockY,state->swizzle);
+							convertColor<Swizzle>(inFormat,outFormat, newSourcePixels,dstPix,blockX,blockY,state->swizzle);
 						else
-							convertColor<Swizzle>(inFormat,outFormat,srcPix,dstPix,blockX,blockY,*state);
+							convertColor<Swizzle>(inFormat,outFormat, newSourcePixels,dstPix,blockX,blockY,*state);
```
devshgraphicsprogramming: instead of having that weird lambda at the start, you want to break down the `convertColor<>` functions into their implementation constituent `decodePixels` and `encodePixels` that use a `uint64_t[4]` (sometimes casted to a `double[4]`) buffer for temporary storage.

Then sometimes (when you have the template parameter) depending on a `if constexpr` you want to shove in a `clamp` between the decode and encode.

devshgraphicsprogramming: because you have the silly lambda you feel the need to allocate temporary storage for the block

include/irr/asset/filters/CSwizzleAndConvertImageFilter.h
```diff
@@ -188,20 +188,50 @@ class CSwizzleAndConvertImageFilter<EF_UNKNOWN,EF_UNKNOWN,Swizzle> : public CIma
...
+					for (auto blockY = 0u; blockY < blockDims.y; blockY++)
+						for (auto blockX = 0u; blockX < blockDims.x; blockX++)
+						{
+							std::array<double, MaxPlanes> decodeBuffer = {};
+							asset::decodePixelsRuntime(inFormat, sourcePixels, decodeBuffer.data(), blockX, blockY);
+
+							for (auto& value : decodeBuffer)
+								core::clamp<double, double>(value, 0, 1);
+
+							asset::encodePixelsRuntime(inFormat, reinterpret_cast<uint8_t*>(buffer->getPointer()) + (blockY * blockDims.x + blockDims.x) * inTexelOrBlockByteSize, decodeBuffer.data());
```
devshgraphicsprogramming: you're making an ungodly mess of it by decoding from INPUT_FORMAT ~~using the runtime function when we have compile time specializations which are used in `convertColor`~~ (actually you havent provided compile-time clamping at all for the compile-time format conversion from the general specialization of this class on lines 144-160), clamping, then converting to OUTPUT_FORMAT and saving to a buffer, then REINTERPRETING THE SAVED DATA as INPUT_FORMAT and having `convertColor` do the conversion to OUTPUT_FORMAT` AGAIN

include/irr/asset/filters/CSwizzleAndConvertImageFilter.h
devshgraphicsprogramming: all my comments basically boil down to:
- see how `convertColor<>` works
- decompose it into a separate decode+encode
- put a clamp between the decode and the encode depending on the constexpr bool template parameter

include/irr/asset/filters/CSwizzleAndConvertImageFilter.h
```diff
@@ -188,20 +188,50 @@ class CSwizzleAndConvertImageFilter<EF_UNKNOWN,EF_UNKNOWN,Swizzle> : public CIma
...
+					const void* newSourcePixels[MaxPlanes] = { buffer->getPointer(), nullptr, nullptr, nullptr };
+
+					for (auto blockY = 0u; blockY < blockDims.y; blockY++)
+						for (auto blockX = 0u; blockX < blockDims.x; blockX++)
+						{
+							std::array<double, MaxPlanes> decodeBuffer = {};
+							asset::decodePixelsRuntime(inFormat, sourcePixels, decodeBuffer.data(), blockX, blockY);
+
+							for (auto& value : decodeBuffer)
+								core::clamp<double, double>(value, 0, 1);
```
devshgraphicsprogramming: clamp to output's format Minimum Value in the Range, and Maximum Value in the Range

NOT 0 and 1

P.S. also not all formats encode from `double`, so there's that to account for... see impl of `convertColor`

include/irr/asset/filters/CSwizzleAndConvertImageFilter.h
```diff
@@ -188,20 +188,50 @@ class CSwizzleAndConvertImageFilter<EF_UNKNOWN,EF_UNKNOWN,Swizzle> : public CIma
...
+					auto buffer = core::make_smart_refctd_ptr<asset::ICPUBuffer>(blockDims.y * blockDims.x * inTexelOrBlockByteSize);
+					const void* newSourcePixels[MaxPlanes] = { buffer->getPointer(), nullptr, nullptr, nullptr };
+
+					for (auto blockY = 0u; blockY < blockDims.y; blockY++)
+						for (auto blockX = 0u; blockX < blockDims.x; blockX++)
+						{
+							std::array<double, MaxPlanes> decodeBuffer = {};
+							asset::decodePixelsRuntime(inFormat, sourcePixels, decodeBuffer.data(), blockX, blockY);
+
+							for (auto& value : decodeBuffer)
```
devshgraphicsprogramming: I'd do a loop from i=0 to output format channel count, if the output format is not `EF_UNDEFINED` (not runtime specified)

from i=0 to 0 otherwise

include/irr/asset/filters/CSwizzleAndConvertImageFilter.h
```diff
@@ -164,8 +164,8 @@ class CSwizzleAndConvertImageFilter : public CImageFilter<CSwizzleAndConvertImag
 		}
 };
 
-template<typename Swizzle>
-class CSwizzleAndConvertImageFilter<EF_UNKNOWN,EF_UNKNOWN,Swizzle> : public CImageFilter<CSwizzleAndConvertImageFilter<EF_UNKNOWN,EF_UNKNOWN,Swizzle>>, public impl::CSwizzleAndConvertImageFilterBase<Swizzle>
+template<typename Swizzle, bool clamp>
+class CSwizzleAndConvertImageFilter<EF_UNKNOWN,EF_UNKNOWN,Swizzle,clamp> : public CImageFilter<CSwizzleAndConvertImageFilter<EF_UNKNOWN,EF_UNKNOWN,Swizzle,clamp>>, public impl::CSwizzleAndConvertImageFilterBase<Swizzle>
```
devshgraphicsprogramming: you've only implemented the clamping for runtime (or at least 1-format-runtime, because we dont bother specializing for half-compile-time cases) format conversion.

I need an implementation for the both-formats-specified-compile-time case, which is the general class above, where `swizzle` lambda is defined between line 144 and 160

include/irr/asset/filters/CSwizzleAndConvertImageFilter.h
devshgraphicsprogramming: for this class a `for (auto i=0; i<MaxChannels; i++)` is desired, for the compile-time general case
> 
> 
> you've only implemented the clamping for runtime (or at least 1-format-runtime, because we dont bother specializing for half-compile-time cases) format conversion.
> 
> I need an implementation for the both-formats-specified-compile-time case, which is the general class above, where `swizzle` lambda is defined between line 144 and 160

I want the for from 0 to output format's channel count (given by the template function, not runtime function)

include/irr/asset/filters/CSwizzleAndConvertImageFilter.h
AnastaZIuk: adjusted in following commit

AnastaZIuk: wait, probably got one thing wrong, but I will show it when pushed

include/irr/asset/filters/CSwizzleAndConvertImageFilter.h
AnastaZIuk: there is not as I look around

include/irr/asset/filters/CSwizzleAndConvertImageFilter.h
AnastaZIuk: adjusted in following commit

include/irr/asset/filters/CSwizzleAndConvertImageFilter.h
AnastaZIuk: adjusted in following commit

include/irr/asset/filters/CSwizzleAndConvertImageFilter.h
AnastaZIuk: adjusted for runtime in following commit, compile time left

include/irr/asset/filters/CSwizzleAndConvertImageFilter.h
AnastaZIuk: I was testing something, but ure right, adjusted in following commit (runtime)

include/irr/asset/filters/CSwizzleAndConvertImageFilter.h
AnastaZIuk: it isn't valid anymore

include/irr/asset/filters/CSwizzleAndConvertImageFilter.h
AnastaZIuk: solved

examples_tests/39.DenoiserTonemapper/main.cpp
```diff
@@ -990,8 +990,8 @@ void main()
 
 			auto getConvertedPNGImageView = [&](core::smart_refctd_ptr<ICPUImage> image)
 			{
-				using CONVERSION_FILTER = CConvertFormatImageFilter<EF_UNKNOWN, EF_UNKNOWN, true>; // TESTING
 				constexpr auto pngFormat = EF_R8G8B8A8_SRGB;
```
devshgraphicsprogramming: drop the alpha, use R8G8B8_SRGB

include/irr/asset/filters/CSwizzleAndConvertImageFilter.h
devshgraphicsprogramming: make one ;)

include/irr/asset/filters/CSwizzleAndConvertImageFilter.h
```diff
@@ -126,22 +130,31 @@ class CSwizzleAndConvertImageFilter : public CImageFilter<CSwizzleAndConvertImag
...
 				assert(blockDims.w==1u);
 			#endif
+
+			typedef std::conditional<asset::isIntegerFormat<inFormat>(), uint64_t, double>::type decodeBufferType;
+			typedef std::conditional<asset::isIntegerFormat<outFormat>(), uint64_t, double>::type encodeBufferType;
+
 			auto perOutputRegion = [&blockDims,&state](const CMatchedSizeInOutImageFilterCommon::CommonExecuteData& commonExecuteData, CBasicImageFilterCommon::clip_region_functor_t& clip) -> bool
 			{
-				auto swizzle = [&commonExecuteData,&blockDims,&state](uint32_t readBlockArrayOffset, core::vectorSIMDu32 readBlockPos)
+				const auto clampBufferChannelsAmount = asset::getFormatChannelCount(outFormat);
```
devshgraphicsprogramming: is there a compile time version ? like `asset:getFormatChannelCount<E_FORMAT format>(void)` ?

devshgraphicsprogramming:
Ok now only Blue Noise Dithering left to do and we're ready to merge ;)

include/irr/asset/filters/CSwizzleAndConvertImageFilter.h
```diff
@@ -151,10 +164,37 @@ class CSwizzleAndConvertImageFilter : public CImageFilter<CSwizzleAndConvertImag
...
+							{
+								auto&& [min, max, encodeValue] = std::make_tuple<encodeBufferType&&, encodeBufferType&&, encodeBufferType*>(asset::getFormatMinValue<encodeBufferType>(outFormat, i), asset::getFormatMaxValue<encodeBufferType>(outFormat, i), reinterpret_cast<encodeBufferType*>(encodeBuffer) + i);
+								*encodeValue = core::clamp(*encodeValue, min, max);
+							}
+						};
+
+						if constexpr (clamp)
+							clampBuffer();
+							
+						asset::encodePixels<outFormat>(dstPix, reinterpret_cast<encodeBufferType*>(encodeBuffer));
```
AnastaZIuk: as I think of it, shouldn't I move a pointer to be a `dstPix+= blockSize` each iteration in block loop?

AnastaZIuk: okay, we do not support encoding to BC, it clears a little

devshgraphicsprogramming: should still advance the pointer though XD ;)

AnastaZIuk: yup, as well as I do it on SAT now

include/irr/asset/filters/CSwizzleAndConvertImageFilter.h
devshgraphicsprogramming: ?

AnastaZIuk: oh sorry, no there isn't

devshgraphicsprogramming: make one.

AnastaZIuk: roger that!

include/irr/asset/filters/CSwizzleAndConvertImageFilter.h
AnastaZIuk: corrected in following commit

examples_tests/39.DenoiserTonemapper/main.cpp
AnastaZIuk: corrected in following commit

include/irr/asset/filters/CSwizzleAndConvertImageFilter.h
AnastaZIuk: added and in use!

include/irr/asset/filters/dithering/CDither.h
```diff
@@ -0,0 +1,47 @@
...
+					return CRTP::get(pixelCoord);
+				}
+
+				class CState : public IDither::IState
+				{
+					public:
+						CState() {}
+						virtual ~CState() {}
+
+						ImageAsset imageAsset; // do we want to only handle ICPUImage or should I distinguish it to ICPUImageView?
```
AnastaZIuk: we should make a decision about it

include/irr/asset/filters/dithering/IDither.h
```diff
@@ -0,0 +1,36 @@
...
+                virtual ~IDither() {}
+                virtual float pGet(const core::vectorSIMDu32& pixelCoord) = 0;
+
+                class IState
+                {
+                    virtual ~IState() {}
+
+                    struct ImageAsset
+                    {
+                        core::smart_refctd_ptr<asset::IAsset> asset;
```
AnastaZIuk: it relates to the above https://github.com/buildaworldnet/IrrlichtBAW/pull/521#discussion_r475792929, I will code further if we decide

AnastaZIuk:
I think we should move `IDither` to asset directory

include/irr/asset/filters/dithering/IDither.h
AnastaZIuk: not sure if we use dds dithering images or whatever in future

include/irr/asset/filters/dithering/CDither.h
```diff
@@ -0,0 +1,47 @@
...
+				};
+
+				using state_type = CState;
+
+			private:
+			
+				float get(const state_type* state, const core::vectorSIMDu32& pixelCoord)
+				{
+					// use blue noise tile ability to fetch suitable texel pointer from texture
```
devshgraphicsprogramming: put a `return CRTP::get` here, this way its clear we expect the derived class to implement (also make function a static)

include/irr/asset/filters/dithering/CDither.h
devshgraphicsprogramming: dont hold this here, remember `CDither` is a base class for many things like:
- CWhiteNoiseDither
- CBayerMatrixDither
- CPrecomputedDither

include/irr/asset/filters/dithering/IDither.h
devshgraphicsprogramming: dont worry about this in the base class, only provide maximum extent that the dither must support.

include/irr/asset/filters/dithering/IDither.h
```diff
@@ -0,0 +1,36 @@
...
+
+namespace irr
+{
+    namespace asset
+    {
+        class IDither
+        {
+            public:
+                virtual ~IDither() {}
+                virtual float pGet(const core::vectorSIMDu32& pixelCoord) = 0;
```
devshgraphicsprogramming: you need a pointer to `IState` as another parameter

include/irr/asset/filters/dithering/CPrecomputedDither.h
```diff
@@ -0,0 +1,71 @@
...
+					@param state Input state
+					@param pixelCoord Current pixel coordinate of processing input image
+					we will be applying dithering to
+				*/
+
+				static float get(const state_type* state, const core::vectorSIMDu32& pixelCoord)
+				{
+					const auto& imageData = state->getImageData();
+
+					// if it has to be float, we should change something there cuz we aren't able to fetch the value due to channels
```
AnastaZIuk: an issue!

include/irr/asset/filters/CSwizzleAndConvertImageFilter.h
```diff
@@ -150,11 +169,52 @@ class CSwizzleAndConvertImageFilter : public CImageFilter<CSwizzleAndConvertImag
...
+							- fetch channel's texel dither img value
+
+							encodeBuffer[x] += dither img value * scale
+
+							where scale is: asset::getFormatPrecision<value_type>(outFormat,channel,unditheredValue)
+							and dither img value is:
+						*/
+
+						// too few info to do it - CHANNELS!
+						const float ditheredValue = blueNoise.dithering.pGet(&blueNoise.state, localOutPos + core::vectorSIMDu32(blockX, blockY));
```
AnastaZIuk: an issue! https://github.com/buildaworldnet/IrrlichtBAW/pull/521/files#r476523336

include/irr/asset/filters/CSwizzleAndConvertImageFilter.h
```diff
@@ -150,11 +149,18 @@ class CSwizzleAndConvertImageFilter : public CImageFilter<CSwizzleAndConvertImag
 			typedef std::conditional<asset::isIntegerFormat<inFormat>(), uint64_t, double>::type decodeBufferType;
 			typedef std::conditional<asset::isIntegerFormat<outFormat>(), uint64_t, double>::type encodeBufferType;
 
-			auto perOutputRegion = [&blockDims,&state](const CMatchedSizeInOutImageFilterCommon::CommonExecuteData& commonExecuteData, CBasicImageFilterCommon::clip_region_functor_t& clip) -> bool
+			struct
+			{
+				using BLUE_NOISE_DITHER = asset::CPrecomputedDither;
+				BLUE_NOISE_DITHER dithering;
+				BLUE_NOISE_DITHER::state_type state;  /* how should I use asset manager there in constructor? i dont have an access */
+			} blueNoise;
```
devshgraphicsprogramming: make the dither type a template parameter and keep a `Dither*` pointer to the object in the state...

then obvisously you create the dither before you call the filter  ;)

include/irr/asset/IDither.h
```diff
@@ -0,0 +1,52 @@
...
+
+                        struct ImageData
+                        {
+                            asset::ICPUBuffer* buffer;
+                            core::vectorSIMDu32 strides;
+                            core::vectorSIMDu32 extent;
+                        };
+                };
+
+                virtual float pGet(const IState* state, const core::vectorSIMDu32& pixelCoord) = 0;
```
devshgraphicsprogramming: add an `int32_t channel` parameter,

include/irr/asset/filters/dithering/CBayerMatrixDither.h
```diff
@@ -0,0 +1,36 @@
...
+						virtual ~CState() {}
+				};
+
+				using state_type = CState;
+			
+				static float get(const state_type* state, const core::vectorSIMDu32& pixelCoord)
+				{
+					// TODO: to define in future
+				}
+		};
```
devshgraphicsprogramming:
```
#if 0 // TODO
#endif
```
don't need it right now

include/irr/asset/filters/dithering/CDither.h
```diff
@@ -5,40 +5,52 @@
...
 	namespace asset
 	{
+		//! Base CRTP class for dithering classes
+		/*
+			There are several dithering classes:
+
+			- CWhiteNoiseDither
+			- CBayerMatrixDither
+			- CPrecomputedDither
+			- CHashDither
```
devshgraphicsprogramming: white noise is hash dither.

include/irr/asset/filters/dithering/CDither.h
```diff
@@ -5,40 +5,52 @@
...
-				}
 
 				class CState : public IDither::IState
 				{
 					public:
 						CState() {}
 						virtual ~CState() {}
 
-						ImageAsset imageAsset; // do we want to only handle ICPUImage or should I distinguish it to ICPUImageView?
+						ImageData imageData;
```
devshgraphicsprogramming: only precomputed keeps image data

include/irr/asset/filters/dithering/CDither.h
```diff
@@ -5,40 +5,52 @@
...
 
+				float pGet(const state_type* state, const core::vectorSIMDu32& pixelCoord) final override
+				{
+					return get(state, pixelCoord);
+				}
+
 			private:
 			
-				float get(const state_type* state, const core::vectorSIMDu32& pixelCoord)
+				static float get(const state_type* state, const core::vectorSIMDu32& pixelCoord)
```
devshgraphicsprogramming: can be public

include/irr/asset/IDither.h
```diff
@@ -0,0 +1,52 @@
...
+                    Holds certain image data which makes a user
+                    able to dither the image
+                */
+
+                class IState
+                {
+                    public:
+                        virtual ~IState() {}
+
+                        struct ImageData
```
devshgraphicsprogramming: no need for this here.

devshgraphicsprogramming: goes to precomputed

include/irr/asset/filters/dithering/CPrecomputedDither.h
devshgraphicsprogramming: add a `int32_t channel parameter`

include/irr/asset/filters/dithering/CBayerMatrixDither.h
AnastaZIuk: corrected in following commit

include/irr/asset/filters/dithering/CDither.h
AnastaZIuk: corrected in following commit

include/irr/asset/IDither.h
AnastaZIuk: corrected in following commit

include/irr/asset/filters/dithering/IDither.h
AnastaZIuk: solved

include/irr/asset/filters/dithering/IDither.h
AnastaZIuk: okay, `IState` will only hold `TexelRange` now (we fill add more in future if needed)
`imageData` will be moved to precomputed dither class and will contain only strides and the buffer

include/irr/asset/filters/CSwizzleAndConvertImageFilter.h
AnastaZIuk: corrected

include/irr/asset/filters/dithering/CPrecomputedDither.h
AnastaZIuk: corrected

include/irr/asset/filters/dithering/CDither.h
AnastaZIuk: I did it wrong (compile time error), need to read probably more about CRTP

include/irr/asset/filters/dithering/CDither.h
AnastaZIuk: corrected

include/irr/asset/IDither.h
AnastaZIuk: corrected

include/irr/asset/filters/dithering/CDither.h
AnastaZIuk: corrected

include/irr/asset/filters/dithering/IDither.h
devshgraphicsprogramming: why should it have a member with texel range?

include/irr/asset/filters/dithering/CDither.h
devshgraphicsprogramming: you can't just do `CRTP::function` if `function` is not static

if its not static you need to do `static_cast<CRTP*>(this)->function`

include/irr/asset/filters/dithering/IDither.h
AnastaZIuk: because we wanted to have max extent the dither support, so likely texel range I think, I can limit it to only extent

include/irr/asset/filters/dithering/CDither.h
AnastaZIuk: done

include/irr/asset/filters/dithering/CDither.h
AnastaZIuk: not an issue now

AnastaZIuk:
@devshgraphicsprogramming I'm done

examples_tests/39.DenoiserTonemapper/main.cpp
```diff
@@ -14,6 +14,9 @@ using namespace irr;
 using namespace asset;
 using namespace video;
 
+//! Comment if you want to apply blue noise on generated png files
+#define BLUE_NOISE_ON_PNG
```
devshgraphicsprogramming: assume always true ;)

examples_tests/39.DenoiserTonemapper/main.cpp
```diff
@@ -1024,7 +1018,17 @@ void main()
 					CONVERSION_FILTER convertFilter;
 					CONVERSION_FILTER::state_type state;
 					
-					state.ditherState = _IRR_NEW(std::remove_pointer<decltype(state.ditherState)>::type, ditheringImage);
+					#ifdef BLUE_NOISE_ON_PNG
+					auto ditheringBundle = am->getAsset("../../media/blueNoiseDithering/LDR_RGBA.png", {});
```
devshgraphicsprogramming, line 1020: load dither from EXR... you'll get better dither

include/irr/asset/filters/dithering/CDither.h
```diff
@@ -47,7 +47,7 @@ namespace irr
 				
 				float get(const IDither::IState* state, const core::vectorSIMDu32& pixelCoord, const int32_t& channel)
 				{
-					return static_cast<CRTP*>(this)->get(reinterpret_cast<const typename CRTP::CState*>(state), pixelCoord, channel);
+					return static_cast<CRTP*>(this)->get(dynamic_cast<const typename CRTP::CState*>(state), pixelCoord, channel);
```
devshgraphicsprogramming: `static_cast`

include/irr/asset/filters/dithering/CPrecomputedDither.h
```diff
@@ -52,12 +52,12 @@ namespace irr
...
 						struct
 						{
-							asset::ICPUBuffer* buffer;
+							const asset::ICPUBuffer* buffer;
 							core::vectorSIMDu32 strides;
 							asset::E_FORMAT format;
 						} ditherImageData;
 
-						core::smart_refctd_ptr<asset::ICPUImage> ditheringImage;
+						const asset::ICPUImage* const ditheringImage;
```
devshgraphicsprogramming: you don't need to keep it around after you construct the state.

AnastaZIuk: Look at how stare gets constructed, I think its okay

AnastaZIuk: state

devshgraphicsprogramming: you dont want to keep the image around (regions, etc.) just its contents.

examples_tests/39.DenoiserTonemapper/main.cpp
AnastaZIuk: corrected in following commit

include/irr/asset/filters/dithering/CDither.h
AnastaZIuk: corrected in following commit

include/irr/asset/filters/dithering/CPrecomputedDither.h
AnastaZIuk: corrected in following commit

examples_tests/39.DenoiserTonemapper/main.cpp
AnastaZIuk: not an issue anymore - discussed on discord

include/irr/asset/filters/CBlitImageFilter.h
```diff
@@ -174,13 +174,26 @@ class CBlitImageFilter : public CImageFilter<CBlitImageFilter<KernelX,KernelX,Ke
...
+				} oldMaxValue, oldMinValue;
+				
 				uint32_t				inMipLevel = 0u;
 				uint32_t				outMipLevel = 0u;
 				ICPUImage*				inImage = nullptr;
 				ICPUImage*				outImage = nullptr;
 				KernelX					kernelX;
 				KernelY					kernelY;
 				KernelZ					kernelZ;
+				asset::DefaultSwizzle defaultSwizzle;
```
devshgraphicsprogramming: the swizzle type should be a template, because there are compile time swizzles and runtime polymorphic swizzles (see our other APIs)

devshgraphicsprogramming: also you need an input swizzle and an output swizzle

include/irr/asset/filters/CBlitImageFilter.h
```diff
@@ -416,6 +429,10 @@ class CBlitImageFilter : public CImageFilter<CBlitImageFilter<KernelX,KernelX,Ke
 								auto sample = lineBuffer+i*MaxChannels;
 								decodePixels<value_type>(inFormat,srcPix,sample,inBlockCoord.x,inBlockCoord.y);
 
+								value_type swizzledSample[MaxChannels];
+								state->defaultSwizzle(sample, swizzledSample, MaxChannels);
+								memcpy(sample, swizzledSample, sizeof(value_type) * MaxChannels);
```
devshgraphicsprogramming: we don't need to memcpy, I think `decodePixels` has a variant that takes a swizzle.

include/irr/asset/filters/CSwizzleAndConvertImageFilter.h
```diff
@@ -150,11 +177,52 @@ class CSwizzleAndConvertImageFilter : public CImageFilter<CSwizzleAndConvertImag
...
+						};
+
+						const Swizzle& swizzle = getSwizzle();
+
+						constexpr auto maxChannels = 4;
+						uint64_t decodeBuffer[maxChannels] = {};
+						uint64_t encodeBuffer[maxChannels] = {};
+
+						asset::decodePixels<inFormat>(srcPix, reinterpret_cast<decodeBufferType*>(decodeBuffer), blockX, blockY);
+						swizzle.template operator()<void, void>(decodeBuffer, encodeBuffer);
```
devshgraphicsprogramming: see how swizzle gets implemented?

include/irr/asset/filters/dithering/CDither.h
```diff
@@ -0,0 +1,56 @@
...
+				CDither() {}
+				virtual ~CDither() {}
+
+				class CState : public IDither::IState
+				{
+					public:
+						CState() {}
+						virtual ~CState() {}
+
+						TexelRange texelRange;
```
devshgraphicsprogramming, line 38: what is texel range?

include/irr/asset/filters/CBlitImageFilter.h
AnastaZIuk: I don't get, we do i need in swizzle and out swizzle if filter producent only one out image
How shoould i process in?

include/irr/asset/filters/dithering/CPrecomputedDither.h
```diff
@@ -0,0 +1,89 @@
...
+					which buffer will be used in dithering process
+					in full extent of image.
+				*/
+				
+				class CState : public CDither::CState
+				{
+					public:
+						CState(const asset::ICPUImage* const ditheringImage) 
+						{
+							ditherImageData.buffer = ditheringImage->getBuffer();
```
devshgraphicsprogramming: you can't just take over a buffer of an image (even if you could you should have constructed a smartrefctd ptr)

because the buffer might not be there or cover too small of an area.

You need the CopyImageToBuffer filter IIRC.

include/irr/asset/filters/CSwizzleAndConvertImageFilter.h
AnastaZIuk: ?

include/irr/asset/filters/dithering/CPrecomputedDither.h
devshgraphicsprogramming: or create a proxy image with its own buffer and flatten the input image into that.

include/irr/asset/filters/dithering/CPrecomputedDither.h
```diff
@@ -0,0 +1,89 @@
...
+					const size_t offset = asset::IImage::SBufferCopy::getLocalByteOffset(tiledPixelCoord, ditherImageData.strides);
+
+					const auto* dstPointer = reinterpret_cast<const uint8_t*>(ditherImageData.buffer->getPointer()) + offset;
+					const void* srcPointers[] = { dstPointer, nullptr, nullptr, nullptr };
+
+					constexpr uint8_t MAX_CHANNELS = 4;
+					double decodeBuffer[MAX_CHANNELS];
+					asset::decodePixelsRuntime(ditherImageData.format, srcPointers, decodeBuffer, 0, 0); // little slow
+
+					return static_cast<float>(decodeBuffer[channel]);
```
devshgraphicsprogramming: nope, the values should have already been decoded.

include/irr/asset/filters/dithering/CPrecomputedDither.h
AnastaZIuk: what?

AnastaZIuk: do we have such?

include/irr/asset/filters/CBlitImageFilter.h
devshgraphicsprogramming: Blit takes an input image and puts it in the output image, while applying convolution kernels (or other kernels)

If you can swizzle, then you should be able to swizzle before and after (after decode and before encode) the kernel.

include/irr/asset/filters/CSwizzleAndConvertImageFilter.h
devshgraphicsprogramming: this is the "correct" way to implement Swizzle

include/irr/asset/filters/dithering/CPrecomputedDither.h
AnastaZIuk: ?

include/irr/asset/filters/dithering/CPrecomputedDither.h
devshgraphicsprogramming: https://github.com/buildaworldnet/IrrlichtBAW/blob/shader_pipeline/include/irr/asset/filters/CFlattenRegionsImageFilter.h

Should be a flatten & convert or you should just only accept images in the RGBA UNORM16 or UNORM32 formats

include/irr/asset/filters/dithering/CDither.h
AnastaZIuk: i will change it to extent maybe?

include/irr/asset/filters/CBlitImageFilter.h
AnastaZIuk: You would not to force me to split what's inside decodepixels otherwise in convert swizzle  filter then

include/irr/asset/filters/CSwizzleAndConvertImageFilter.h
```diff
@@ -28,13 +27,17 @@ namespace impl
...
 class CSwizzleAndConvertImageFilterBase : public CMatchedSizeInOutImageFilterCommon
 {
 	public:
 
 		class CState : public CMatchedSizeInOutImageFilterCommon::state_type, public Swizzle
 		{
+			public:
+				Dither dither;
+				using DitherState = typename Dither::state_type;
+				DitherState* ditherState = nullptr;					//! To use dithering in the filter - allocate, create and make the dither state pointer valid
```
devshgraphicsprogramming: there should be some sort of conditional (`enable_if`) declaration of these member variables

Basically if I'd set `Dither=void` it means no dither decided at compile time.

include/irr/asset/filters/CBlitImageFilter.h
AnastaZIuk: but i will check it

include/irr/asset/filters/CSwizzleAndConvertImageFilter.h
```diff
@@ -103,45 +117,58 @@ inline void DefaultSwizzle::operator()(const InT* in, OutT* out) const
...
 
-// do a per-pixel recombination of image channels while converting
-template<E_FORMAT inFormat=EF_UNKNOWN, E_FORMAT outFormat=EF_UNKNOWN, typename Swizzle=DefaultSwizzle>
-class CSwizzleAndConvertImageFilter : public CImageFilter<CSwizzleAndConvertImageFilter<inFormat,outFormat,Swizzle>>, public impl::CSwizzleAndConvertImageFilterBase<Swizzle>
+//! Compile time CSwizzleAndConvertImageFilter
+/*
+	Do a per-pixel recombination of image channels while converting
+*/
+
+template<E_FORMAT inFormat=EF_UNKNOWN, E_FORMAT outFormat=EF_UNKNOWN, typename Swizzle=DefaultSwizzle, bool Clamp = false, typename Dither = asset::CPrecomputedDither>
```
devshgraphicsprogramming: default dither `void` a.k.a. no dither.

include/irr/asset/filters/CSwizzleAndConvertImageFilter.h
devshgraphicsprogramming: see how there's a separate state member for `PolymorphicSwizzle` and there isn't when the swizzle is not polymorphic

include/irr/asset/filters/CSwizzleAndConvertImageFilter.h
AnastaZIuk: okay

include/irr/asset/filters/dithering/CDither.h
devshgraphicsprogramming: oh ok so its a measure of dimension?

AnastaZIuk: It's extent + offset field

devshgraphicsprogramming: ah kwl

devshgraphicsprogramming: you only need extent to be honest... and also not sure what its doing in the CDither class

include/irr/asset/filters/CBlitImageFilter.h
AnastaZIuk: `decodePixels` doesn't have such variant - `convertColor` does, but we don't want it

AnastaZIuk: not an issue anymore

include/irr/asset/filters/CSwizzleAndConvertImageFilter.h
AnastaZIuk: done

examples_tests/39.DenoiserTonemapper/main.cpp
```diff
@@ -4,6 +4,7 @@
 #include <irrlicht.h>
 
 #include "CommandLineHandler.hpp"
+#include "../../include/irr/asset/filters/dithering/CPrecomputedDither.h"
```
devshgraphicsprogramming: this should already be included in `irr/asset/asset.h"

AnastaZIuk: corrected

include/irr/asset/filters/dithering/CWhiteNoiseDither.h
```diff
@@ -21,20 +19,46 @@ namespace irr
...
 					public:
-						CState() {}
+						CState()
+							: sampler(std::chrono::high_resolution_clock::now().time_since_epoch().count())
+						{
+						
+						}
 						virtual ~CState() {}
+
+						core::RandomSampler sampler;
```
devshgraphicsprogramming: nope, you dont want a random number generator to implement this... you want a hash function and feed it the `pixelCoord` and `channel` as input to permute into a single 32 bit value (that you then divide by ~0u to get a float in the correct range)

AnastaZIuk:
**CBlitFilter** has worked with new API recently correctly without issues but I wonder if we should query the `seed` for wang hash in the way like I do it now (maybe we should change it to something like `std::chrono::high_resolution_clock::now().time_since_epoch().count()` or similar?)

AnastaZIuk:
spotted an issue however!
image
it seems it isn't due to dithering - when using Identity dither it generates the same weird output on bottom and right 
EDIT: it's probably a bug https://github.com/buildaworldnet/IrrlichtBAW/issues/460#issuecomment-683741305

include/irr/asset/filters/dithering/CWhiteNoiseDither.h
AnastaZIuk: solved

include/irr/asset/filters/CSwizzleAndConvertImageFilter.h
AnastaZIuk: there is a IdentityDither class now according to DC conversation and it works fine (at compile time)

include/irr/asset/filters/dithering/CPrecomputedDither.h
AnastaZIuk: done, but not sure if I should still force RGBA UNORM16 or UNORM32 formats

include/irr/asset/filters/CBlitImageFilter.h
```diff
@@ -192,17 +197,15 @@ class CBlitImageFilter : public CImageFilter<CBlitImageFilter<KernelX,KernelX,Ke
 				KernelX					kernelX;
 				KernelY					kernelY;
 				KernelZ					kernelZ;
-				asset::DefaultSwizzle defaultSwizzle;
 				bool normalize = false;
```
devshgraphicsprogramming: compile time template parameter would be better

AnastaZIuk: everything is now templated

devshgraphicsprogramming:
> 
> 
> **CBlitFilter** has worked with new API recently correctly without issues but I wonder if we should query the `seed` for wang hash in the way like I do it now (maybe we should change it to something like `std::chrono::high_resolution_clock::now().time_since_epoch().count()` or similar?)

`seed` should be in the whitenoise dither's state

AnastaZIuk:
> > **CBlitFilter** has worked with new API recently correctly without issues but I wonder if we should query the `seed` for wang hash in the way like I do it now (maybe we should change it to something like `std::chrono::high_resolution_clock::now().time_since_epoch().count()` or similar?)
> 
> `seed` should be in the whitenoise dither's state

okay

include/irr/asset/filters/CBlitImageFilter.h
AnastaZIuk: oh u mean `normalize`

AnastaZIuk: so what kind of classes should I pass to `normalize`? or should it be only a bool in template param?

AnastaZIuk: and it would do the same SAT's stuff

include/irr/asset/filters/dithering/CPrecomputedDither.h
```diff
@@ -58,11 +102,15 @@ namespace irr
 
 					private:
 
+						static constexpr auto decodeFormat = EF_R64G64B64A64_SFLOAT;
```
devshgraphicsprogramming, line 105: if the definition of `get` is `float` then your decode format is 32-bit float not 64bit ;)

include/irr/asset/filters/dithering/CPrecomputedDither.h
```diff
@@ -26,30 +26,74 @@ namespace irr
...
 					which buffer will be used in dithering process
-					in full extent of image.
+					in full extent of image. Dithering will be applied
+					using first 0th mipmap and layer either.
 				*/
 				
 				class CState : public CDither::CState
 				{
 					public:
 						CState(const asset::ICPUImage* const ditheringImage)
```
devshgraphicsprogramming, line 36: maybe take an `ICPUImageView` then one has a choice (indirectly) of the mip-map and layer that could be used for the dither.

include/irr/asset/filters/dithering/CWhiteNoiseDither.h
```diff
@@ -29,12 +32,28 @@ namespace irr
...
+						seed = (seed ^ 61) ^ (seed >> 16);
+						seed *= 9;
+						seed = seed ^ (seed >> 4);
+						seed *= 0x27d4eb2d;
+						seed = seed ^ (seed >> 15);
+						return seed;
+					};
+					
+					const auto hash = static_cast<double>(getWangHash());
+					return static_cast<float>(static_cast<double>(hash) / double(~0u));
```
devshgraphicsprogramming: your wang hash is done in 32bit uint, your return type is `float`... there's no need for using `double`s

include/irr/asset/filters/dithering/CWhiteNoiseDither.h
```diff
@@ -29,12 +32,28 @@ namespace irr
 
 				static float get(const state_type* state, const core::vectorSIMDu32& pixelCoord, const int32_t& channel)
 				{
-					// TODO: to define in future
+					auto getWangHash = [&]()
+					{
+						IImage::SBufferCopy::getLocalByteOffset(pixelCoord, bufferStridesHash);
+						uint32_t seed = IImage::SBufferCopy::getLocalByteOffset(pixelCoord, bufferStridesHash) * channel;
```
devshgraphicsprogramming: dont use byte offsets, just compute a pixel index value `(((channel*extent.depth+z)*extent.height+y)*extent.width+x`

include/irr/asset/filters/CBlitImageFilter.h
devshgraphicsprogramming: > 
> 
> so what kind of classes should I pass to `normalize`? or should it be only a bool in template param?

SAT, ConvertSwizzle and Blit

do we have any more questions after discord chat?

AnastaZIuk: no, there are not

include/irr/asset/filters/dithering/CWhiteNoiseDither.h
AnastaZIuk: Okay, so i will devide by float and stop using those casts

include/irr/asset/filters/dithering/CPrecomputedDither.h
AnastaZIuk: Okay

include/irr/asset/filters/dithering/CPrecomputedDither.h
AnastaZIuk: Oh okay

include/irr/asset/filters/dithering/CWhiteNoiseDither.h
AnastaZIuk: what about it https://github.com/buildaworldnet/IrrlichtBAW/pull/521#issuecomment-683958930?

include/irr/asset/filters/dithering/CWhiteNoiseDither.h
AnastaZIuk: corrected in following commit

include/irr/asset/filters/dithering/CWhiteNoiseDither.h
AnastaZIuk: corrected in following commit

include/irr/asset/filters/CBlitImageFilter.h
AnastaZIuk: now we've got an interface for that

AnastaZIuk: corrected

include/irr/asset/filters/CSwizzleableAndDitherableFilterBase.h
```diff
@@ -17,20 +17,124 @@ namespace irr
 	{
 		namespace impl
 		{
+			namespace detail
+			{
+				union NormalizeValues
```
devshgraphicsprogramming, line 22: you want a `struct` and then an "anonymous union" of the member variables

include/irr/asset/filters/CSwizzleableAndDitherableFilterBase.h
```diff
@@ -17,20 +17,124 @@ namespace irr
...
+
+						/*
+							Runtime version of normalize
+
+							@see normalize
+						*/
+
+						template<typename Tenc>
+						void normalize(const E_FORMAT& format, Tenc* encodeBuffer, const uint8_t& channels)
+						{
```
devshgraphicsprogramming, line 100: `static_assert` and `assert` that the format is not a pure-integer format

include/irr/asset/filters/CSwizzleableAndDitherableFilterBase.h
```diff
@@ -17,20 +17,124 @@ namespace irr
...
+				};
+
+				template<>
+				class CNormalizeState<false>
+				{
+					public:
+						CNormalizeState() {}
+						virtual ~CNormalizeState() {}
+
+						NormalizeValues oldMaxValue, oldMinValue;
```
devshgraphicsprogramming, line 117: dont declare the members in this specialization

include/irr/asset/filters/dithering/CDither.h
```diff
@@ -47,7 +47,9 @@ namespace irr
 				
 				float get(const IDither::IState* state, const core::vectorSIMDu32& pixelCoord, const int32_t& channel)
 				{
-					return static_cast<CRTP*>(this)->get(static_cast<const typename CRTP::CState*>(state), pixelCoord, channel);
+					const auto& return_value = static_cast<CRTP*>(this)->get(static_cast<const typename CRTP::CState*>(state), pixelCoord, channel);
+					assert(return_value >= 0 && return_value <= 1, "The dither value must be limited to the range!");
```
devshgraphicsprogramming, line 51: only assert `#ifdef _IRR_DEBUG`

devshgraphicsprogramming: also a non-static assert does not take a message argument AFAIK

include/irr/asset/filters/dithering/CPrecomputedDither.h
devshgraphicsprogramming: you want to use `floats` instead of `double`s basically.

devshgraphicsprogramming:
This should be merged into @Crisspl 's branch when its done.

include/irr/asset/filters/CSwizzleableAndDitherableFilterBase.h
AnastaZIuk: not an issue, discussed on the dc

include/irr/asset/filters/CSwizzleableAndDitherableFilterBase.h
AnastaZIuk: not an issue, discussed on the dc

include/irr/asset/filters/CSwizzleableAndDitherableFilterBase.h
AnastaZIuk: corrected in following commit

include/irr/asset/filters/dithering/CDither.h
AnastaZIuk: corrected in following commit

include/irr/asset/filters/dithering/CPrecomputedDither.h
AnastaZIuk: corrected in following commit

PR #522, opened 2020-07-27
GLSL Goodies

devshgraphicsprogramming:


include/irr/builtin/material_compiler/glsl/common.glsl
```diff
@@ -0,0 +1,528 @@
+#version 430 core
```
devshgraphicsprogramming, line 1: rename this header, its not common to all material compilers, only one particular backend (or split out the common and non common part).

For example virtual texturing is only really a part of a backend that does not use the descriptor indexing Vulkan capability.

also path is wrong, should be `irr/builtin/glsl/material_compiler` I think

devshgraphicsprogramming:
If I merge this, will the denoiser break? (it needs a functional mitsuba loader to load camera XMLs)

include/irr/builtin/material_compiler/glsl/common.glsl
Crisspl: > its not common to all material compilers, only one particular backend

thats why its in subdir for glsl backend

> also path is wrong, should be irr/builtin/glsl/material_compiler I think

that would imply scattering material_compiler sources over language/backend-based directories like glsl/material_compiler, backendX/material_compiler, backendY/material_compiler etc

devshgraphicsprogramming: hmm ok makes sense, buut there are many possible backends for GLSL:
- GLSL separate pipeline/shader per BSDF (no multi draw indirect)
- GLSL with virtual geometry and virtual textures (no descriptor indexing)
- GLSL with bindless (descriptor indexing)

plus there is the question if the rasterizer backend will want/need the sampling functons (for stochastic shadows)

Crisspl:
> If I merge this, will the denoiser break? (it needs a functional mitsuba loader to load camera XMLs)

no, it loads metadata just fine, provided shaders wont work obviously but thats it

devshgraphicsprogramming:
ok fix beckmann and ggx, then I'll merge ;)

include/irr/builtin/material_compiler/glsl/common.glsl
devshgraphicsprogramming: @Crisspl take not of this comment, I'm closing the PR

PR #530, opened 2020-08-19
No pipeline fix

Hazardu:


Hazardu:
Not ready to merge
```
irr/builtin/shaders/loaders/mtl/fragment_impl.glsl:212: error: 'UV' : undeclared identifier
irr/builtin/shaders/loaders/mtl/fragment_impl.glsl:212: error: 'constructor' : not enough data provided for constructionirr/builtin/shaders/loaders/mtl/fragment_impl.glsl:212: error: '=' :  cannot convert from ' const float' to ' temp highp 2X2 matrix of float'
```

this is most likely because the hash for the metadata is _wrong_
The engine treats a pipeline without uv like its a pipeline with uv, or vice versa

src/irr/asset/COBJMeshFileLoader.cpp
```diff
@@ -63,6 +63,16 @@ constexpr uint32_t BND_NUM = 0u;
...
 	SVertexInputParams vtxParams;
+	SBlendParams blendParams;
+	SPrimitiveAssemblyParams primParams;
+	SRasterizationParams rasterParams;
+
+	blendParams.blendParams[0].blendEnable = true;
+	blendParams.blendParams[0].srcColorFactor = EBF_ONE;
+	blendParams.blendParams[0].srcAlphaFactor = EBF_ONE;
+	blendParams.blendParams[0].dstColorFactor = EBF_ONE_MINUS_SRC_ALPHA;
+	blendParams.blendParams[0].dstAlphaFactor = EBF_ONE_MINUS_SRC_ALPHA;
```
devshgraphicsprogramming: default should not blend!

src/irr/asset/COBJMeshFileLoader.cpp
```diff
@@ -76,32 +86,87 @@ COBJMeshFileLoader::COBJMeshFileLoader(IAssetManager* _manager) : AssetManager(_
...
+			auto& semantic = (shaderInputsMetadata->end() - i - 1u)[0];
+			semantic.type = types[i];
+			semantic.descriptorSection.type = IPipelineMetadata::ShaderInput::ET_UNIFORM_BUFFER;
+			semantic.descriptorSection.uniformBufferObject.binding = ds1layout->getBindings().begin()[0].binding;
+			semantic.descriptorSection.uniformBufferObject.set = 1u;
+			semantic.descriptorSection.uniformBufferObject.relByteoffset = relOffsets[i];
+			semantic.descriptorSection.uniformBufferObject.bytesize = sizes[i];
+			semantic.descriptorSection.shaderAccessFlags = ICPUSpecializedShader::ESS_VERTEX;
+		}
+	}
```
devshgraphicsprogramming: layouts, shaders and metadata are common for both actual MTL pipelines and a dummy default one.

Also why are you creating this default pipeline in OBJ loader, and not MTL loader?

src/irr/asset/COBJMeshFileLoader.cpp
```diff
@@ -458,10 +523,7 @@ asset::SAssetBundle COBJMeshFileLoader::loadAsset(io::IReadFile* _file, const as
 			//if there's no pipeline for this meshbuffer, set dummy one
 			if (!submeshes[i]->getPipeline())
 			{
-				auto pipeline = getDefaultAsset<ICPURenderpassIndependentPipeline, IAsset::ET_RENDERPASS_INDEPENDENT_PIPELINE>(
-					hasUV ? DUMMY_PIPELINE_UV_CACHE_KEY : DUMMY_PIPELINE_NO_UV_CACHE_KEY,
-					AssetManager
-				);
+				auto pipeline = getDefaultAsset<ICPURenderpassIndependentPipeline, IAsset::ET_RENDERPASS_INDEPENDENT_PIPELINE>(hasUV? DEFAULT_MTL_PIPELINE_UV_CACHE_KEY : DEFAULT_MTL_PIPELINE_NO_UV_CACHE_KEY, AssetManager);
```
devshgraphicsprogramming: what's this new function?

src/irr/asset/COBJMeshFileLoader.cpp
Hazardu: because if there is no mtl file, the pipeline for a mesh buffer is assigned in obj loader

devshgraphicsprogramming: the default result of a loader should be handled by the loader.

src/irr/asset/CGraphicsPipelineLoaderMTL.cpp
```diff
@@ -148,6 +169,55 @@ CGraphicsPipelineLoaderMTL::CGraphicsPipelineLoaderMTL(IAssetManager* _am) : m_a
...
+   
+
+   
+    
+    */
+    auto texture_file = m_assetMgr->getFileSystem()->createMemoryReadFile(MISSING_CHECKERBOARD_TEXTURE_CONTENT, sizeof(MISSING_CHECKERBOARD_TEXTURE_CONTENT), "checkerboard.png");
+
+    SAssetLoadParams params;
+    auto imageBundle = m_assetMgr->getAsset(texture_file, std::string("checkerboard.png"), params);
+    auto image = core::smart_refctd_ptr_dynamic_cast<ICPUImage>(imageBundle.getContents().begin()[0]);
```
Hazardu: I'm getting access violations on this line. I have no idea why

src/irr/asset/CGraphicsPipelineLoaderMTL.cpp
```diff
@@ -324,19 +379,30 @@ SAssetBundle CGraphicsPipelineLoaderMTL::loadAsset(io::IReadFile* _file, const I
...
+                        _override->insertAssetIntoCache(bundle, dsCacheKey, ctx.inner, _hierarchyLevel + ICPURenderpassIndependentPipeline::DESC_SET_HIERARCHYLEVELS_BELOW);
+                    }
+                }
+            }
+            else {
+                const asset::IAsset::E_TYPE types[]{ asset::IAsset::ET_DESCRIPTOR_SET, (asset::IAsset::E_TYPE)0u };
+                auto b = m_assetMgr->findAssets(dsCacheKey, types);
+                if (b)
+                {
+                    ds3 = core::smart_refctd_ptr_dynamic_cast<ICPUDescriptorSet>(b->begin()->getContents().begin()[0]);
```
Hazardu: when using loadAsset() from the constructor, _override is equal to null ptr, and only then. The plan is to load the cached descriptor set with the magenta checkerboard and use it.

src/irr/asset/CGraphicsPipelineLoaderMTL.cpp
devshgraphicsprogramming: bundle is empty?

Hazardu: i found the problem, CImageWriterPNG was initialized after CGraphicsPipelineLoaderMTL

devshgraphicsprogramming: erm, maybe
just maybe

we should have adding of built-in assets be separate from the constructors?

i.e. first all loaders are constructed and added, then we call `init` ?

devshgraphicsprogramming:
@Crisspl can you do the final reviews on this and merge this into your mitsuba megabranch?

Hazardu:
Could you delay merging untill later today?
There are some changes to level curve extraction that I made for the purpose of sending a zip file to a client. 
I'll be home past 17:30, so the last commit should be pushed shortly after

PR #532, opened 2020-08-22
Blit Kernels Test

AnastaZIuk:
status:

todo:

- check all the available kernels and save their outputs


devshgraphicsprogramming:
should I merge anyway @AnastaZIuk ?

AnastaZIuk:
if everything works after your cleanups, there is no need for that

PR #541, opened 2020-10-07
Added a unit test for LRU

Hazardu:
Removed Przemog1's example, and replaced it with a unit test for LRU cache.
At the moment, there is a line that behaves not how its supposed to do. Looking at line 66 of main.cpp, its supposed to insert a new value into a cache. Cache's capacity is 5, and it already has one element.
Expected behavior is adding an element with address of 1 in the array in FixedCapacityDoublyLinkedList.h 
Current behavior is as follows: alloc_address returns 472, triggers assert `cap > address`.


include/irr/core/containers/LRUCache.h
```diff
@@ -167,14 +167,6 @@ class LRUCache : private impl::LRUCacheBase<Key,Value,MapHash,MapEquals>
...
 
 		//get the value from cache at an associated Key, or nullptr if Key is not contained within cache. Does not alter the value use order
-		inline Value* peek(const Key& key)
-		{
-			uint32_t i = common_peek(key);
-			if (i!=invalid_iterator)
-				return &(m_list.get(i)->data.second);
-			else
-				return nullptr;
-		}
```
devshgraphicsprogramming: dont remove the non-const `peek`, because then we'll be never able to peek a `Value*` and access it to modify the value.

Issue #451, opened 2020-04-02
Asset Pipeline V3: Immutable (Const) cache entries and separation of `IAssetMetadata` from `IAsset`


devshgraphicsprogramming:
We have ran into problems with some of the "built-in"s turning out to be mutable (don't want to mutate a built-in), a clone function has been added but it seems not enough.

Mentioned here:
https://github.com/buildaworldnet/IrrlichtBAW/issues/384#issuecomment-544130414

Furthermore making the `IAsset` hold a pointer to `IAssetMetadata` leads to unnecessary asset duplication and problems when trying to optimize the usage of GPU resources (pipelines and memory, in most cases).

To be achieved:
```
A) Immutable cached objects (think const Asset Cache)
B) Metadata not built-into the IAsset (asset cache holding bundles of pair<smart_refctd_ptr<IAsset>,smart_refctd_ptr<IAssetMetadata>>
```

Issue #23, opened 2018-01-31
Make getMesh() thread-safe


devshgraphicsprogramming:
Issue #22 is a pre-requisite

Also requires:
1) making MeshLoaders thread safe.
2) #24 Rewriting the VAO (IMeshDataFormatDesc) system for multiple contexts

Issue #22, opened 2018-01-31
Make getTexture() thread-safe


devshgraphicsprogramming:
Involves:
1) Rework the Atomic Lock interface to be more irrlicht-ty
2) Rewriting IReferenceCounted to include an Atomic Lock
3) making the texture cache thread-safe

PR #549, opened 2020-10-14
Immutable assets

Crisspl:


include/irr/asset/IAsset.h
```diff
@@ -167,25 +173,40 @@ class IAsset : virtual public core::IReferenceCounted
 
         virtual core::smart_refctd_ptr<IAsset> clone(uint32_t _depth = ~0u) const = 0;
 
+		inline E_MUTABILITY getMutability() const { return m_mutability; }
+		inline bool isMutable() const { return getMutability() == EM_MUTABLE; }
+		inline bool canBeConvertedToDummy() const { return (getMutability()&EM_CPU_PERSISTENT) != EM_CPU_PERSISTENT; }
```
devshgraphicsprogramming: not sure I want these as public.

We should have a `setMutability`

include/irr/asset/IAsset.h
```diff
@@ -167,25 +173,40 @@ class IAsset : virtual public core::IReferenceCounted
...
 	private:
 		friend IAssetManager;
 		core::smart_refctd_ptr<IAssetMetadata> m_metadata;
 
-		void setMetadata(core::smart_refctd_ptr<IAssetMetadata>&& _metadata) { m_metadata = std::move(_metadata); }
+		void setMetadata(core::smart_refctd_ptr<IAssetMetadata>&& _metadata) 
+		{
+			//we have to talk about it (TODO)
+			//if (isImmutable_debug())
+			//	return;
```
devshgraphicsprogramming: doesn't matter what you do here because of #451 which means that it should be the `SAssetBundle` that holds the metadata, not the `IAsset` (a future TODO along with #22 #23 which are still open because of a rare scheduling loophole)

include/irr/asset/ICPUBufferView.h
```diff
@@ -44,11 +44,24 @@ class ICPUBufferView : public IBufferView<ICPUBuffer>, public IAsset
 		_IRR_STATIC_INLINE_CONSTEXPR auto AssetType = ET_BUFFER_VIEW;
 		inline E_TYPE getAssetType() const override { return AssetType; }
 
-		ICPUBuffer* getUnderlyingBuffer() { return m_buffer.get(); }
+		ICPUBuffer* getUnderlyingBuffer() 
+		{
+			return m_buffer.get();
```
devshgraphicsprogramming, line 48: forgot something?

include/irr/asset/ICPUDescriptorSet.h
```diff
@@ -148,6 +151,10 @@ class ICPUDescriptorSet final : public IDescriptorSet<ICPUDescriptorSetLayout>,
 		//! Can modify the array of descriptors bound to a particular bindings
 		inline core::SRange<SDescriptorInfo> getDescriptors(uint32_t index) 
 		{ 
+			//this must be commented-out, otherwise cpu2gpu doesnt have access to descriptors
+			//if (isImmutable_debug())
+			//	return core::SRange<SDescriptorInfo>{nullptr, nullptr};
```
devshgraphicsprogramming: does the cpu2gpu need non-`const` `SDescriptorInfo` ?

include/irr/asset/ICPUDescriptorSet.h
```diff
@@ -128,7 +128,10 @@ class ICPUDescriptorSet final : public IDescriptorSet<ICPUDescriptorSetLayout>,
 		_IRR_STATIC_INLINE_CONSTEXPR auto AssetType = ET_DESCRIPTOR_SET;
```
devshgraphicsprogramming, line 135: convertToDummyObject above this line does not check for mutability

include/irr/asset/ICPUDescriptorSet.h
Crisspl: need non-const descriptors

include/irr/asset/ICPUDescriptorSet.h
Crisspl: because it doesnt have to, it would change nothing

include/irr/asset/ICPUImage.h
```diff
@@ -65,10 +65,14 @@ class ICPUImage final : public IImage, public IAsset
 
 		virtual bool validateCopies(const SImageCopy* pRegionsBegin, const SImageCopy* pRegionsEnd, const ICPUImage* src)
 		{
+			//TODO why is it not const method?
```
devshgraphicsprogramming: if it can be const, make it const

include/irr/asset/ICPUDescriptorSet.h
devshgraphicsprogramming: but why?

src/irr/asset/IAssetManager.cpp
```diff
@@ -90,7 +90,7 @@ std::function<void(SAssetBundle&)> irr::asset::makeAssetGreetFunc(const IAssetMa
 		auto rng = _asset.getContents();
 		//assets being in the cache must be immutable
         for (auto ass : rng)
-			_mgr->setAssetMutable(ass.get(), false);
+			_mgr->setAssetMutability(ass.get(), IAsset::EM_IMMUTABLE);
```
devshgraphicsprogramming: why are you changing our default caching to "always immutable" ?

devshgraphicsprogramming:
I see a problem that
```
            if (isDummyObjectForCacheAliasing)
                return;
```
is used an an "early out" on all `convertToDummyObject`

Now because the call is recursive (goes through children) and whether a child is converted depends on its mutability state... this means that just because it was called before on an asset, won't guarantee that a second call is not going to produce side-effects.

This logic was broken even before mutability concerns, when we first introduced the recursion depth parameter.

src/irr/asset/IAssetManager.cpp
devshgraphicsprogramming: default behaviour should be no change (assets go in as MUTABLE)

Crisspl:
> I see a problem that
> 
> ```
>             if (isDummyObjectForCacheAliasing)
>                 return;
> ```
> 
> is used an an "early out" on all `convertToDummyObject`
> 
> Now because the call is recursive (goes through children) and whether a child is converted depends on its mutability state... this means that just because it was called before on an asset, won't guarantee that a second call is not going to produce side-effects.
> 
> This logic was broken even before mutability concerns, when we first introduced the recursion depth parameter.

this `if` doesn't concern mutability at all. However I see your point.. for example at first I call converttodummy(0) and after that converttodummy (2). And levels 1 and 2 won't be converted then

include/irr/asset/IAsset.h
Crisspl: setting asset's mutability is exclusive for asset manager and it's already a friend of IAsset

devshgraphicsprogramming: ah ok, but still the public functions here are a bit ugly, they should be protected.

include/irr/asset/IAsset.h
```diff
@@ -173,6 +173,17 @@ class IAsset : virtual public core::IReferenceCounted
 
         virtual core::smart_refctd_ptr<IAsset> clone(uint32_t _depth = ~0u) const = 0;
 
+		//! restore() is not recursive!
+		virtual bool restore(IAsset* _other)
+		{
+			if (!this->isADummyObjectForCache())
+				return false;
+			assert(!_other->isADummyObjectForCache());
```
devshgraphicsprogramming: would be useful if we called it `swapContents` and had the two assets swap stuff around

include/irr/asset/IAssetManager.h
```diff
@@ -422,8 +422,7 @@ class IAssetManager : public core::IReferenceCounted, public core::QuitSignallin
         //TODO change name
         bool insertAssetIntoCache(SAssetBundle& _asset)
         {
-            const uint32_t ix = IAsset::typeFlagToIndex(_asset.getAssetType());
-            return m_assetCache[ix]->insert(_asset.getCacheKey(), _asset);
+            return insertAssetIntoCache(_asset, IAsset::EM_CPU_PERSISTENT);
```
devshgraphicsprogramming: this is now mutable by default, yes?

include/irr/asset/ICPUPipelineCache.h
```diff
@@ -110,10 +110,21 @@ class ICPUPipelineCache final : public IAsset
...
 			m_cache.clear();
 	}
 
+	bool restore(IAsset* _other) override
+	{
+		if (!IAsset::restore(_other))
+			return false;
+
+		auto* other = static_cast<ICPUPipelineCache*>(_other);
+		std::swap(m_cache, other->m_cache);
```
devshgraphicsprogramming: can we even restore this?

include/irr/asset/ICPUBuffer.h
```diff
@@ -60,6 +60,18 @@ class ICPUBuffer : public asset::IBuffer, public asset::IAsset
...
+            auto* other = static_cast<ICPUBuffer*>(_other);
+            std::swap(data, other->data);
+            assert(size == other->size);
+
+            return true;
+        }
+
         virtual void convertToDummyObject(uint32_t referenceLevelsBelowToConvert=0u) override
         {
             if (isDummyObjectForCacheAliasing)
```
devshgraphicsprogramming: maybe fold this check into `canBeConvertedToDummy`

devshgraphicsprogramming: btw, `convertToDummyObject_common` will set `isDummyObjectForCacheAliasing = true;`

include/irr/asset/ICPUPipelineCache.h
```diff
@@ -110,10 +110,21 @@ class ICPUPipelineCache final : public IAsset
 	size_t conservativeSizeEstimate() const override { return 0ull; /*TODO*/ }
 	void convertToDummyObject(uint32_t referenceLevelsBelowToConvert = 0u) override
 	{
-		if (canBeConvertedToDummy())
+		if (!isADummyObjectForCache() && canBeConvertedToDummy())
```
devshgraphicsprogramming: as I wrote, you could fold `usADummyObjectForCache` into `canBeConvertedToDummy`

include/irr/asset/ICPUPipelineCache.h
Crisspl: hm.. good point. This asset is kinda special because the idea was GPU object->asset->potentially save on disk

include/irr/asset/IAssetManager.h
Crisspl: public `IAssetManager::insertIntoCache` sets asset's mutability as cpu persistent (so that everything that is added to cache explicitely by user is cpu persistent), but `IAssetLoaderOverride` sets mutability as mutable (everything loaded by `getAsset` and subassets that are added to cache during the loading process are mutable)
All the builtins are added to cache as immutable

include/irr/asset/IAsset.h
Crisspl: shouldnt user be able to query asset's mutability level?

include/irr/asset/IAsset.h
```diff
@@ -196,7 +196,7 @@ class IAsset : virtual public core::IReferenceCounted
 
 		inline E_MUTABILITY getMutability() const { return m_mutability; }
 		inline bool isMutable() const { return getMutability() == EM_MUTABLE; }
-		inline bool canBeConvertedToDummy() const { return (getMutability()&EM_CPU_PERSISTENT) != EM_CPU_PERSISTENT; }
+		inline bool canBeConvertedToDummy() const { return !isADummyObjectForCache() && (getMutability()&EM_CPU_PERSISTENT) != EM_CPU_PERSISTENT; }
```
devshgraphicsprogramming: maybe `getMutability()<EM_CPU_PERSISTENT`

devshgraphicsprogramming: instead of bitfield

include/irr/asset/IAsset.h
devshgraphicsprogramming: yes, but we shouldn't have multiple functions to query, just `getMutability` and the enum

obvs `canBeConvertedToDummy` can stay, cause it does more than checking an enum

include/irr/asset/ICPUPipelineCache.h
devshgraphicsprogramming: we'd have to check a few things before restoring, then I guess?

include/irr/asset/IAssetManager.h
devshgraphicsprogramming: ok I guess it kind-of makes sense.

devshgraphicsprogramming: not entirely convinced about the asset manager default behaviour.

include/irr/asset/IAsset.h
```diff
@@ -173,14 +173,24 @@ class IAsset : virtual public core::IReferenceCounted
 
         virtual core::smart_refctd_ptr<IAsset> clone(uint32_t _depth = ~0u) const = 0;
 
-		//! restore() is not recursive!
-		virtual bool restore(IAsset* _other)
+		bool restoreFromDummy(IAsset* _other, uint32_t _levelsBelow)
```
devshgraphicsprogramming: needs default argument for `_levelsBelow` should `~0u`

include/irr/asset/IAsset.h
```diff
@@ -173,14 +173,24 @@ class IAsset : virtual public core::IReferenceCounted
...
+			restoreFromDummy_impl(_other, _levelsBelow);
+			return true;
+		}
+
+		virtual bool canBeRestoredFrom(const IAsset* _other) const
+		{
+			if (getMutability() != EM_MUTABLE)
+				return false;
+			if (!isADummyObjectForCache())
+				return false;
```
devshgraphicsprogramming: this should not be here, just because asset is not mutable or is not a dummy object (cannot be restored, doesn't matter what `_other` is) does not meen the dependencies couldn't be restored.

Same if `_other` is a dummy object, the dependants of `this` which are mutable and dummy could be restored from dependants of `_other` which are mutable and non-dummy.

**However if `_other` does not "match"/"is not compatible" with `this` (i.e. different layout contents) then obviously there's no point restoring the dependants.** 

P.S. when restoring you should check that `_other` is mutable.

include/irr/asset/IAsset.h
```diff
@@ -173,14 +173,24 @@ class IAsset : virtual public core::IReferenceCounted
...
 
-		//! restore() is not recursive!
-		virtual bool restore(IAsset* _other)
+		bool restoreFromDummy(IAsset* _other, uint32_t _levelsBelow)
 		{
-			if (!this->isADummyObjectForCache())
+			if (!canBeRestoredFrom(_other))
+				return false;
+
+			restoreFromDummy_impl(_other, _levelsBelow);
```
devshgraphicsprogramming, line 183: you should probably split this into `restoreFromDummy_this` and `restoreFromDummy_recurse`, to solve comment
https://github.com/buildaworldnet/IrrlichtBAW/pull/549/commits/750f5f9dfa92c3d1e05482fca38a5723768a4682#r507655849

you can only restore `this` from `_other` if:
1. `_other` matches `this`
2. `this` is dummy
3. `this` is mutable
4. `_other` is not a dummy
5. `_other` is mutable

You can only recurse if **`_other` "matches" `_this`, no other conditions necessary**

include/irr/asset/ICPUBuffer.h
```diff
@@ -112,6 +112,13 @@ class ICPUBuffer : public asset::IBuffer, public asset::IAsset
     protected:
         uint64_t size;
         void* data;
+
+        void restoreFromDummy_impl(IAsset* _other, uint32_t _levelsBelow) override
+        {
+            auto* other = static_cast<ICPUBuffer*>(_other);
+
+            std::swap(data, other->data);
```
devshgraphicsprogramming: you're forgetting the size.

devshgraphicsprogramming: ah ok, size is supposed to be identical

include/irr/asset/ICPUComputePipeline.h
```diff
@@ -32,9 +32,9 @@ class ICPUComputePipeline : public IComputePipeline<ICPUSpecializedShader, ICPUP
 		}
 	}
 
-    bool restore(IAsset* _other) override
+    bool canBeRestoredFrom(const IAsset* _other) const override
     {
-        if (!IAsset::restore(_other))
+        if (!IAsset::canBeRestoredFrom(_other))
             return false;
```
devshgraphicsprogramming: the "does `this` match `_other`"restoration check needs to be fully recursive, so need to check whether layouts match (etc.) before trying to restore top level objects

devshgraphicsprogramming: kinda paired up with https://github.com/buildaworldnet/IrrlichtBAW/pull/549#discussion_r507659230

include/irr/asset/ICPUDescriptorSet.h
```diff
@@ -193,6 +193,41 @@ class ICPUDescriptorSet final : public IDescriptorSet<ICPUDescriptorSetLayout>,
...
+					case IDescriptor::EC_BUFFER:
+						static_cast<asset::ICPUBuffer*>(descriptor)->restoreFromDummy(static_cast<ICPUBuffer*>(d_other), _levelsBelow);
+						break;
+					case IDescriptor::EC_IMAGE:
+						static_cast<asset::ICPUImageView*>(descriptor)->restoreFromDummy(static_cast<ICPUImageView*>(d_other), _levelsBelow);
+						if (descriptor->getTypeCategory() == IDescriptor::EC_IMAGE && it->image.sampler)
+							it->image.sampler->restoreFromDummy(other->m_descriptors->begin()[i].image.sampler.get(), _levelsBelow);
+						break;
+					case IDescriptor::EC_BUFFER_VIEW:
+						static_cast<asset::ICPUBufferView*>(descriptor)->restoreFromDummy(static_cast<ICPUBufferView*>(d_other), _levelsBelow);
```
devshgraphicsprogramming: if we'll already check if we can restore entire dependency tree, then this should call `restoreFromDummy_impl` directly instead of `restoreFromDummy` for dependencies

Tied up with the same thread https://github.com/buildaworldnet/IrrlichtBAW/pull/549#discussion_r507664092

include/irr/asset/ICPUDescriptorSetLayout.h
```diff
@@ -66,21 +66,29 @@ class ICPUDescriptorSetLayout : public IDescriptorSetLayout<ICPUSampler>, public
...
 		virtual ~ICPUDescriptorSetLayout() = default;
+
+private:
+    void restoreFromDummy_impl(IAsset* _other, uint32_t _levelsBelow) override
+    {
+        auto* other = static_cast<ICPUDescriptorSetLayout*>(_other);
+
+        //TODO hmm should samplers be swapped (and dropped in convertToDummy()) or recursively restored??
+        std::swap(m_bindings, other->m_bindings);
+        std::swap(m_samplers, other->m_samplers);
```
devshgraphicsprogramming: pointer values of `m_samplers` should not be swapped, only contents (because it would mess up CPU 2 GPU caches)

so recursively restore.

include/irr/asset/ICPUImage.h
```diff
@@ -206,6 +206,14 @@ class ICPUImage final : public IImage, public IAsset
 				return _a.imageSubresource.mipLevel < _b.imageSubresource.mipLevel;
 			}
 		};
+private:
+		void restoreFromDummy_impl(IAsset* _other, uint32_t _levelsBelow) override
+		{
+			auto* other = static_cast<ICPUImage*>(_other);
+
+			std::swap(regions, other->regions);
```
devshgraphicsprogramming: there's also a buffer to consider, not only regions

include/irr/asset/ICPUShader.h
```diff
@@ -55,9 +55,13 @@ class ICPUShader : public IAsset, public IShader
...
 
-		bool restore(IAsset* _other) override
+		bool canBeRestoredFrom(const IAsset* _other) const override
 		{
-			if (!IAsset::restore(_other))
+			if (!IAsset::canBeRestoredFrom(_other))
+				return false;
+
+			auto* other = static_cast<const ICPUShader*>(_other);
+			if (m_containsGLSL != other->m_containsGLSL)
```
devshgraphicsprogramming, line 70: need to think about whether to require this.

include/irr/asset/ICPUSpecializedShader.h
```diff
@@ -65,17 +65,18 @@ class ICPUSpecializedShader : public IAsset, public ISpecializedShader
...
-			auto* other = static_cast<ICPUSpecializedShader*>(_other);
-			assert(m_specInfo.entryPoint == other->m_specInfo.entryPoint);
-			assert(m_specInfo.m_filePathHint == other->m_specInfo.m_filePathHint);
-			assert(m_specInfo.shaderStage == other->m_specInfo.shaderStage);
-			
-			std::swap(m_specInfo.m_entries, other->m_specInfo.m_entries);
+			auto* other = static_cast<const ICPUSpecializedShader*>(_other);
+			if (m_specInfo.entryPoint != other->m_specInfo.entryPoint)
+				return false;
+			if (m_specInfo.m_filePathHint != other->m_specInfo.m_filePathHint)
```
devshgraphicsprogramming: should we require this?

include/irr/asset/IAssetManager.h
```diff
@@ -666,7 +666,7 @@ class IAssetManager : public core::IReferenceCounted, public core::QuitSignallin
                 _outs << "\tKey: " << wtr.first << ", Value: " << static_cast<void*>(wtr.second) << '\n';
         }
 
-        void restoreDummyAsset(SAssetBundle& _bundle)
+        void restoreDummyAsset(SAssetBundle& _bundle, uint32_t _levelsBelow = 0u)
```
devshgraphicsprogramming, line 673: this function is dangerous, it takes no limits/constraints on asset type to be restored (you will usually know the type of asset you're after).

The amount of duplication should be not top level, but top level + `_levelsBelow`

Also handling of mixed dummy and non dummy assets in a bundle is "meh".

Also no check that length of new bundle == old bundle

include/irr/asset/IAsset.h
```diff
@@ -175,17 +175,23 @@ class IAsset : virtual public core::IReferenceCounted
...
 		{
+			assert(getAssetType() == _other->getAssetType());
+
 			if (getMutability() != EM_MUTABLE)
 				return false;
+			if (_other->getMutability() != EM_MUTABLE)
+				return false;
 			if (!isADummyObjectForCache())
 				return false;
 			if (_other->isADummyObjectForCache())
```
devshgraphicsprogramming, line 197: needs a call to `canBeRestoredFrom_recurseDAG` at the end

include/irr/asset/IAsset.h
```diff
@@ -194,6 +200,8 @@ class IAsset : virtual public core::IReferenceCounted
 			return true;
 		}
 
+		virtual bool canBeRestoredFrom_recurseDAG(const IAsset* _other) const = 0;
```
devshgraphicsprogramming: make it protected

include/irr/asset/ICPUSpecializedShader.h
devshgraphicsprogramming: we probably shouldn't require this.

include/irr/asset/ICPUDescriptorSetLayout.h
```diff
@@ -85,9 +82,14 @@ class ICPUDescriptorSetLayout : public IDescriptorSetLayout<ICPUSampler>, public
...
+        const bool restorable = canBeRestoredFrom(_other);
+
         //TODO hmm should samplers be swapped (and dropped in convertToDummy()) or recursively restored??
-        std::swap(m_bindings, other->m_bindings);
-        std::swap(m_samplers, other->m_samplers);
+        if (restorable)
+        {
+            std::swap(m_bindings, other->m_bindings);
+            std::swap(m_samplers, other->m_samplers);
+        }
```
devshgraphicsprogramming: you should recursively restore, dont swap what are essentially pointer lists!

include/irr/asset/ICPUDescriptorSetLayout.h
```diff
@@ -66,11 +66,8 @@ class ICPUDescriptorSetLayout : public IDescriptorSetLayout<ICPUSampler>, public
 			    m_samplers = nullptr;
 		}
 
-        bool canBeRestoredFrom(const IAsset* _other) const override
+        bool canBeRestoredFrom_recurseDAG(const IAsset* _other) const override
         {
-            if (!IAsset::canBeRestoredFrom(_other))
-                return false;
-
```
devshgraphicsprogramming: see https://github.com/buildaworldnet/IrrlichtBAW/pull/549/files/e94f07ab352808077b42c37c490b5c93986df2e5..2ddd5e994ba249c82b43de1b43625f1ff57465a3#r508847179

include/irr/asset/IAsset.h
devshgraphicsprogramming: but the `restoreFromDummy` should not call `canBeRestoredFrom` but `canBeRestoredFrom_this_node` or something like that

include/irr/asset/ICPURenderpassIndependentPipeline.h
```diff
@@ -55,6 +52,16 @@ class ICPURenderpassIndependentPipeline : public IRenderpassIndependentPipeline<
 			if (m_disableOptimizations != other->m_disableOptimizations)
 				return false;
 
+			for (uint32_t i = 0u; i < SHADER_STAGE_COUNT; ++i)
+			{
+				if ((!m_shaders[i] || !other->m_shaders[i]) && m_shaders[i] != other->m_shaders[i])
```
devshgraphicsprogramming: I dont get this if statement

include/irr/asset/ICPUPipelineLayout.h
```diff
@@ -58,11 +58,25 @@ class ICPUPipelineLayout : public IAsset, public IPipelineLayout<ICPUDescriptorS
...
 		}
 
-        bool canBeRestoredFrom(const IAsset* _other) const override
+        bool canBeRestoredFrom_recurseDAG(const IAsset* _other) const override
         {
-            if (!IAsset::canBeRestoredFrom(_other))
-                return false;
+            auto* other = static_cast<const ICPUPipelineLayout*>(_other);
 
+            if ((!m_pushConstantRanges || !other->m_pushConstantRanges) && m_pushConstantRanges != other->m_pushConstantRanges)
```
devshgraphicsprogramming: I dont get this if statement

include/irr/asset/ICPUPipelineLayout.h
```diff
@@ -58,11 +58,25 @@ class ICPUPipelineLayout : public IAsset, public IPipelineLayout<ICPUDescriptorS
...
+                return false;
+            if (m_pushConstantRanges->size() != other->m_pushConstantRanges->size())
+                return false;
```
devshgraphicsprogramming: this is bad, its okay to have no push constant ranges, this will fail if either is 0

include/irr/asset/ICPUPipelineLayout.h
```diff
@@ -58,11 +58,25 @@ class ICPUPipelineLayout : public IAsset, public IPipelineLayout<ICPUDescriptorS
...
+            for (uint32_t i = 0u; i < m_pushConstantRanges->size(); ++i)
+                if ((*m_pushConstantRanges)[i] != (*other->m_pushConstantRanges)[i])
+                    return false;
+
+            for (uint32_t i = 0u; i < DESCRIPTOR_SET_COUNT; ++i)
+            {
+                if ((!m_descSetLayouts[i] || !other->m_descSetLayouts[i]) && m_descSetLayouts[i] != other->m_descSetLayouts[i])
```
devshgraphicsprogramming: I dont get this if statement

include/irr/asset/ICPUMeshBuffer.h
```diff
@@ -149,6 +146,31 @@ class ICPUMeshBuffer : public IMeshBuffer<ICPUBuffer, ICPUDescriptorSet, ICPURen
             return false;
         if (normalAttrId != other->normalAttrId)
             return false;
+        if (m_indexBufferBinding.offset != other->m_indexBufferBinding.offset)
+            return false;
+        if ((!m_indexBufferBinding.buffer || !other->m_indexBufferBinding.buffer) && m_indexBufferBinding.buffer != other->m_indexBufferBinding.buffer)
```
devshgraphicsprogramming: I dont get this if statement

include/irr/asset/ICPUMeshBuffer.h
```diff
@@ -149,6 +146,31 @@ class ICPUMeshBuffer : public IMeshBuffer<ICPUBuffer, ICPUDescriptorSet, ICPURen
...
+            return false;
+        if (!m_indexBufferBinding.buffer->canBeRestoredFrom_recurseDAG(other->m_indexBufferBinding.buffer.get()))
```
devshgraphicsprogramming: will fail with null index buffers

include/irr/asset/ICPUMeshBuffer.h
```diff
@@ -149,6 +146,31 @@ class ICPUMeshBuffer : public IMeshBuffer<ICPUBuffer, ICPUDescriptorSet, ICPURen
...
+            return false;
+        for (uint32_t i = 0u; i < MAX_ATTR_BUF_BINDING_COUNT; ++i)
+        {
+            if (m_vertexBufferBindings[i].offset != other->m_vertexBufferBindings[i].offset)
+                return false;
+            if ((!m_vertexBufferBindings[i].buffer || !other->m_vertexBufferBindings[i].buffer) && m_vertexBufferBindings[i].buffer != other->m_vertexBufferBindings[i].buffer)
+                return false;
+            if (!m_vertexBufferBindings[i].buffer->canBeRestoredFrom_recurseDAG(other->m_vertexBufferBindings[i].buffer.get()))
```
devshgraphicsprogramming: same bug as all your optional pointer comparisons

include/irr/asset/ICPUMeshBuffer.h
```diff
@@ -149,6 +146,31 @@ class ICPUMeshBuffer : public IMeshBuffer<ICPUBuffer, ICPUDescriptorSet, ICPURen
...
+                return false;
+        }
+
+        if ((!m_descriptorSet || !other->m_descriptorSet) && m_descriptorSet != other->m_descriptorSet)
+            return false;
+        if (!m_descriptorSet->canBeRestoredFrom_recurseDAG(other->m_descriptorSet.get()))
```
devshgraphicsprogramming: pointer mess again

include/irr/asset/ICPUMeshBuffer.h
```diff
@@ -149,6 +146,31 @@ class ICPUMeshBuffer : public IMeshBuffer<ICPUBuffer, ICPUDescriptorSet, ICPURen
...
+            return false;
+
+        if ((!m_pipeline || !other->m_pipeline) && m_pipeline != other->m_pipeline)
+            return false;
+        if (!m_pipeline->canBeRestoredFrom_recurseDAG(other->m_pipeline.get()))
```
devshgraphicsprogramming: pipeline is not optional IIRC

include/irr/asset/IAsset.h
devshgraphicsprogramming: leave a comment pointing to the github issue.

devshgraphicsprogramming:
TODO:
- [x] Comments
- [x] ~~non-const getters on non-mutable assets to return `nullptr` etc.~~ make non-const getters on non-mutable assets `assert(false)`
- [x] `const_cast` in CPU2GPU converter (so that const getters can be used)

include/irr/asset/IAsset.h
devshgraphicsprogramming: ok `isMutable` might actually be useful

include/irr/asset/ICPUMeshBuffer.h
Crisspl: it is, it's not used at all right now in a way other than returning pipeline from mesh loader

devshgraphicsprogramming: you cant draw a meshbuffer without a pipeline, you dont even know what primitive type it is

Crisspl: i mean `video` namespace ignores it

devshgraphicsprogramming: !?

AFAIK its an error to have a ICPUMeshBuffer with a nullptr ICPUPipeline

include/irr/asset/IAsset.h
Crisspl: cant because an asset has to be able to call it on its dependencies

include/irr/asset/IAsset.h
Crisspl: > needs a call to canBeRestoredFrom_recurseDAG at the end

what, no, why

Crisspl: what i've noted from our conversation:
* `restoreFromDummy` first calls `canBeRestoredFrom_recurseDAG` and early exits if false, otherwise calls `restoreFromDummy_impl`
* `restoreFromDummy_impl` calls `canBeRestoredFrom` and restores if true, but regardless of result calls `restoreFromDummy` on dependencies

include/irr/asset/IAsset.h
devshgraphicsprogramming: but asset and its dependencies both inherit from `IAsset` ?

include/irr/asset/ICPUPipelineLayout.h
devshgraphicsprogramming: missed this one

include/irr/asset/IAsset.h
devshgraphicsprogramming: What you're missing that is the public `canBeRestoredFrom` checks for whether mutability and dummy state of self and this is ok (what I suggest calling `canBeRestoredFrom_this_node`, which would be equivalent to your current 4 if-statements) AND THEN the "matching" of the DAG

because of this, the `restoreFromDummy_impl` should not call this function to prevent the resotration of current asset in the DAG, but a protected `canBeRestoredFrom_this_node`.

`restoreFromDummy` shouldn't call the public function either but only `canBeRestoredFrom_recurseDAG` to early out.

include/irr/asset/IAsset.h
devshgraphicsprogramming: actually https://github.com/Devsh-Graphics-Programming/Nabla/issues/14

include/irr/asset/IAsset.h
Crisspl: soo
> needs a call to canBeRestoredFrom_recurseDAG at the end

is not the case
Your description above perfectly matches what actually is now

include/irr/asset/IAssetManager.h
Crisspl: > this function is dangerous, it takes no limits/constraints on asset type to be restored

asset bundle can hold only one type of assets if thats what you mean

include/irr/asset/IAsset.h
Crisspl: doesnt matter

devshgraphicsprogramming: ok there's a simple workaround for that, move `canBeRestoredFrom_recurseDAG` to protected and also provide this as a workaround:
```
inline bool canBeRestoredFrom_recurseDAG_call(const IAsset* _other) const
{
    return canBeRestoredFrom_recurseDAG(_other);
}
```
then because `canBeRestoredFrom_recurseDAG_call` is not virtual derived classes should be able to call it, correcT?

include/IDriver.h
```diff
@@ -376,7 +376,7 @@ class IDriver : public virtual core::IReferenceCounted, public IVideoCapabilityR
 		//! Utility function to convert all your CPU asset data into GPU objects ready for use
         // With a custom converter, you can override it to for example; pack all buffers into one, pack all images into one atlas, etc.
 
-        template<typename AssetType>
+        /*template<typename AssetType>
```
devshgraphicsprogramming: what happened here?

include/irr/asset/CCPUMesh.h
```diff
@@ -55,15 +55,23 @@ class CCPUMesh final : public ICPUMesh
...
 				return MeshBuffers[nr].get();
 			else
 				return nullptr;
 		}
 
+		//! returns pointer to a mesh buffer
+		virtual ICPUMeshBuffer* getMeshBuffer(uint32_t nr) override
+		{
+			if (isImmutable_debug())
+				return nullptr;
```
devshgraphicsprogramming: just assert ;)

include/irr/asset/IAsset.h
```diff
@@ -206,8 +206,8 @@ class IAsset : virtual public core::IReferenceCounted
 		inline bool isMutable() const { return getMutability() == EM_MUTABLE; }
 		inline bool canBeConvertedToDummy() const { return !isADummyObjectForCache() && (getMutability()&EM_CPU_PERSISTENT) != EM_CPU_PERSISTENT; }
 
-    protected:
 		virtual void restoreFromDummy_impl(IAsset* _other, uint32_t _levelsBelow) = 0;
```
devshgraphicsprogramming: why public?

include/irr/asset/ICPUBuffer.h
```diff
@@ -99,22 +99,21 @@ class ICPUBuffer : public asset::IBuffer, public asset::IAsset
...
-                return nullptr;
+            assert(!isImmutable_debug());
             return data; 
         }
 
-    protected:
-        uint64_t size;
-        void* data;
-
         void restoreFromDummy_impl(IAsset* _other, uint32_t _levelsBelow) override
```
devshgraphicsprogramming, line 107: why public?

include/irr/asset/ICPUComputePipeline.h
```diff
@@ -58,32 +58,32 @@ class ICPUComputePipeline : public IComputePipeline<ICPUSpecializedShader, ICPUP
...
     const ICPUSpecializedShader* getShader() const { return m_shader.get(); }
     void setShader(ICPUSpecializedShader* _cs) 
     {
-        if (isImmutable_debug())
-            return;
+        assert(!isImmutable_debug());
         m_shader = core::smart_refctd_ptr<ICPUSpecializedShader>(_cs); 
     }
 
-private:
```
devshgraphicsprogramming: why

include/irr/asset/ICPUPipelineCache.h
```diff
@@ -133,7 +133,6 @@ class ICPUPipelineCache final : public IAsset
 		return core::make_smart_refctd_ptr<ICPUPipelineCache>(std::move(cache_cp));
 	}
 
-private:
```
devshgraphicsprogramming, line 127: why

source/Irrlicht/IDriver.cpp
```diff
@@ -57,7 +57,7 @@ struct AssetBundleIterator
 };
 
 //! Maybe we can reduce code duplication here some day
-template<typename AssetType>
+/*template<typename AssetType>
```
devshgraphicsprogramming: I need an explainer about this

src/irr/asset/CPLYMeshWriter.cpp
```diff
@@ -169,8 +169,10 @@ bool CPLYMeshWriter::writeAsset(io::IWriteFile* _file, const SAssetWriteParams&
             memcpy(indices, buf->getPointer(), buf->getSize());
         }
     }
-    else
-        indices = meshBuffer->getIndices();
+    else {
+        indices = _IRR_ALIGNED_MALLOC(idxCnt * (idxtype == asset::EIT_16BIT ? 2u : 4u), _IRR_SIMD_ALIGNMENT);
+        memcpy(indices, meshBuffer->getIndices(), idxCnt * (idxtype == asset::EIT_16BIT ? 2u : 4u));
```
devshgraphicsprogramming, line 174: why?

source/Irrlicht/IDriver.cpp
Crisspl: ah i forgot to re-enable this

include/irr/asset/IAsset.h
Crisspl: > then because canBeRestoredFrom_recurseDAG_call is not virtual derived classes should be able to call it, correcT?

no, why?
hm, i think workaround would be put this into IAsset
```
inline static bool IAsset::canBeRestoredFrom_recurseDAG_call(const IAsset* _this_child, const IAsset* _other_child)
{
    return _this_child->canBeRestoredFrom_recurseDAG(_other_child);
}
```
and then call in derived classes, i.e. in case of DS:
`canBeRestoredFrom_recurseDAG_call(getLayout(), other->getLayout())`

same for `restoreFromDummy_impl`

Crisspl: im doing this

include/irr/asset/IAsset.h
devshgraphicsprogramming: TODO

include/irr/asset/IAsset.h
devshgraphicsprogramming: if only asset manager can call this (just how it is for `convertToDummyObject`) then dont make the function public?

include/irr/asset/IAsset.h
devshgraphicsprogramming: no its not, there's one thing missing

The fucntion here, `canBeRestoredFrom`, should do the 4 if statements plus a call to `canBeRestoredFrom_recurseDAG`

include/irr/asset/IAsset.h
devshgraphicsprogramming: leave a comment

include/irr/asset/IAssetManager.h
devshgraphicsprogramming: hmm ok can keep as is, just document want Criss wrote on the wiki @AnastaZIuk

include/irr/asset/IAsset.h
devshgraphicsprogramming: ugh I said it wrong again.

include/irr/asset/IAsset.h
```diff
@@ -167,25 +173,81 @@ class IAsset : virtual public core::IReferenceCounted
...
+		inline static bool canBeRestoredFrom_recurseDAG_call(const IAsset* _this_child, const IAsset* _other_child)
+		{
+			return _this_child->canBeRestoredFrom_recurseDAG(_other_child);
+		}
+		inline static void restoreFromDummy_impl_call(IAsset* _this_child, IAsset* _other_child, uint32_t _levelsBelow)
+		{
+			_this_child->restoreFromDummy_impl(_other_child, _levelsBelow);
+		}
+
+		virtual bool canBeRestoredFrom_recurseDAG(const IAsset* _other) const = 0;
```
devshgraphicsprogramming: should take a `uint32_t _levelsBelow` parameter

include/irr/asset/ICPUImageView.h
```diff
@@ -57,9 +55,44 @@ class ICPUImageView final : public IImageView<ICPUImage>, public IAsset
...
+			if (params.flags != rhs.flags)
+				return false;
+			if (params.format != rhs.format)
+				return false;
+			if (params.components != rhs.components)
+				return false;
+			if (params.viewType != rhs.viewType)
+				return false;
+			if (memcmp(&params.subresourceRange, &rhs.subresourceRange, sizeof(params.subresourceRange)))
+				return false;
```
devshgraphicsprogramming, line 78: ugh this is so coupled.

include/irr/asset/IAsset.h
```diff
@@ -167,25 +173,81 @@ class IAsset : virtual public core::IReferenceCounted
...
+			assert(getAssetType() == _other->getAssetType());
+
+			if (!canBeRestoredFrom_recurseDAG(_other))
+				return false;
+
+			restoreFromDummy_impl(_other, _levelsBelow);
+			return true;
+		}
+
+		bool canBeRestoredFrom(const IAsset* _other) const
```
devshgraphicsprogramming, line 187: ok lets take it from the top.

`canBeRestoredFrom(const IAsset* _other)` as you have it now, should be called `canBeRestoredFrom(const IAsset* _other, uint32_t levelsBelow=~0u)`

Because one SHOULD call `canBeRestoredFrom` with the exact same parameters before calling `restoreFromDummy` via the asset manager, to make sure we don't fuck up during the restore.

so a call to `canBeRestoredFrom_recurseDAG` is in order at the end

ALSO THE CURRENT CODE OF THIS FUNCTION SHOULD MOVE TO A FUNCTION CALLED `canBeRestoredFrom_self`

include/irr/asset/IAsset.h
devshgraphicsprogramming: oh, we already discussed it

devshgraphicsprogramming: just remind me of the corner case that would break a `_levelsBelow` parameter version?

include/irr/asset/IAsset.h
devshgraphicsprogramming: The general idea

User calls `restoreFromDummy` via Asset Manager, and is supposed to check if calling `restoreFromDummy` would result in catastrophic failure by first making sure calling `canBeRestoredFrom` with identical arguments returns true.

Therefore, `canBeRestoredFrom` needs to only check for "asset matching", NOT mutability and dummy states of particular assets.

I propose a `willSelfBeRestoredFrom` which would do the current checks that your current version of `canBeRestoredFrom` does, and you'd use that for the `const bool restorable` in your implementations of restoration (the non recusive portion if-statement).

include/irr/asset/IAssetManager.h
devshgraphicsprogramming: but I'd instead add a `mutability` argument to the function with ` IAsset::EM_CPU_PERSISTENT` as the default parameter so user doesn't have to do the long form of first inserting an asset and then changing its mutability to what he actually wants

include/irr/asset/IAssetManager.h
devshgraphicsprogramming: no that's not what I meant, I mean that you pass a bundle of some type, then use the bundle's cache key value to try and get a fresh load (badly because only top level will be duplicated and for restore you need the full DAG).

Then you dont limit the asset types you will retrieve/load from `getAssetInHierarchy` so the bundle you might get back could be a different length and even type.

Furthermore you `assert` and not fail gracefully if you dont get same size bundle.

Then you restore without extra checks.

Its basically a surefire way to program termination either via `assert` or trying to restore assets from different types.

Basically the usefulness of this method approaches 0

include/irr/asset/IAsset.h
Crisspl: anything, you want to restore pipeline to the level of pipeline layout so you check only those levels but it might turn out that pipelines are incompatible on some lower level (ds layouts for example)

devshgraphicsprogramming: ok, thx

include/irr/asset/ICPUMeshBuffer.h
Crisspl: never heard of this really

Crisspl:
All done

PR #550, opened 2020-10-16
Spirv opt

Hazardu:
Optimizes fragment shader code in example 42: `Fragment Shader Path Tracer` using SPIRV Tools


devshgraphicsprogramming:
Done on NAbla

PR #553, opened 2020-11-04
GPU Arithmetics test

Hazardu:


examples_tests/48.ArithmeticUnitTest/main.cpp
```diff
@@ -0,0 +1,230 @@
...
+	if (result == video::E_DRIVER_FENCE_RETVAL::EDFR_TIMEOUT_EXPIRED || result == video::E_DRIVER_FENCE_RETVAL::EDFR_FAIL)
+	{
+		os::Printer::log("Could not download the buffer from the GPU, fence not signalled!", ELL_ERROR);
+		downloadStagingArea->multi_free(1u, &address, &BUFFER_SIZE, nullptr);
+		return false;
+	}
+	if (downloadStagingArea->needsManualFlushOrInvalidate())
+		driver->invalidateMappedMemoryRanges({ {downloadStagingArea->getBuffer()->getBoundMemory(),address,BUFFER_SIZE} });
+
+	auto dataFromBuffer = reinterpret_cast<uint32_t*>(reinterpret_cast<uint8_t*>(downloadStagingArea->getBufferPointer()) + address);
```
Hazardu, line 115: I'd to mess up somewhere, as the returned buffer is full of zeroes. I do not know whether it is initially filled with zeroes and never assigned (messed up binding), all the methods in the compute shader return zero or if the buffer gets zeroed out during down streaming.

examples_tests/48.ArithmeticUnitTest/main.cpp
```diff
@@ -0,0 +1,230 @@
...
+	//now check if the data obtained has valid values
+	uint32_t* end = dataFromBuffer + BUFFER_SIZE / sizeof(uint32_t);
+	uint32_t invocationIndex = 0;
+	for (uint32_t* i = dataFromBuffer; i < end; i++)
+	{
+		uint32_t val = *(inputData + invocationIndex);	//TODO instead the template function and compare result
```
Hazardu: Here i need to emulate what is being done on the GPU and then compare it to the downloaded result

examples_tests/48.ArithmeticUnitTest/main.cpp
```diff
@@ -0,0 +1,230 @@
...
+		/*if (val != *i)
+			goto validationEndFalse;*/
```
Hazardu: Don't fire me for using `goto` pls

devshgraphicsprogramming: pack your things

examples_tests/48.ArithmeticUnitTest/main.cpp
```diff
@@ -0,0 +1,230 @@
...
+	TO_MUL,
+	TO_MIN,
+	TO_MAX
+};
+
+template<typename T, typename OP, uint32_t subgroupSize>
+T emulatedSubgroupReduction(const T* data, const uint32_t globalInvocationIndex)
+{
+	auto subgroupID = globalInvocationIndex / subgroupSize;
+	//auto subgroupOffset = data + subgroupID * subgroupSize;
```
devshgraphicsprogramming: this formula is correct, although faster thing would be to `assert` that "subgroupSize is power of two"

then compute `subgroupOffset = data+localInvocationIndex&(subgroupSize-1u);`

examples_tests/48.ArithmeticUnitTest/main.cpp
```diff
@@ -0,0 +1,230 @@
...
+	irr::SIrrlichtCreationParameters params;
+	params.Bits = 24; //may have to set to 32bit for some platforms
+	params.ZBufferBits = 24; //we'd like 32bit here
+	params.DriverType = video::EDT_OPENGL; //! Only Well functioning driver, software renderer left for sake of 2D image drawing
+	params.WindowSize = dimension2d<uint32_t>(1280, 720);
+	params.Fullscreen = false;
+	params.Vsync = true; //! If supported by target platform
+	params.Doublebuffer = true;
+	params.Stencilbuffer = false; //! This will not even be a choice soon
+	params.StreamingDownloadBufferSize = BUFFER_SIZE * 2;
```
devshgraphicsprogramming: why 200% the size?

examples_tests/48.ArithmeticUnitTest/main.cpp
```diff
@@ -0,0 +1,230 @@
...
+
+	uint32_t* inputData = new uint32_t[BUFFER_SIZE / sizeof(uint32_t)];
+
+	auto inputDataBuffer = make_smart_refctd_ptr<ICPUBuffer>(BUFFER_SIZE);
+	{
+		uint32_t* ptr = static_cast<uint32_t*>(inputDataBuffer->getPointer());
+		for (size_t i = 0; i < BUFFER_SIZE / sizeof(uint32_t); i++)
+		{
+			auto memAddr = ptr + i;
+			uint32_t randomValue = std::rand();
```
devshgraphicsprogramming: use C++11 mersenne twister etc.

devshgraphicsprogramming: ran only has a 16bit range or something like that

examples_tests/48.ArithmeticUnitTest/main.cpp
```diff
@@ -0,0 +1,230 @@
...
+			*memAddr = randomValue;
+			inputData[i] = randomValue;
+		}
+	}
+	core::smart_refctd_ptr<IGPUBuffer> gpuinputDataBuffer = core::smart_refctd_ptr<IGPUBuffer>(driver->getGPUObjectsFromAssets(&inputDataBuffer, &inputDataBuffer + 1)->front()->getBuffer());
```
devshgraphicsprogramming: dont do the long way of creating an ICPUBuffer from the array, and then converting it... there's a special function in the driver called "create Filled Device Local On Ded Mem" IIRC which only takes buffer size and contents as parameter

examples_tests/48.ArithmeticUnitTest/main.cpp
```diff
@@ -0,0 +1,230 @@
...
+		{TO_AND,EDT_STORAGE_BUFFER,1u,IGPUSpecializedShader::ESS_COMPUTE,nullptr},
+		{TO_XOR,EDT_STORAGE_BUFFER,1u,IGPUSpecializedShader::ESS_COMPUTE,nullptr},
+		{TO_OR,EDT_STORAGE_BUFFER,1u,IGPUSpecializedShader::ESS_COMPUTE,nullptr},
+		{TO_ADD,EDT_STORAGE_BUFFER,1u,IGPUSpecializedShader::ESS_COMPUTE,nullptr},
+		{TO_MUL,EDT_STORAGE_BUFFER,1u,IGPUSpecializedShader::ESS_COMPUTE,nullptr},
+		{TO_MIN,EDT_STORAGE_BUFFER,1u,IGPUSpecializedShader::ESS_COMPUTE,nullptr},
+		{TO_MAX,EDT_STORAGE_BUFFER,1u,IGPUSpecializedShader::ESS_COMPUTE,nullptr},
+	};
+	auto gpuDSLayout = driver->createGPUDescriptorSetLayout(binding, binding + 8);
+	constexpr uint32_t pushconstantSize = 1;
```
devshgraphicsprogramming: IIRC push constants size are given in bytes and sizes+offsets need to be aligned to 4 bytes

examples_tests/48.ArithmeticUnitTest/main.cpp
```diff
@@ -0,0 +1,230 @@
...
+{
+	auto alteredGLSL = glsl.glsl.replace(glsl.workgroup_definition_position, 4, std::to_string(wg_count));
+	auto shaderUnspecialized = core::make_smart_refctd_ptr<asset::ICPUShader>(alteredGLSL.data());
+	asset::ISpecializedShader::SInfo specinfo(nullptr, nullptr, "main", IGPUSpecializedShader::ESS_COMPUTE,"../file.comp");
+	auto cs = core::make_smart_refctd_ptr<asset::ICPUSpecializedShader>(std::move(shaderUnspecialized), std::move(specinfo));
+	auto cs_rawptr = cs.get();
+	core::smart_refctd_ptr<IGPUSpecializedShader> shader = driver->getGPUObjectsFromAssets(&cs_rawptr, &cs_rawptr + 1)->front();
+	return shader;
+};
+	for (size_t current_Workgroup = 2; current_Workgroup < 1024; current_Workgroup++)
```
devshgraphicsprogramming: leave a comment that subgroup size is hardcoded to 4 (there's a requirement that workgroup be at least half the size) for now and max workgroup size is hardcoded to 256

examples_tests/48.ArithmeticUnitTest/common.glsl
```diff
@@ -0,0 +1,35 @@
+
+#include "irr/builtin/glsl/subgroup/arithmetic_portability.glsl"
+
```
devshgraphicsprogramming: you're missing `layout (workgroup_size_x=_IRR_GLSL_WORKGROUP_SIZE_) in;` to actually declare the size of the workgroup to GLSL

examples_tests/48.ArithmeticUnitTest/main.cpp
```diff
@@ -0,0 +1,230 @@
...
+	auto device = createDeviceEx(params);
+
+	if (!device)
+		return 1; // could not create selected driver.
+
+	video::IVideoDriver* driver = device->getVideoDriver();
+	io::IFileSystem* filesystem = device->getFileSystem();
+	asset::IAssetManager* am = device->getAssetManager();
+
+	uint32_t* inputData = new uint32_t[BUFFER_SIZE / sizeof(uint32_t)];
```
devshgraphicsprogramming, line 156: # @Hazardu **get into the habit of declaring a `delete/free` as soon as you type out a `new/malloc` with your keyboard**

examples_tests/48.ArithmeticUnitTest/testExclusive.comp
```diff
@@ -0,0 +1,15 @@
+#version 450
+#define _IRR_GLSL_WORKGROUP_SIZE_ ???? 
+
+#include "common.glsl"
+
+void main()
+{
+	andOutput [gl_LocalInvocationIndex] = irr_glsl_subgroupExclusiveAnd(inputValue[gl_LocalInvocationIndex]);
```
devshgraphicsprogramming: `gl_LocalInvocationIndex` is the index within a workgroup, not within all dispatches

examples_tests/48.ArithmeticUnitTest/main.cpp
```diff
@@ -0,0 +1,230 @@
...
+
+		if (device->run())
+		{
+			driver->beginScene(true);
+			for (size_t i = 0; i < 3; i++)	//1 because skipping reduce! UNDO THIS
+			{
+				driver->bindComputePipeline(pipelines[i].get());
+				const video::IGPUDescriptorSet* ds = descriptorSet.get();
+				driver->bindDescriptorSets(video::EPBP_COMPUTE, pipelines[i]->getLayout(), 0u, 1u, &ds, nullptr);
+				driver->dispatch(BUFFER_SIZE / (sizeof(uint32_t)), 1, 1);
```
devshgraphicsprogramming: this is wrong, if you have 32 milion invocations to process, and between 2 and 1024 invocations per workgroup.. then you need to divide by the workgroup size as well because dispatch counts in workgroups not invocations

you also need to handle the case when the number of invocations in a workgroup times the size of a uint doesn't divide the buffer size exactly

examples_tests/48.ArithmeticUnitTest/main.cpp
devshgraphicsprogramming: you compute shader never runs, see other comments

examples_tests/48.ArithmeticUnitTest/main.cpp
```diff
@@ -0,0 +1,230 @@
...
+				video::COpenGLExtensionHandler::extGlMemoryBarrier(GL_ALL_BARRIER_BITS);
```
devshgraphicsprogramming: note for the future, because you're only using the buffer for a copy and then reusing it as an ssbo, only the following combination of flags is required
`GL_BUFFER_UPDATE_BARRIER_BIT|GL_SHADER_STORAGE_BARRIER_BIT`

examples_tests/48.ArithmeticUnitTest/main.cpp
Hazardu: i was testing if downloading fails

examples_tests/48.ArithmeticUnitTest/main.cpp
Hazardu: well, that was the code you provided me with, and `subgroupOffset` was never used

devshgraphicsprogramming: it was to emulate a single workgroup, not the whole dispatch

examples_tests/48.ArithmeticUnitTest/common.glsl
Hazardu: `workgroup_size_x` didnt exist, so i used `local_size_x `

examples_tests/48.ArithmeticUnitTest/main.cpp
```diff
@@ -19,25 +20,71 @@ enum TestOperation {
...
+//subgroup method emulations on the CPU, to verify the results of the GPU methods
+template<typename T>
+T emulatedSubgroupReduction(const T* data, const uint32_t globalInvocationIndex, uint32_t subgroupSize, arithmeticFuncPtr pFunc)
 {
 	auto subgroupID = globalInvocationIndex / subgroupSize;
-	//auto subgroupOffset = data + subgroupID * subgroupSize;
 	T retval = data[0];
 	for (auto i = 1u; i < subgroupSize; i++)
-		retval = OP(retval, data[i]);
+		retval = pFunc(retval, data[i]);
```
devshgraphicsprogramming, line 49: its wront you need the subgroup offset to data 🤦

examples_tests/48.ArithmeticUnitTest/main.cpp
```diff
@@ -19,25 +20,71 @@ enum TestOperation {
...
 }
-uint32_t add(uint32_t a, uint32_t b)
+template<typename T>
+T emulatedScanExclusive(const T* data, const uint32_t globalInvocationIndex, uint32_t subgroupSize, arithmeticFuncPtr pFunc)
 {
-	return a + b;
+	auto subgroupID = globalInvocationIndex / subgroupSize;
+	auto subgroupInvocationID = globalInvocationIndex % subgroupSize;
+	T retval = data[0];
+	for (auto i = 1u; i < subgroupSize; i++)
```
devshgraphicsprogramming, line 58: you could have put `<subgroupInvocationID` in the for-loop condition

examples_tests/48.ArithmeticUnitTest/main.cpp
Hazardu: so it actually had a use then...

examples_tests/48.ArithmeticUnitTest/main.cpp
```diff
@@ -19,25 +20,71 @@ enum TestOperation {
...
+	}
+	return retval;
 }
+template<typename T>
+T emulatedScanInclusive(const T* data, const uint32_t globalInvocationIndex, uint32_t subgroupSize, arithmeticFuncPtr pFunc)
+{
+	auto subgroupID = globalInvocationIndex / subgroupSize;
+	auto subgroupInvocationID = globalInvocationIndex % subgroupSize;
+	T retval = data[0];
+	for (auto i = 1u; i < subgroupSize; i++)
```
devshgraphicsprogramming, line 72: you could have put `<=subgroupInvocationID` in the forloop condition instead of the break

examples_tests/48.ArithmeticUnitTest/main.cpp
```diff
@@ -19,25 +20,71 @@ enum TestOperation {
...
 	return retval;
 }
-uint32_t add(uint32_t a, uint32_t b)
+template<typename T>
+T emulatedScanExclusive(const T* data, const uint32_t globalInvocationIndex, uint32_t subgroupSize, arithmeticFuncPtr pFunc)
 {
-	return a + b;
+	auto subgroupID = globalInvocationIndex / subgroupSize;
+	auto subgroupInvocationID = globalInvocationIndex % subgroupSize;
+	T retval = data[0];
```
devshgraphicsprogramming, line 57: DATA offset!!!!!

examples_tests/48.ArithmeticUnitTest/main.cpp
```diff
@@ -19,25 +20,71 @@ enum TestOperation {
...
+		retval = pFunc(retval, data[i]);
+	}
+	return retval;
 }
+template<typename T>
+T emulatedScanInclusive(const T* data, const uint32_t globalInvocationIndex, uint32_t subgroupSize, arithmeticFuncPtr pFunc)
+{
+	auto subgroupID = globalInvocationIndex / subgroupSize;
+	auto subgroupInvocationID = globalInvocationIndex % subgroupSize;
+	T retval = data[0];
```
devshgraphicsprogramming, line 71: DATA offset!!!!!

examples_tests/48.ArithmeticUnitTest/main.cpp
devshgraphicsprogramming: everything is fine, but the variable name is misleading

